PR #174, opened 2020-04-07
Machete file absence handling

micpiotrowski:
- Appropriate message is displayed in the center of machete plugin window in cases of absense of machete file or when it's empty
- `Utils` class was added to `backendRoot` - actually as a proxy object for invoking `GitCoreRepository#getGitFolderPathByRepoRootPath` method from `frontendUI` module


frontendUi/src/main/java/com/virtuslab/gitmachete/frontend/ui/table/GitMacheteGraphTableManager.java
```diff
@@ -48,47 +50,68 @@ public GitMacheteGraphTableManager(Project project, VcsRootDropdown vcsRootDropd
     this.gitMacheteRepositoryBuilderFactory = BackendFactoryModule.getInjector()
         .getInstance(IGitMacheteRepositoryBuilderFactory.class);
     this.repositoryGraphFactory = new RepositoryGraphFactory();
-    this.pathToRepoRoot = Paths.get(vcsRootDropdown.getValue().getRoot().getPath());
+    this.vcsRootDropdown = vcsRootDropdown;
 
     // InitalizationChecker allows us to invoke instance methods below because the class is final
```
mkondratek, line 55: s/Initalization/Initialization

mkondratek:
image

This may not be urgent but I think that text wrapping shall be taken into consideration :)

micpiotrowski:
> image
> 
> This may not be urgent but I think that text wrapping shall be taken into consideration :)

Oh, yea, I forgot about this. Now text is divided into two lines.
In future maybe a good idea will be displaying own component (that could be customized) rather than using `empty text` from `JBTable`

PR #175, opened 2020-04-08
Remove dependency injection completely from the project

PawelLipski:
In hindsight, I've got the feeling that my decision to support DI use for this project was kinda overkill... WDYT guys?


mkondratek:
> In hindsight, I've got the feeling that my decision to support DI use for this project was kinda overkill... WDYT guys?

I fully agree. For now (and it does not seem to gonna change soon) we had no significant gain (probably rather troubles) with DI

PR #181, opened 2020-04-09
[WIP] Add a sample test with ide-probe fat jar

PawelLipski:


PawelLipski:
It seems it's rather hard to open view/tab while IntelliJ is being run in headless mode... let's skip this for now, we have other priorities.

PR #182, opened 2020-04-09
Enforce explicit annotation import in Checker *.astub files

PawelLipski:
Necessary coz Checker doesn't issue any warning when using an annotation that's not imported.


scripts/enforce-explicit-annotation-import-in-astubs
```diff
@@ -0,0 +1,21 @@
...
+set -e -o pipefail -u
+
+self_dir=$(cd "$(dirname "$0")" &>/dev/null; pwd -P)
+source "$self_dir"/utils.sh
+
+exit_code=0
+
+for file in $(git ls-files '*.astub'); do
+  output=$(comm -23   `: print the lines present in the first output but missing from the second` \
+    <(grep -Po '(?<=@)[a-zA-Z0-9.]*' "$file" | sort -u   `: all annotations used across the file`) \
```
PawelLipski, line 12: https://tldp.org/LDP/abs/html/process-sub.html ... note that's different from `$(...)`

scripts/enforce-explicit-annotation-import-in-astubs
```diff
@@ -0,0 +1,21 @@
...
+
+set -e -o pipefail -u
+
+self_dir=$(cd "$(dirname "$0")" &>/dev/null; pwd -P)
+source "$self_dir"/utils.sh
+
+exit_code=0
+
+for file in $(git ls-files '*.astub'); do
+  output=$(comm -23   `: print the lines present in the first output but missing from the second` \
```
PawelLipski, line 11: ``` `: ....` ``` is a tricky way to introduce comments in shell that won't reach the end of line (and thus include the trailing `\`)

Issue #153, opened 2020-03-30
Git Machete tab loads significantly longer than VCS Log tab for repositories with 5+ branches


PawelLipski:
~3-5 seconds when loading machete-sandbox repository (set up by https://github.com/PawelLipski/git-machete-blog/blob/master/sandbox-setup.sh)... would be nice to at least investigate the reasons for that delay (since it doesn't need to extract significantly more data/perform more computations than the Log tab).

PawelLipski:
@ up: https://github.com/typetools/checker-framework/issues/3253

PR #192, opened 2020-04-14
Reimplement merge-base using RevFilter.MERGE_BASE + reimplement is-ancestor using merge-base comparisons

PawelLipski:


PawelLipski:
#153 was mostly addressed by @mkondratek's #166; this PR only improves it a bit further, to the point where the plugin opens smoothly even for repos with large git history or complex git-machete branch layout.

gitCoreJGit/src/main/java/com/virtuslab/gitcore/impl/jgit/GitCoreRepository.java
```diff
@@ -149,24 +152,50 @@ private boolean isBranchMissing(String path) throws GitCoreException {
...
-  public boolean isAncestor(BaseGitCoreCommit presumedAncestor, BaseGitCoreCommit presumedDescendant)
-      throws GitCoreException {
+  @Nullable
+  private RevCommit deriveMergeBase(BaseGitCoreCommit a, BaseGitCoreCommit b) throws GitCoreException {
     RevWalk walk = new RevWalk(jgitRepo);
-    walk.sort(RevSort.TOPO);
+    walk.setRevFilter(RevFilter.MERGE_BASE);
     try {
-      ObjectId descendantObjectId = jgitRepo.resolve(presumedDescendant.getHash().getHashString());
+      ObjectId descendantObjectId = jgitRepo.resolve(a.getHash().getHashString());
```
PawelLipski, line 160: Var naming obsolete

gitCoreJGit/src/main/java/com/virtuslab/gitcore/impl/jgit/GitCoreRepository.java
```diff
@@ -149,24 +152,50 @@ private boolean isBranchMissing(String path) throws GitCoreException {
...
   }
+
+  private final java.util.Map<Tuple2<BaseGitCoreCommit, BaseGitCoreCommit>, @Nullable GitCoreCommitHash> mergeBaseCache = new java.util.HashMap<>();
+
+  @Nullable
+  private GitCoreCommitHash deriveMergeBaseIfNeeded(BaseGitCoreCommit a, BaseGitCoreCommit b) throws GitCoreException {
+    var abKey = Tuple.of(a, b);
+    var baKey = Tuple.of(b, a);
+    if (mergeBaseCache.containsKey(abKey)) {
+      return mergeBaseCache.get(abKey);
```
mkondratek, line 181: I think it is worth to notice here that:
https://stackoverflow.com/questions/4553624/hashmap-get-put-complexity

Of course, it would a minor improvement (supposing that no one changes `Map` implementation) but instead of `containsKey` we could just compare the `get` result with `null` (and reduce the complexity twice)

PawelLipski: Hmm the problem is that specific map is that values are `@Nullable` (to handle rare cases when merge-base can't be found for a specific pair of commits)... so `null` is a rare-yet-valid value and isn't equivalent to a missing key :/

Issue #158, opened 2020-04-01
IDE Fatal Error: Access is allowed from event dispatch thread only


PawelLipski:
Plugin version: `0.0.46-SNAPSHOT+git.b6b6f15`

.git directory: 
[git.zip](https://github.com/VirtusLab/git-machete-intellij-plugin/files/4416745/git.zip)

Offending action: nothing specifically, just looking at the Git Machete tab + toggling list commits; error seems to be happening spuriously.

Stacktrace:

```Access is allowed from event dispatch thread only.

com.intellij.openapi.diagnostic.RuntimeExceptionWithAttachments: EventQueue.isDispatchThread()=false Toolkit.getEventQueue()=com.intellij.ide.IdeEventQueue@1b2a97d2
Current thread: Thread[SE-SearchTask-ActionSearchEverywhereContributor,4,Idea Thread Group] 1728974396
SystemEventQueueThread: Thread[AWT-EventQueue-0,6,Idea Thread Group] 670192335
	at com.intellij.openapi.application.impl.ApplicationImpl.assertIsDispatchThread(ApplicationImpl.java:959)
	at com.intellij.openapi.application.impl.ApplicationImpl.assertIsDispatchThread(ApplicationImpl.java:948)
	at com.intellij.openapi.fileEditor.impl.FileEditorManagerImpl.assertDispatchThread(FileEditorManagerImpl.java:1540)
	at com.intellij.openapi.fileEditor.impl.FileEditorManagerImpl.getSelectedTextEditor(FileEditorManagerImpl.java:1232)
	at com.intellij.openapi.fileEditor.impl.FileEditorManagerImpl.getSelectedTextEditor(FileEditorManagerImpl.java:1227)
	at com.virtuslab.gitmachete.frontend.ui.table.GitMacheteGraphTable.getData(GitMacheteGraphTable.java:114)
	at com.intellij.ide.impl.DataManagerImpl.getDataFromProvider(DataManagerImpl.java:89)
	at com.intellij.ide.impl.DataManagerImpl.getData(DataManagerImpl.java:67)
	at com.intellij.ide.impl.DataManagerImpl.access$200(DataManagerImpl.java:46)
	at com.intellij.ide.impl.DataManagerImpl$MyDataContext.calcData(DataManagerImpl.java:386)
	at com.intellij.ide.impl.DataManagerImpl$MyDataContext.doGetData(DataManagerImpl.java:378)
[...]

Issue #142, opened 2020-03-24
FileEditorManager.getInstance invoked on a disposed project when closing the IDE


PawelLipski:
```
2020-03-24 12:14:37,058 [  45130]  ERROR - .impl.MessageBusConnectionImpl - Already disposed: Project (name=machete-sandbox, containerState=DISPOSE_COMPLETED, componentStore=/home/pawellipski/machete-sandbox)  
java.lang.AssertionError: Already disposed: Project (name=machete-sandbox, containerState=DISPOSE_COMPLETED, componentStore=/home/pawellipski/machete-sandbox) 
        at com.intellij.openapi.components.impl.ComponentManagerImpl.lambda$throwAlreadyDisposed$0(ComponentManagerImpl.java:153)
        at com.intellij.openapi.application.ReadAction.lambda$run$1(ReadAction.java:62)
        at com.intellij.openapi.application.impl.ApplicationImpl.runReadAction(ApplicationImpl.java:821)
        at com.intellij.openapi.application.ReadAction.compute(ReadAction.java:69)
        at com.intellij.openapi.application.ReadAction.run(ReadAction.java:62)
        at com.intellij.openapi.components.impl.ComponentManagerImpl.throwAlreadyDisposed(ComponentManagerImpl.java:151)
        at com.intellij.openapi.components.impl.ComponentManagerImpl.getPicoContainer(ComponentManagerImpl.java:144)
        at com.intellij.serviceContainer.PlatformComponentManagerImpl.getComponent(PlatformComponentManagerImpl.kt:265)
        at com.intellij.openapi.fileEditor.FileEditorManager.getInstance(FileEditorManager.java:21)
        at com.virtuslab.gitmachete.frontend.ui.table.GitMacheteGraphTable.getData(GitMacheteGraphTable.java:109)
        at com.intellij.ide.impl.DataManagerImpl.getDataFromProvider(DataManagerImpl.java:89)
        at com.intellij.ide.impl.DataManagerImpl.getData(DataManagerImpl.java:67)
        at com.intellij.ide.impl.DataManagerImpl.access$200(DataManagerImpl.java:46)
        at com.intellij.ide.impl.DataManagerImpl$MyDataContext.calcData(DataManagerImpl.java:386)
        at com.intellij.ide.impl.DataManagerImpl$MyDataContext.doGetData(DataManagerImpl.java:378)
[...]

PR #199, opened 2020-04-16
Remove CommonDataKeys.EDITOR from GitMacheteGraphTable keys 

mkondratek:
This is a fix for https://github.com/VirtusLab/git-machete-intellij-plugin/issues/142 and https://github.com/VirtusLab/git-machete-intellij-plugin/issues/158.
---
There are two and a half of interesting things here.

**The first** is that since https://github.com/VirtusLab/git-machete-intellij-plugin/pull/102/commits/c3429962b2486bbd756b2a09d53e3f96e21e841a 
 `com.virtuslab.gitmachete.frontend.ui.table.GitMacheteGraphTable` implements
`com.intellij.openapi.actionSystem.DataProvider`. This is related to data keys handling. The fact that is particularly thought-provoking is that `com.intellij.openapi.actionSystem.CommonDataKeys#EDITOR` appeared as one of handled keys even though there was no reference to it (in our codebase). 

**The half** is that https://github.com/VirtusLab/git-machete-intellij-plugin/pull/125/commits/ae919cf45b5eb45c17f56759b5860006ba6ecfbe. :smile: 

**The second**, related to issues #142 and #158 is why and how the same unused key is used. The answer to these questions can be found in `DataProvider` Javadoc:
```
/**
 * Allows a component hosting actions to provide context information to the actions. When a specific
 * data item is requested, the component hierarchy is walked up from the currently focused component,
 * and every component implementing the {@code DataProvider} interface is queried for the data
 * until one of them returns the data. Data items can also be mapped to each other - for example,
 * if a data provider provides an item for {@link CommonDataKeys#NAVIGATABLE}, an item for
 * {@link CommonDataKeys#NAVIGATABLE_ARRAY} can be generated from it automatically.
 *
 * @see DataContext
 */
```
**(...) every component implementing the {@code DataProvider} interface is queried for the data until one of them returns the data.**

And finally there is a class`com.intellij.usages.UsageTargetUtil` that contains a following line:
`Editor editor = CommonDataKeys.EDITOR.getData(dataProvider);`

My understanding of when this line is executed is equivalent to the meaning of `EDITOR` data key for `GitMacheteGraphTable`. (I suppose that may happen for different actions, eg. opening a file.)


PawelLipski:
@mkondratek since _the component hierarchy is walked up from the currently focused component_... actually, what's the `GitMacheteGraphTable` instance's parent in the component hierarchy? Is that anything that we manage or some VCS View stuff beyond our direct control?...

mkondratek:
@PawelLipski after a quick look at the stacktrace I suppose that the lowest common place is `com.intellij.ide.impl.DataManagerImpl$MyDataContext`. `DataManager` seems to be something that is a level higher than `DataProvider`. 

Note that there are also some other keys like `EDITOR` that hit `GitMacheteGraphTable#getData`:
`psi.File`, `psi.Element`, `selectedItem`, `virtualFile`, `psi.Element.array`, and more

PR #211, opened 2020-04-21
Use `hub pr` instead of `git fetch refs/pull/X/merge` to find base branch on CI + remove Gradle build cache from CI image

PawelLipski:


Dockerfile
```diff
@@ -4,22 +4,27 @@
...
-    `# workaround for error in debian-slim during openjdk installation` \
-    mkdir -p /usr/share/man/man1 && \
-    apt-get update && \
-    `# installing JDK and not just JRE to provide javadoc executable` \
-    apt-get install --no-install-recommends -y curl git openjdk-11-jdk-headless openssh-client python3 python3-pip xxd && \
-    pip3 install git-machete==2.13.6 && \
-    apt-get autoremove -y python3-pip && \
-    rm -rf /var/lib/apt/lists/*
+  `# workaround for error in debian-slim during openjdk installation` \
+  && mkdir -p /usr/share/man/man1 \
```
PawelLipski, line 8: `&&` on the beginning make it easier to spot a missing one than at the end

Dockerfile
```diff
@@ -4,22 +4,27 @@
...
+  && apt-get update \
+  `# installing JDK and not just JRE to provide javadoc executable` \
+  && apt-get install --no-install-recommends -y \
+    curl git hub openjdk-11-jdk-headless openssh-client python3 python3-pip xxd \
```
PawelLipski, line 12: Just added `hub` (GitHub CLI)

scripts/enforce-version-bump
```diff
@@ -8,28 +8,25 @@ source "$self_dir"/utils.sh
 if [[ ${CI_PULL_REQUEST-} ]]; then
   # In case of a PR build, CI_PULL_REQUEST should be a link of the form https://github.com/VirtusLab/git-machete-intellij-plugin/pull/123
   pr_num=${CI_PULL_REQUEST##*/}
-  # A hackish way to derive base commit of a PR on Github: first, fetch the ref pointing to PR head commit merged to the base,
-  # available on GH under refs/pull/<pr-number>/merge...
-  if git fetch --quiet origin refs/pull/$pr_num/merge:refs/remotes/origin/pull-$pr_num/merge; then
```
PawelLipski, line 13: This was well, hackish, and very prone to race conditions/other heisenbugs

Dockerfile
```diff
@@ -4,22 +4,27 @@
...
+  && pip3 install git-machete==2.13.6 \
+  && apt-get purge --autoremove -y python3-pip \
```
PawelLipski, line 14: `purge` also removes config files/data (unlike the vanilla `remove`/`autoremove` which only removes binaries, but retains the config in case one might need that when reinstalling in the future)

Issue #190, opened 2020-04-14
[Heisenbug] java.lang.IndexOutOfBoundsException in RepositoryGraph#getGraphElement


PawelLipski:
Plugin version: 0.0.66-SNAPSHOT (`./gradlew runIde` at commit `b17f6292ede`)

Project: [machete-sandbox.zip](https://github.com/VirtusLab/git-machete-intellij-plugin/files/4476976/machete-sandbox.zip)

Action: switching b/w repos from the dropdown; I couldn't reproduce it on another IDE run though (_heisenbug_).

Stack trace:

```
2020-04-14 20:10:02,303 [  95237]  ERROR - llij.ide.plugins.PluginManager - get(15) on List of length 13 
java.lang.IndexOutOfBoundsException: get(15) on List of length 13
        at io.vavr.collection.List.get(List.java:917)
        at com.virtuslab.gitmachete.frontend.graph.repository.RepositoryGraph.getGraphElement(RepositoryGraph.java:47)
        at com.virtuslab.gitmachete.frontend.ui.table.GraphTableModel.getValueAt(GraphTableModel.java:40)
        at java.desktop/javax.swing.JTable.getValueAt(JTable.java:2706)
        at java.desktop/javax.swing.JTable.prepareRenderer(JTable.java:5721)
        at com.intellij.ui.table.JBTable.prepareRenderer(JBTable.java:558)
        at java.desktop/javax.swing.plaf.basic.BasicTableUI.paintCell(BasicTableUI.java:2185)
        at java.desktop/javax.swing.plaf.basic.BasicTableUI.paintCells(BasicTableUI.java:2087)
        at java.desktop/javax.swing.plaf.basic.BasicTableUI.paint(BasicTableUI.java:1883)
        at java.desktop/javax.swing.plaf.ComponentUI.update(ComponentUI.java:161)
        at java.desktop/javax.swing.JComponent.paintComponent(JComponent.java:797)
        at com.intellij.ui.table.JBTable.paintComponent(JBTable.java:314)
        at java.desktop/javax.swing.JComponent.paint(JComponent.java:1074)
        at com.intellij.ui.table.JBTable.paint(JBTable.java:410)
        at java.desktop/javax.swing.JComponent.paintChildren(JComponent.java:907)
        at java.desktop/javax.swing.JComponent.paint(JComponent.java:1083)
        at java.desktop/javax.swing.JViewport.paint(JViewport.java:737)
[...]

PR #215, opened 2020-04-23
Handle repos creation & removal during runtime 

mkondratek:
- implement `DumbAware` in `RefreshGitMacheteStatusAction` to prevent its gray out during indiecies loading
- revamp  `VcsRootDropdown` action to `VcsRootComboBox` component
- handle subrepositories creation and removal (changes seen after refresh and according to vcs root mapping)


frontendUiTable/src/main/java/com/virtuslab/gitmachete/frontend/ui/table/GitMacheteGraphTableManager.java
```diff
@@ -74,7 +76,7 @@ private static IGitMacheteRepositoryFactory getGitMacheteRepositoryFactoryInstan
 
   private void subscribeToVcsRootChanges() {
     // The method reference is invoked when user changes repository in dropdown menu
```
PawelLipski, line 78: ? ;)
```suggestion
    // The method reference is invoked when user changes repository in combo box menu
```

frontendUiTable/src/main/java/com/virtuslab/gitmachete/frontend/ui/table/GitMacheteGraphTableManager.java
```diff
@@ -145,7 +147,16 @@ public void updateAndRefreshInBackground() {
         @Override
         @UIEffect
         public void run(ProgressIndicator indicator) {
-          GitRepository gitRepository = selectionChangeObservable.getValue();
+
+          // GitUtil.getRepositories(project) should never return empty list because it means there is no git repository
+          // in
```
PawelLipski, line 152: Fix formatting + `in an opened project`

frontendUiTable/src/main/java/com/virtuslab/gitmachete/frontend/ui/VcsRootComboBox.java
```diff
@@ -0,0 +1,90 @@
...
+    super.updateUI();
+    this.setVisible(getModel().getItems().size() > 1);
+  }
+
+  public void addObserver(ISelectionChangeObserver observer) {
+    observers.add(observer);
+  }
+
+  @UIEffect
+  public static JComponent shrinkWrap(JComponent component) {
```
PawelLipski, line 84: Hmm maybe sth like `createShrinkWrapper`? Also is it `shrinking` or rather `shrunk`?

frontendUiTable/src/main/java/com/virtuslab/gitmachete/frontend/ui/VcsRootComboBox.java
```diff
@@ -0,0 +1,90 @@
...
+   */
+  @UIEffect
+  public VcsRootComboBox(@MinLen(1) List<GitRepository> repositories) {
+    super(new MutableCollectionComboBoxModel<>(/* items */ repositories, /* selection */ repositories.get(0)));
+    setRenderer(SimpleListCellRenderer.create("", repo -> repo.getRoot().getName()));
+  }
+
+  @Override
+  public MutableCollectionComboBoxModel<GitRepository> getModel() {
+    return (MutableCollectionComboBoxModel<GitRepository>) super.getModel();
```
PawelLipski, line 37: Hmm maybe to avoid downcast I'd extract the `model` to a private final field

mkondratek: This would cause a situation where there is no appropriate model set (before `setModel` in the constructor). Furthermore, it seems to cause some strange `java.lang.IllegalArgumentException: dataModel must be non null`. I would prefer to keep the current solution.

PawelLipski: Apparently the runtime non-null check is generated by some JetBrains' annotation processor for `@org.jetbrains.annotations.NotNull`-marked params & results... which is kinda necessary since they don't have any means of consistently enforcing that non-nullness on the library consumers in the compile time.

PawelLipski: (at least wrt. params; non-null return values can be ofc always enforced in the compile time)

frontendUiTable/src/main/java/com/virtuslab/gitmachete/frontend/ui/VcsRootComboBox.java
mkondratek: renamed to `createShrinkingWrapper`

PawelLipski:
Incidentally, the commit https://github.com/VirtusLab/git-machete-intellij-plugin/pull/215/commits/5c8f2693c705bbde3666fe851385f360d5b9fd8f likely solved #190 as well

PR #214, opened 2020-04-21
Add logging and messaging

micpiotrowski:
Our repo logging is divided to topics/categories. At this moment we have:
- `backend` for `backendImpl` module
- `backendRoot` for `backendRoot` module
- `branchLayout` for `branchLayoutImpl` module
- `frontendActions` for `frontendActions` module
- `frontendRootUi` for `frontendRootUi` module
- `frontendUiTable` for `frontendUiTable` module
- `gitCore` for `gitCoreJGit` module
but we can flexibly changing this (with accuracy as to single class)

By default intelliJ logging everything with level `INFO` and above into `idea.log` file (localization depends on IntelliJ version and OS). To find where is this file located go to `Help` -> `Show Log in Files`.
To enable logging in `DEBUG` level we must add selected topics/categories to list in `Help` -> `Diagnostic Tools` -> `Debug Log Settings`. For our classes there are customized and provided above but by default topic/category is full class path for each class.

Also handling errors in more user friendly way is covered by this PR - in case of errors user should see error messagebox and VCS notification.
Unfortunately, something is still wrong - exceptions that should be catched and not rethrowed still goes to top of stacktrace :cry:

PR #216, opened 2020-04-23
Ensure a correct set of jars is packaged into the plugin zip

PawelLipski:
Related to changes on #214.


.circleci/config.yml
```diff
@@ -44,9 +44,14 @@ jobs:
       - run:
           name: Build plugin artifact
           command: ./gradlew buildPlugin
+      - run:
+          name: Verify plugin artifact contents
+          command: |
+            command -v zipinfo || { apt-get update && apt-get install -y unzip; }
```
PawelLipski, line 50: Can be removed once the CI image is regenerated, whenever that happens.

micpiotrowski: Done. Image regenerated and pushed.

PawelLipski: Thanks

PR #234, opened 2020-04-28
Bugfix/graph table scrolling

mkondratek:


frontendUiTable/src/main/java/com/virtuslab/gitmachete/frontend/ui/table/GitMacheteGraphTableManager.java
```diff
@@ -133,6 +133,7 @@ private void refreshGraphTable(Path macheteFilePath, boolean isMacheteFilePresen
     }
 
     gitMacheteGraphTable.repaint();
+    gitMacheteGraphTable.revalidate();
```
micpiotrowski, line 136: And who would have thought that that was the only thing :)

PR #235, opened 2020-04-28
Apply static verification for is.../as... downcast methods across the project

PawelLipski:
Essentially, if someone accidentally changed e.g. the following piece from `com.virtuslab.gitmachete.frontend.graph.impl.render.RenderPartColorIdProvider#getColorId`:

```java
    if (element.isNode()) {
      nodeIndex = element.asNode().getNodeIndex();
    } else { // isEdge
      nodeIndex = element.asEdge().getDownNodeIndex();
    }
```

into:

```java
    if (element.isNode()) {
      nodeIndex = element.asEdge().getDownNodeIndex(); // WHOOPS
    } else { // isEdge
      nodeIndex = element.asNode().getNodeIndex();
    }
```

then `SubtypingChecker` is gonna raise an error (quite cryptic one, alas):

```
/home/pawellipski/git-machete-intellij-plugin/frontendGraphImpl/src/main/java/com/virtuslab/gitmachete/frontend/graph/impl/render/RenderPartColorIdProvider.java:22: error: [[all, confirmedbranchitem, confirmedcommititem, confirmedgraphedge, confirmedgraphnode, confirmednonrootbranch, confirmedrootbranch, subtyping, subtypingbottom, subtypingtop]:contracts.precondition.not.satisfied] element.asEdge's precondition about 'element' is not satisfied
      nodeIndex = element.asEdge().getDownNodeIndex();
                                ^
/home/pawellipski/git-machete-intellij-plugin/frontendGraphImpl/src/main/java/com/virtuslab/gitmachete/frontend/graph/impl/render/RenderPartColorIdProvider.java:24: error: [[all, confirmedbranchitem, confirmedcommititem, confirmedgraphedge, confirmedgraphnode, confirmednonrootbranch, confirmedrootbranch, subtyping, subtypingbottom, subtypingtop]:contracts.precondition.not.satisfied] element.asNode's precondition about 'element' is not satisfied
      nodeIndex = element.asNode().getNodeIndex();
                                ^
```

### DIFF
--- a/frontendActions/build.gradle
+++ b/frontendActions/build.gradle
@@ -10,6 +10,8 @@ dependencies {
   implementation project(':frontendDataKeys')
 }
 
+applySubtypingChecker()
+
 apply plugin: 'org.jetbrains.intellij'
 disableIntellijTasks()
 intellij {
--- a/backendApi/build.gradle
+++ b/backendApi/build.gradle
@@ -1,5 +1,6 @@
 vavr()
 
 dependencies {
-    api project(":branchLayoutApi")
+  implementation project(':qual')
+  api project(':branchLayoutApi')
 }
--- a/backendApi/src/main/java/com/virtuslab/gitmachete/backend/api/BaseGitMacheteBranch.java
+++ b/backendApi/src/main/java/com/virtuslab/gitmachete/backend/api/BaseGitMacheteBranch.java
@@ -3,15 +3,30 @@ package com.virtuslab.gitmachete.backend.api;
 import io.vavr.collection.List;
 import io.vavr.control.Option;
 import org.checkerframework.checker.nullness.qual.Nullable;
+import org.checkerframework.framework.qual.EnsuresQualifierIf;
+import org.checkerframework.framework.qual.RequiresQualifier;
+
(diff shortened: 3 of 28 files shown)


frontendActions/build.gradle
```diff
@@ -10,6 +10,8 @@ dependencies {
   implementation project(':frontendDataKeys')
 }
 
+applySubtypingChecker()
```
PawelLipski, line 13: It's not really a problem if someone forgets about this line in build.gradle of a module that uses some `is...`/`as...` methods... coz then Checker is gonna spit out some weiiiird errors that I don't want to understand

PawelLipski:
Btw state-of-the-art solution for this problem is **pattern matching with exhaustiveness check**... but that's not widely supported in programming languages (well, Scala and probably some niche functional ones like Haskell/Ocaml etc.)

PawelLipski:
So we'd do:

```scala
val result = someBranch match {
  case b: GitMacheteRootBranch => /* do sth wth b */
  case b: GitMacheteNonRootBranch => /* do sth wth b */
// if we forget to cover any possible type, compiler's gonna scream at us
}
```

Issue #220, opened 2020-04-25
Ensure Actions are registered declaratively rather than programatically


mkondratek:
_Originally posted by @mkondratek in https://github.com/VirtusLab/git-machete-intellij-plugin/pull/217#discussion_r415101157_:

So there are two ways to register actions. Programatically (like in `com.virtuslab.gitmachete.frontend.ui.root.GitMachetePanel#createGitMacheteVerticalToolbar`) or declaratively in `plugin.xml`. 

Reference: 
https://www.jetbrains.org/intellij/sdk/docs/basics/action_system.html

If it comes to `GitMachetePanel#createGitMacheteVerticalToolbar`: the refresh action is (and was) implemented here. The slide-out action requires to call it after it is done. Current implementation and (probably too strict?) relation between the panel and refresh disallow registering the refresh action via plugin.xml. That is why I used in-code registering.

Actually, to avoid the re-registered action issue that happened here:

<details><summary>stack trace</summary>
<pre>
2020-04-24 17:31:09,286 [  44157]  ERROR - nSystem.impl.ActionManagerImpl - action with the ID "GitMachete.RefreshAction" was already registered. Action being registered is Refresh Status (Refresh status); Registered action is Refresh Status (Refresh status) 
java.lang.Throwable: action with the ID "GitMachete.RefreshAction" was already registered. Action being registered is Refresh Status (Refresh status); Registered action is Refresh Status (Refresh status)
        at com.intellij.openapi.diagnostic.Logger.error(Logger.java:146)
        at com.intellij.openapi.actionSystem.impl.ActionManagerImpl.reportActionError(ActionManagerImpl.java:400)
        at com.intellij.openapi.actionSystem.impl.ActionManagerImpl.reportActionError(ActionManagerImpl.java:389)
        at com.intellij.openapi.actionSystem.impl.ActionManagerImpl.registerChameleon(ActionManagerImpl.java:1244)
        at com.intellij.openapi.actionSystem.impl.ActionManagerImpl.addToMap(ActionManagerImpl.java:1217)
        at com.intellij.openapi.actionSystem.impl.ActionManagerImpl.registerAction(ActionManagerImpl.java:1188)
        at com.intellij.openapi.actionSystem.impl.ActionManagerImpl.registerAction(ActionManagerImpl.java:1180)
        at com.intellij.openapi.actionSystem.impl.ActionManagerImpl.registerAction(ActionManagerImpl.java:1255)
[...]

PR #236, opened 2020-04-29
Declare the entire toolbar action group in plugin.xml rather than programatically

PawelLipski:
Another (still not final!) step towards #220

### DIFF
--- a/frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/RebaseCurrentBranchOntoParentAction.java
+++ b/frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/RebaseCurrentBranchOntoParentAction.java
@@ -46,7 +46,7 @@ public class RebaseCurrentBranchOntoParentAction extends BaseRebaseBranchOntoPar
 
       } else {
         var upstreamBranch = currentBranchOption.get().asNonRootBranch().getUpstreamBranch();
-        presentation.setDescription("Rebase '${currentBranchOption.get().getName()}' onto ${upstreamBranch.getName()}");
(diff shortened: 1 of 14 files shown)


src/main/resources/META-INF/plugin.xml
```diff
@@ -20,40 +20,48 @@
...
-                class="com.virtuslab.gitmachete.frontend.actions.RebaseCurrentBranchOntoParentAction"
-                text="Rebase Current Branch Onto Parent"
-                description="Rebase current branch onto parent"
-                icon="AllIcons.Actions.Menu_cut"/>
+        <group id="GitMachete.Toolbar">
+            <action id="GitMachete.RefreshStatusAction"
+                    class="com.virtuslab.gitmachete.frontend.actions.RefreshStatusAction"
+                    text="Refresh Status"
+                    description="Refresh status"
+                    icon="AllIcons.Actions.Refresh"/>
```
mkondratek, line 46: If you add `use-shortcut-of="Refresh"` here we will be able to refresh our graph table with a key shortcut. (CTRL + F5)

PawelLipski: Done

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/RebaseCurrentBranchOntoParentAction.java
```diff
@@ -57,7 +57,7 @@ public void update(AnActionEvent anActionEvent) {
    */
   @Override
   public void actionPerformed(AnActionEvent anActionEvent) {
-    LOG.debug(() -> "Performing ${getClass().getSimpleName()}");
+    LOG.debug(() -> "Performing");
```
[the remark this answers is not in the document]
PawelLipski: Done on a follow-up PR

PR #238, opened 2020-04-29
Defensively check presence of required DataKeys within actions

PawelLipski:
Try e.g. clicking super-fast on the Refresh or Toggle List Commits button before and after this change ;)

My aim was to remove all but the really really safe assertions from `frontendActions` (only 2 are left, in `ActionUtils`)

### DIFF
--- a/frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/ActionUtils.java
+++ b/frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/ActionUtils.java
@@ -51,7 +51,7 @@ final class ActionUtils {
 
   static Option<BaseGitMacheteBranch> getSelectedMacheteBranch(AnActionEvent anActionEvent) {
     return getGitMacheteRepository(anActionEvent).flatMap(
-        repository -> getSelectedBranchName(anActionEvent).flatMap(branchName -> repository.getBranchByName(branchName)));
+        repository -> getSelectedBranchName(anActionEvent).flatMap(repository::getBranchByName));
   }
 
   static Option<GitRepository> getSelectedVcsRepository(AnActionEvent anActionEvent) {
--- a/frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/ActionUtils.java
+++ b/frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/ActionUtils.java
@@ -1,57 +1,60 @@
 package com.virtuslab.gitmachete.frontend.actions;
 
+import java.nio.file.Path;
+
 import com.intellij.openapi.actionSystem.AnActionEvent;
+import com.intellij.openapi.project.Project;
 import git4idea.repo.GitRepository;
 import io.vavr.control.Option;
 
+import com.virtuslab.branchlayout.api.IBranchLayout;
 import com.virtuslab.gitmachete.backend.api.BaseGitMacheteBranch;
 import com.virtuslab.gitmachete.backend.api.BaseGitMacheteNonRootBranch;
 import com.virtuslab.gitmachete.backend.api.IGitMacheteRepository;
 import com.virtuslab.gitmachete.frontend.datakeys.DataKeys;
 import com.virtuslab.gitmachete.frontend.ui.api.table.IGraphTableManager;
 
-public final class ActionUtils {
+final class ActionUtils {
 
   private ActionUtils() {}
 
-  static GitRepository getPresentIdeaRepository(AnActionEvent anActionEvent) {
-    GitRepository repository = anActionEvent.getData(DataKeys.KEY_SELECTED_VCS_REPOSITORY);
-    assert repository != null : "Can't get selected GitRepository";
-    return repository;
+  static Option<IBranchLayout> getBranchLayout(AnActionEvent anActionEvent) {
+    return Option.of(anActionEvent.getData(DataKeys.KEY_BRANCH_LAYOUT));
   }
 
-  static IGraphTableManager getPresentGraphTableManager(AnActionEvent anActionEvent) {
-    IGraphTableManager graphTableManager = anActionEvent.getData(DataKeys.KEY_GRAPH_TABLE_MANAGER);
-    assert graphTableManager != null : "Can't get graph table manager";
-    return graphTableManager;
+  static Option<BaseGitMacheteNonRootBranch> getCurrentMacheteNonRootBranch(AnActionEvent anActionEvent) {
+    return getGitMacheteRepository(anActionEvent).flatMap(repository -> repository.getCurrentBranchIfManaged().flatMap(
+        currentBranch -> currentBranch.isNonRootBranch() ? Option.some(currentBranch.asNonRootBranch()) : Option.none()));
   }
 
-  /**
-   * This method relies on key `KEY_GIT_MACHETE_REPOSITORY` corresponding to a non-null value
-   * and hence must always be called after checking the git machete repository readiness.
-   * See {@link BaseRebaseBranchOntoParentAction#update} and {@link DataKeys#KEY_IS_GIT_MACHETE_REPOSITORY_READY}.
-   */
-  static IGitMacheteRepository getPresentMacheteRepository(AnActionEvent anActionEvent) {
-    IGitMacheteRepository gitMacheteRepository = anActionEvent.getData(DataKeys.KEY_GIT_MACHETE_REPOSITORY);
-    assert gitMacheteRepository != null : "Can't get gitMacheteRepository";
+  static Option<IGitMacheteRepository> getGitMacheteRepository(AnActionEvent anActionEvent) {
+    return Option.of(anActionEvent.getData(DataKeys.KEY_GIT_MACHETE_REPOSITORY));
+  }
 
-    return gitMacheteRepository;
+  static Option<Path> getGitMacheteFilePath(AnActionEvent anActionEvent) {
+    return Option.of(anActionEvent.getData(DataKeys.KEY_GIT_MACHETE_FILE_PATH));
+  }
+
+  static IGraphTableManager getGraphTableManager(AnActionEvent anActionEvent) {
+    return anActionEvent.getData(DataKeys.KEY_GRAPH_TABLE_MANAGER);
+  }
+
+  static Project getProject(AnActionEvent anActionEvent) {
+    var project = anActionEvent.getProject();
+    assert project != null : "Can't get project from action event";
+    return project;
+  }
+
+  static Option<String> getSelectedBranchName(AnActionEvent anActionEvent) {
+    return Option.of(anActionEvent.getData(DataKeys.KEY_SELECTED_BRANCH_NAME));
   }
 
   static Option<BaseGitMacheteBranch> getSelectedMacheteBranch(AnActionEvent anActionEvent) {
-    IGitMacheteRepository gitMacheteRepository = getPresentMacheteRepository(anActionEvent);
-    String selectedBranchName = anActionEvent.getData(DataKeys.KEY_SELECTED_BRANCH_NAME);
-    assert selectedBranchName != null : "Can't get selected branch";
-    return gitMacheteRepository.getBranchByName(selectedBranchName);
+    return getGitMacheteRepository(anActionEvent).flatMap(
+        repository -> getSelectedBranchName(anActionEvent).flatMap(branchName -> repository.getBranchByName(branchName)));
   }
 
-  static BaseGitMacheteNonRootBranch getCurrentBaseMacheteNonRootBranch(AnActionEvent anActionEvent) {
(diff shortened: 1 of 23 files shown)


frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/BaseSlideOutBranchAction.java
```diff
@@ -59,22 +50,20 @@ public BaseSlideOutBranchAction() {}
...
-    assert branchLayout != null : "Can't get branch layout";
-
-    var branchName = branchToSlideOut.getName();
+    LOG.debug(() -> "Entering: branchToSlideOut = ${branchToSlideOut}");
+    String branchName = branchToSlideOut.getName();
+    Project project = ActionUtils.getProject(anActionEvent);
+    var branchLayout = ActionUtils.getBranchLayout(anActionEvent);
+    var gitMacheteFilePath = ActionUtils.getGitMacheteFilePath(anActionEvent);
+    if (branchLayout.isEmpty() || gitMacheteFilePath.isEmpty()) {
+      return;
```
micpiotrowski, line 59: Maybe some log or even noticeable message for user in that cases? The same for other actions.

frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/GitMacheteGraphTableManager.java
```diff
@@ -203,8 +207,14 @@ private void updateRepository(Path mainDirectoryPath, Path gitDirectoryPath, boo
...
-        IBranchLayout branchLayout = updateBranchLayoutAndMacheteFilePath();
-        gitMacheteRepositoryRef.set(gitMacheteRepositoryFactory.create(mainDirectoryPath, gitDirectoryPath, branchLayout));
+        GitRepository gitRepository = gitRepositorySelectionProvider.getSelectedRepository();
+        if (gitRepository != null) {
+          Path macheteFilePath = getMacheteFilePath(gitRepository);
+          IBranchLayout branchLayout = createBranchLayout(macheteFilePath);
+          graphTable.setBranchLayout(branchLayout);
+          graphTable.setMacheteFilePath(macheteFilePath);
+          gitMacheteRepositoryRef.set(gitMacheteRepositoryFactory.create(mainDirectoryPath, gitDirectoryPath, branchLayout));
+        }
```
micpiotrowski, line 217: Maybe `else` with some logging (and maybe even message for user) in case when something unexpected (like `gitRepository == null`) will happen?

frontendUiApi/src/main/java/com/virtuslab/gitmachete/frontend/ui/api/root/IGitRepositorySelectionProvider.java
```diff
@@ -3,9 +3,11 @@
 import git4idea.repo.GitRepository;
 import io.vavr.collection.List;
 import org.checkerframework.checker.guieffect.qual.UIEffect;
+import org.checkerframework.checker.nullness.qual.Nullable;
 import org.checkerframework.common.value.qual.MinLen;
 
 public interface IGitRepositorySelectionProvider {
+  @Nullable
```
mkondratek, line 10: Why? :thinking: We can rely on the fact that our plugin is visible only if at least one repository exists, can't we?

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/ActionUtils.java
```diff
@@ -1,57 +1,60 @@
...
-    assert currentBranchOption.isDefined() : "Can't get current branch";
-    var baseGitMacheteBranch = currentBranchOption.get();
-    assert baseGitMacheteBranch.isNonRootBranch() : "Selected branch is a root branch";
+  static Option<String> getSelectedBranchName(AnActionEvent anActionEvent) {
+    return Option.of(anActionEvent.getData(DataKeys.KEY_SELECTED_BRANCH_NAME));
+  }
+
+  static Option<BaseGitMacheteBranch> getSelectedMacheteBranch(AnActionEvent anActionEvent) {
+    return getGitMacheteRepository(anActionEvent).flatMap(
+        repository -> getSelectedBranchName(anActionEvent).flatMap(branchName -> repository.getBranchByName(branchName)));
```
mkondratek, line 54: You can make this line shorter by method reference usage :shorts:

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/CheckOutBranchAction.java
```diff
@@ -54,20 +53,21 @@ public void actionPerformed(AnActionEvent anActionEvent) {
       return;
```
mkondratek, line 53: So why don't we perform this check in the `update` making an action` impossible to perform in the first place?

frontendDataKeys/src/main/java/com/virtuslab/gitmachete/frontend/datakeys/DataKeys.java
```diff
@@ -18,14 +19,13 @@ private DataKeys() {}
 
   public static final DataKey<IBranchLayout> KEY_BRANCH_LAYOUT = DataKey.create("BRANCH_LAYOUT");
   public static final DataKey<Path> KEY_GIT_MACHETE_FILE_PATH = DataKey.create("GIT_MACHETE_FILE_PATH");
-  public static final DataKey<IGraphTableManager> KEY_GRAPH_TABLE_MANAGER = DataKey.create("GRAPH_TABLE_MANAGER");
   public static final DataKey<IGitMacheteRepository> KEY_GIT_MACHETE_REPOSITORY = DataKey.create("GIT_MACHETE_REPOSITORY");
-  public static final DataKey<Boolean> KEY_IS_GIT_MACHETE_REPOSITORY_READY = DataKey.create("IS_GIT_MACHETE_REPOSITORY_READY");
+  /** This key must always be available in the container hierarchy, and a DataProvider must always return a non-null value. */
+  public static final DataKey<@NonNull IGraphTableManager> KEY_GRAPH_TABLE_MANAGER = DataKey.create("GRAPH_TABLE_MANAGER");
```
mkondratek, line 24: Is `@NonNull` required here?

PawelLipski: Nullness check behaves really weird with generics, esp. the ones defined outside of Checker-processed codebase... I'm just adding annotations defensively ;D without that `@NonNull` (and. esp without `@Nullable` for all other `DataKeys` here, which I've also added), Checker isn't going to raise certain errors and/or will raise them where unneeded :/

frontendUiApi/src/main/java/com/virtuslab/gitmachete/frontend/ui/api/root/IGitRepositorySelectionProvider.java
PawelLipski: Changed to `Option` actually; see the updated codebase for rationale (the underlying `com.intellij.ui.CollectionComboBoxModel#getSelected` returns a nullable value :/)

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/CheckOutBranchAction.java
PawelLipski: Done, added a (rather defensive) check in `update` as well

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/BaseSlideOutBranchAction.java
PawelLipski: Hmm TBH I'm reluctant to pollute our action codebase with more code for handling extraordinary situations - I'd realistically expect any of this keys to be empty only in case of some nasty race conditions. I don't even think we should bother the user, but rather let them click again even it happens.

PawelLipski: But if it comes to some WARN-level logging unnoticable for user - I think it's okay, I'm gonna add that now

frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/GitMacheteGraphTableManager.java
PawelLipski: Done, added just a `LOG.warn` (this shouldn't be displayed to do user, right?)

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/CheckOutBranchAction.java
```diff
@@ -34,18 +33,22 @@
...
   @Override
   @UIEffect
   public void update(AnActionEvent anActionEvent) {
     super.update(anActionEvent);
+
+    var selectedBranchName = ActionUtils.getSelectedBranchName(anActionEvent);
+    var presentation = anActionEvent.getPresentation();
+    // It's very unlikely that selectedBranchName is empty at this point since it's assigned directly before invoking this
+    // action in GitMacheteGraphTable.GitMacheteGraphTableMouseAdapter.mouseClicked; still, it's better to be safe.
+    presentation.setEnabled(selectedBranchName.isDefined());
```
mkondratek, line 45: or `setEnabledAndVisible`?

logging/src/main/java/com/virtuslab/logger/PrefixedLambdaLogger.java
```diff
@@ -60,38 +60,18 @@ public void info(String format) {
     logger.info(() -> getLogMessagePrefix() + format);
   }
 
-  @Override
-  public void info(Supplier<String> msgSupplier) {
```
PawelLipski, line 64: @micpiotrowski to confirm, but I don't think there's any benefit of having lazy log string evaluation for levels above DEBUG; coz IIUC IntelliJ sets the default level to INFO for all topic anyway

frontendDataKeys/src/main/java/com/virtuslab/gitmachete/frontend/datakeys/DataKeys.java
PawelLipski: I tried to circumvent that somehow with another custom subtyping hierarchy, but this time to no avail :/

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/BaseRebaseBranchOntoParentAction.java
```diff
@@ -94,31 +89,35 @@ public void update(AnActionEvent anActionEvent) {
...
-
-    doRebase(project, macheteRepository, gitRepository, branchToRebase);
+    Project project = ActionUtils.getProject(anActionEvent);
+    Option<IGitMacheteRepository> gitMacheteRepository = getGitMacheteRepository(anActionEvent);
+    Option<GitRepository> gitRepository = getSelectedVcsRepository(anActionEvent);
+
+    if (gitMacheteRepository.isDefined() && gitRepository.isDefined()) {
+      doRebase(project, gitMacheteRepository.get(), gitRepository.get(), branchToRebase);
+    } else {
+      LOG.warn("Skipping the action because gitMacheteRepository and/or gitRepository is empty");
```
mkondratek, line 99: `empty` is misleading. It is understandable here but in log level would be ambiguous. Maybe `undefined`?

PawelLipski: Done, fixed everywhere

frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/GitMacheteGraphTableManager.java
micpiotrowski: You are right. It's only displayed in logs. In general all logging methods are not directly visible for user.

logging/src/main/java/com/virtuslab/logger/PrefixedLambdaLogger.java
micpiotrowski: But it also doesn't bother us and maybe in future default loglevel in IntelliJ will be changed.
But ok, it also can be deleted: use your judgement.

PawelLipski: Restored by yourself on #242, let's consider it resolved then

PR #242, opened 2020-04-29
Logging improvements

micpiotrowski:
- From now logging in error level with stacktrace will not been interpreting by IntelliJ as error in plugin.
- Error handling in `GitMacheteGraphTableManager` was reverted to way that was before + small improvements was added

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -214,10 +214,25 @@ subprojects {
   ext {
     applyIntellijPlugin = { Map params ->
       apply plugin: 'org.jetbrains.intellij'
-      disableIntellijTasks()
+
+      // For the frontend subprojects we only use the Intellij plugin to provide dependencies,
+      // but don't want the associated tasks to be available; they should only be available in the root project.
+      buildPlugin.enabled = false
+      buildSearchableOptions.enabled = false
+      downloadRobotServerPlugin.enabled = false
+      jarSearchableOptions.enabled = false
+      patchPluginXml.enabled = false
+      prepareSandbox.enabled = false
+      prepareTestingSandbox.enabled = false
+      prepareUiTestingSandbox.enabled = false
+      publishPlugin.enabled = false
+      runIde.enabled = false
+      runIdeForUiTests.enabled = false
+      verifyPlugin.enabled = false
+
       intellij {
         version intellijVersion
-        if(params.withGit4idea) {
+        if (params.withGit4Idea) {
           plugins 'git4idea'
         }
       }
@@ -233,23 +248,6 @@ subprojects {
         def qualClassDir = project(':qual').sourceSets.main.output.classesDirs.asPath
         extraJavacArgs += "-AqualDirs=${qualClassDir}"
       }
-    }
-
-    // For the frontend subprojects we only use the Intellij plugin to provide dependencies,
-    // but don't want the associated tasks to be available; they should only be available in the root project.
-    disableIntellijTasks = { ->
-      buildPlugin.enabled = false
-      buildSearchableOptions.enabled = false
-      downloadRobotServerPlugin.enabled = false
-      jarSearchableOptions.enabled = false
-      patchPluginXml.enabled = false
-      prepareSandbox.enabled = false
-      prepareTestingSandbox.enabled = false
-      prepareUiTestingSandbox.enabled = false
-      publishPlugin.enabled = false
-      runIde.enabled = false
-      runIdeForUiTests.enabled = false
-      verifyPlugin.enabled = false
     }
 
     jgit = { ->
--- a/build.gradle
+++ b/build.gradle
@@ -212,6 +212,17 @@ subprojects {
 
 
   ext {
+    applyJetBrainsPlugin = {withGit4ide ->
+      apply plugin: 'org.jetbrains.intellij'
+      disableIntellijTasks()
+      intellij {
+        version intellijVersion
+        if(withGit4ide) {
+          plugins 'git4idea'
+        }
+      }
+    }
+
     applySubtypingChecker = { ->
       dependencies {
         implementation project(':qual')
--- a/build.gradle
+++ b/build.gradle
@@ -212,12 +212,12 @@ subprojects {
 
 
   ext {
-    applyJetBrainsPlugin = {withGit4ide ->
+    applyIntellijPlugin = { Map params ->
       apply plugin: 'org.jetbrains.intellij'
       disableIntellijTasks()
       intellij {
         version intellijVersion
-        if(withGit4ide) {
+        if(params.withGit4idea) {
           plugins 'git4idea'
         }
       }
--- a/frontendActions/build.gradle
+++ b/frontendActions/build.gradle
@@ -10,5 +10,5 @@ dependencies {
   implementation project(':frontendDataKeys')
 }
 
-applyIntellijPlugin(withGit4idea: true)
+applyIntellijPlugin(withGit4Idea: true)
 applySubtypingChecker()
--- a/frontendActions/build.gradle
+++ b/frontendActions/build.gradle
@@ -10,5 +10,5 @@ dependencies {
   implementation project(':frontendDataKeys')
 }
 
-applyJetBrainsPlugin(/* withGit4idea */ true)
+applyIntellijPlugin(withGit4idea: true)
 applySubtypingChecker()
--- a/frontendActions/build.gradle
+++ b/frontendActions/build.gradle
@@ -10,11 +10,5 @@ dependencies {
   implementation project(':frontendDataKeys')
 }
 
+applyJetBrainsPlugin(/* withGit4idea */ true)
 applySubtypingChecker()
-
-apply plugin: 'org.jetbrains.intellij'
-disableIntellijTasks()
-intellij {
-  version intellijVersion
-  plugins 'git4idea'
-}
--- a/frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/GitMacheteGraphTableManager.java
+++ b/frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/GitMacheteGraphTableManager.java
@@ -198,7 +198,6 @@ public final class GitMacheteGraphTableManager implements IGraphTableManager {
    * Updates repository which is the base of graph table model. The change will be seen after
    * {@link GitMacheteGraphTableManager#refreshGraphTable()}.
    */
-  @SuppressWarnings({"IllegalCatch", "subtyping"})
   private void updateRepository(Path mainDirectoryPath, Path gitDirectoryPath, boolean isMacheteFilePresent) {
     LOG.debug(() -> "Entering: mainDirectoryPath = ${mainDirectoryPath}, gitDirectoryPath = ${gitDirectoryPath}" +
         "isMacheteFilePresent = ${isMacheteFilePresent}");
@@ -228,12 +227,10 @@ public final class GitMacheteGraphTableManager implements IGraphTableManager {
   }
 
   private void handleUpdateRepositoryExceptions(Throwable t) {
-    // Only for silencing checkstyle
-    Throwable cause = t;
-
-    LOG.error("Unable to create Git Machete repository", cause);
+    LOG.error("Unable to create Git Machete repository", t);
 
     // Getting the innermost exception since it's usually the primary cause that gives most valuable message
+    Throwable cause = t;
     while (cause.getCause() != null) {
       cause = cause.getCause();
     }
--- a/frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/GitMacheteGraphTableManager.java
+++ b/frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/GitMacheteGraphTableManager.java
@@ -37,7 +37,7 @@ import com.virtuslab.gitmachete.frontend.graph.api.repository.IRepositoryGraphFa
 import com.virtuslab.gitmachete.frontend.ui.api.root.IGitRepositorySelectionProvider;
 import com.virtuslab.gitmachete.frontend.ui.api.table.IGraphTableManager;
 import com.virtuslab.logger.IPrefixedLambdaLogger;
-import com.virtuslab.logger.LoggingUtils;
+import com.virtuslab.logger.IntelliJLoggingUtils;
 import com.virtuslab.logger.PrefixedLambdaLoggerFactory;
 
 public final class GitMacheteGraphTableManager implements IGraphTableManager {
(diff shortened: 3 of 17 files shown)


frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/GitMacheteGraphTableManager.java
```diff
@@ -197,36 +197,47 @@ public void run(ProgressIndicator indicator) {
    * Updates repository which is the base of graph table model. The change will be seen after
    * {@link GitMacheteGraphTableManager#refreshGraphTable()}.
    */
-  @SuppressWarnings({"IllegalCatch"})
+  @SuppressWarnings({"IllegalCatch", "subtyping"})
```
PawelLipski, line 200: Hmm why `subtyping` suppression is necessary here? :thinking:

frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/GitMacheteGraphTableManager.java
```diff
@@ -197,36 +197,47 @@ public void run(ProgressIndicator indicator) {
...
-            () -> Messages.showErrorDialog(
-                exceptionMessage != null
-                    ? exceptionMessage
-                    : "Repository instantiation failed. For more information, please look at the IntelliJ logs",
-                "Something Went Wrong..."),
-            ModalityState.NON_MODAL);
+          return gitMacheteRepositoryFactory.create(mainDirectoryPath, gitDirectoryPath, branchLayout);
+        })
+            .onSuccess(repository -> gitMacheteRepositoryRef.set(repository))
+            .onFailure(cause -> {
```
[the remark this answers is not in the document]
PawelLipski: Yeah, possibly some of this logic (`VcsNotifier....notifyError`, `GuiUtils.invokeLaterIfNeeded(() -> Messages.showErrorDialog...` could even be extracted to a utility class

logging/src/main/java/com/virtuslab/logger/PrefixedLambdaLogger.java
```diff
@@ -35,6 +37,12 @@ private String getLogMessagePrefix() {
     return className + "#" + element.getMethodName() + ": ";
   }
 
+  private String throwableToString(Throwable t) {
```
PawelLipski, line 40: `getStackTraceAsString(t)`?

logging/src/main/java/com/virtuslab/logger/PrefixedLambdaLogger.java
```diff
@@ -87,11 +95,11 @@ public void error(Supplier<String> msgSupplier) {
 
   @Override
   public void error(String format, Throwable t) {
-    logger.error(() -> getLogMessagePrefix() + format, t);
+    logger.error(() -> getLogMessagePrefix() + format + System.lineSeparator() + throwableToString(t));
```
PawelLipski, line 98: Isn't "\n" equivalent to `System.lineSeparator()`? Although not sure how about Windows where EOL is represented with 2 chars :thinking:

frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/GitMacheteGraphTableManager.java
micpiotrowski: Coz I wasn't be able to do `cause.getCause()` without this

PawelLipski: Hmm and what was the error? :hushed:

frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/GitMacheteGraphTableManager.java
```diff
@@ -197,36 +197,47 @@ public void run(ProgressIndicator indicator) {
...
+              LOG.error("Unable to create Git Machete repository", cause);
+
+              // Getting most inner exception to printout only proper message later
```
PawelLipski, line 221: (?)
```suggestion
              // Getting the innermost exception since it's usually the primary cause that gives the most valuable message
```

frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/GitMacheteGraphTableManager.java
PawelLipski: Ok, I'm looking into that now, possibly some defaults are missing in `SubtypingTop`

PawelLipski: Ok, fixed, just remove `@SuppressWarnings` completely here and remove `@DefaultFor` from `SubtypingBottom`:

```diff
diff --git a/frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/GitMacheteGraphTableManager.java b/frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/GitMacheteGraphTableManager.java
index d77cd3d6..93e32517 100644
--- a/frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/GitMacheteGraphTableManager.java
+++ b/frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/GitMacheteGraphTableManager.java
@@ -197,7 +197,6 @@ public final class GitMacheteGraphTableManager implements IGraphTableManager {
    * Updates repository which is the base of graph table model. The change will be seen after
    * {@link GitMacheteGraphTableManager#refreshGraphTable()}.
    */
-  @SuppressWarnings({"IllegalCatch", "subtyping"})
   private void updateRepository(Path mainDirectoryPath, Path gitDirectoryPath, boolean isMacheteFilePresent) {
     LOG.debug(() -> "Entering: mainDirectoryPath = ${mainDirectoryPath}, gitDirectoryPath = ${gitDirectoryPath}" +
         "isMacheteFilePresent = ${isMacheteFilePresent}");
diff --git a/qual/src/main/java/com/virtuslab/qual/internal/SubtypingBottom.java b/qual/src/main/java/com/virtuslab/qual/internal/SubtypingBottom.java
index 118454ba..d179961f 100644
--- a/qual/src/main/java/com/virtuslab/qual/internal/SubtypingBottom.java
+++ b/qual/src/main/java/com/virtuslab/qual/internal/SubtypingBottom.java
@@ -3,7 +3,6 @@ package com.virtuslab.qual.internal;
 import java.lang.annotation.ElementType;
 import java.lang.annotation.Target;
 
-import org.checkerframework.framework.qual.DefaultFor;
 import org.checkerframework.framework.qual.SubtypeOf;
 import org.checkerframework.framework.qual.TargetLocations;
 import org.checkerframework.framework.qual.TypeUseLocation;
@@ -28,5 +27,4 @@ import com.virtuslab.qual.gitmachete.frontend.graph.api.items.ConfirmedCommitIte
     ConfirmedBranchItem.class,
     ConfirmedCommitItem.class,
 })
-@DefaultFor(TypeUseLocation.LOWER_BOUND)
 public @interface SubtypingBottom {}
```

PawelLipski: Apparently, with that default, Checker was assuming that `Consumer`'s type param is always `@SubtypingBottom`, which doesn't make much sense

PawelLipski: But actually... there's also a different style problem here - I'd rather not re-assign anything to lambda param, preferably make a separate var for the use in `while`

logging/src/main/java/com/virtuslab/logger/PrefixedLambdaLogger.java
PawelLipski: Would be nice to check, I'm pretty curious actually how Java handles that

frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/GitMacheteGraphTableManager.java
```diff
@@ -197,36 +197,47 @@ public void run(ProgressIndicator indicator) {
...
-        GuiUtils.invokeLaterIfNeeded(
-            () -> Messages.showErrorDialog(
-                exceptionMessage != null
-                    ? exceptionMessage
-                    : "Repository instantiation failed. For more information, please look at the IntelliJ logs",
-                "Something Went Wrong..."),
-            ModalityState.NON_MODAL);
+          return gitMacheteRepositoryFactory.create(mainDirectoryPath, gitDirectoryPath, branchLayout);
+        })
+            .onSuccess(repository -> gitMacheteRepositoryRef.set(repository))
```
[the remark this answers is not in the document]
micpiotrowski: :+1: I just followed @PawelLipski's tips regarding lambdas and method references  ;)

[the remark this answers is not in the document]
PawelLipski: Hmmm wait what specific tip do you mean? :thinking:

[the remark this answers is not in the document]
PawelLipski: I only meant to avoid lambdas for **nullary** methods coz that's kinda confusing... but as for unary methods, lambdas FTW

logging/src/main/java/com/virtuslab/logger/PrefixedLambdaLogger.java
micpiotrowski: IIRC `System.lineSeparator()` is universal line separator that will be expanded to proper line separator regarding OS

PawelLipski: Ok, right https://stackoverflow.com/a/33505921/2750812; forbid `\n` in checkstyle then maybe

PawelLipski: Looks that `\n` resolves to a single line feed/`0x0a` byte on each OS, incl. Windows

logging/src/main/java/com/virtuslab/logger/PrefixedLambdaLogger.java
```diff
@@ -23,7 +23,7 @@
     className = classPath[classPath.length - 1];
   }
 
-  private String getLogMessagePrefix() {
+  private String getStackTraceAsString() {
```
PawelLipski, line 26: Whoops not this method should have been renamed?

frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/GitMacheteGraphTableManager.java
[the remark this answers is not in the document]
micpiotrowski: Ok, so sorry, maybe I misunderstood you

logging/src/main/java/com/virtuslab/logger/PrefixedLambdaLogger.java
micpiotrowski: Whoops...

frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/GitMacheteGraphTableManager.java
micpiotrowski: It seems to work  :tada:

build.gradle
```diff
@@ -212,6 +212,17 @@ subprojects {
 
 
   ext {
+    applyJetBrainsPlugin = {withGit4ide ->
```
PawelLipski, line 215: Nit: missing space after `{`

PawelLipski: Nice, a helper like that was missing :)

PawelLipski: Actually, a more idiomatic Groovy would be sth like: `applyJetBrainsPlugin = { params ->`, where `params` is a Map; then you'd need to do `if (params.withGit4Idea) {` below and `applyJetBrainsPlugin(withGit4Idea: true)`

PawelLipski: https://mrhaki.blogspot.com/2015/09/groovy-goodness-turn-method-parameters.html

Not sure though if this syntax works with those `ext`'s lambdas, maybe only with regular methods

frontendActions/build.gradle
```diff
@@ -10,11 +10,5 @@ dependencies {
   implementation project(':frontendDataKeys')
 }
 
+applyJetBrainsPlugin(/* withGit4idea */ true)
```
PawelLipski, line 13: Actually, it should be called `applyIntellijPlugin`

logging/build.gradle
```diff
@@ -1 +1,3 @@
 slf4jLambdaApi()
+
+applyJetBrainsPlugin(/* withGit4idea */ false)
```
PawelLipski, line 3: Just make sure that this plugin doesn't expose any `com.intellij` dependencies as `api` (coz otherwise each module that depends on `logging` will transitively see `com.intellij` stuff, which we don't want to be the case in backend)

build.gradle
micpiotrowski: Nice! Applied

logging/build.gradle
micpiotrowski: It seems like non of `com.intellij` packages are available in `backend*` modules

frontendActions/build.gradle
```diff
@@ -10,11 +10,5 @@ dependencies {
   implementation project(':frontendDataKeys')
 }
 
+applyIntellijPlugin(withGit4idea: true)
```
PawelLipski, line 13: Nit: uppercase `Idea`

build.gradle
```diff
@@ -212,6 +212,17 @@ subprojects {
 
 
   ext {
+    applyIntellijPlugin = { Map params ->
+      apply plugin: 'org.jetbrains.intellij'
+      disableIntellijTasks()
```
PawelLipski, line 217: If `disableIntellijTasks` are only called here (which I think is now the case), then consider inlining

build.gradle
```diff
@@ -212,6 +212,17 @@ subprojects {
...
+      intellij {
+        version intellijVersion
+        if(params.withGit4idea) {
```
PawelLipski, line 220: Nit: add space

logging/src/main/java/com/virtuslab/logger/LoggingUtils.java
```diff
@@ -0,0 +1,17 @@
+package com.virtuslab.logger;
+
+import com.intellij.openapi.application.ModalityState;
+import com.intellij.openapi.ui.Messages;
+import com.intellij.ui.GuiUtils;
+
+public final class LoggingUtils {
```
PawelLipski, line 7: Maybe rename to `IntelliJLoggingUtils` for clarity

Issue #104, opened 2020-02-26
Agree one name of our product for end user


micpiotrowski:
In different places we are using different name of our product (`git-machete`, `Git Machete`, `Git-Machete`, and so on) esp. in logging messages and exceptions. We must agree one name and change to this name every end user's message.

PawelLipski:
Decided on `Git Machete` (Title Case, as recommended in https://www.jetbrains.org/intellij/sdk/docs/basics/plugin_structure/plugin_configuration_file.html)

PR #249, opened 2020-05-01
Extract user-facing strings from plugin.xml to a resource bundle

PawelLipski:
Actually, #104 is already resolved for some time, in favor of `Git Machete` (Title Case, as recommended in https://www.jetbrains.org/intellij/sdk/docs/basics/plugin_structure/plugin_configuration_file.html)

### DIFF
--- a/frontendActions/src/main/resources/actions/ResourceBundle.properties
+++ b/frontendActions/src/main/resources/actions/ResourceBundle.properties
@@ -1,0 +1,33 @@
+
+# Context menu
+
+action.GitMachete.CheckOutBranchAction.text=Git Machete: Check Out Selected Branch
+action.GitMachete.CheckOutBranchAction.GitMacheteContextMenu.text=_Check Out Branch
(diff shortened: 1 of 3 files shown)


frontendActions/src/main/resources/actions/ResourceBundle.properties
```diff
@@ -0,0 +1,33 @@
+
+# Context menu
```
PawelLipski, line 2: This is more wieldy since it can be easily accessed from Java code (as opposed to when it was dug deep in plugin.xml) + it can be i18ned (which I don't think we'll ever do anyway xD)

mkondratek: Y not? :poland: :rofl: 
> Rebazuj obecną gałąź na nadrzędną

PawelLipski: awwww :star_struck:

PR #251, opened 2020-05-03
Bugfix/heisenbug

mkondratek:


frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/GitMacheteGraphTableManager.java
```diff
@@ -130,7 +132,11 @@ private void refreshGraphTable(Path macheteFilePath, boolean isMacheteFilePresen
         LOG.info("Machete file (${macheteFilePath}) is empty");
       }
     }
-    graphTable.getModel().setRepositoryGraph(repositoryGraph);
```
PawelLipski, line 133: Hmmm okay... so essentially, `refreshGraphTable`, despite being marked as `@UIEffect`, has in fact been NOT invoked on the UI thread, right? Shouldn't `@UIEffect` annotation be then removed from this method (and probably transitively from its callers as well)?

PawelLipski: Also, where was `refreshGraphTable` called from when outside of UI thread? Was it from `new Task.Backgroundable -> run()` in `GitMacheteGraphTableManager#updateAndRefreshGraphTableInBackground`? I guess `Task.Backgroundable`s aren't executed on UI thread, right?

PawelLipski: Now I see that `GuiEffectChecker`, despite having a type hierarchy for **prohibiting execution of UI-effecting stuff outside of UI thread**, definitely lacks stuff for **prohibiting execution of non-UI-safe stuff on UI thread**. If there was a way to mark `com.intellij.openapi.progress.Task.Backgroundable#run` (or rather `com.intellij.openapi.progress.Progressive#run`) as non-UI-safe, the problem could not occur in the first place :/

PawelLipski: Actually, this is the same issue as https://github.com/typetools/checker-framework/issues/3253, but approached from a different direction: my issue says that we shouldn't allow certain methods to be executed on UI thread coz they might stall the UI; the problem here is that we shouldn't unsoundly assume that certain methods (like `Backgroundable#run` might be executed on UI thread while in fact they can't.

PawelLipski: I'm gonna raise that as separate, more serious issue (bug, not enhancent) against Checker

mkondratek: You are right and you are right. Annotation can be removed now and actually there is on base to assume that backgroundable tasks are executed in UI thread (they are not). 

As a curiosity, I can tell that there is `org.jetbrains.annotations.CalledInAwt` annotation. It is kind of our `@UiEffect`.

frontendUiRootImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/root/VcsRootComboBox.java
```diff
@@ -39,7 +39,7 @@ public VcsRootComboBox(@MinLen(1) List<GitRepository> repositories) {
 
   @Override
   @UIEffect
-  public void updateRepositories(@MinLen(1) List<GitRepository> repositories) {
+  public synchronized void updateRepositories(@MinLen(1) List<GitRepository> repositories) {
```
mkondratek, line 42: Changes to this file are related neither to #190 nor to #252. Those are to avoid another error in a case when `getSelectedRepository` is called during `updateRepositories` (exactly when previous repos was removed but new ones are not added yet)... :exploding_head:

frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/GitMacheteGraphTableManager.java
PawelLipski: Actually, I discovered a possibly serious bug in (GuiEffect? maybe not only)Checker: Checker shouldn't have allowed us to override a `@SafeEffect` method (`run`) of `Progressive` as a `@UIEffect`... raising an issue with them now...

PawelLipski: https://github.com/typetools/checker-framework/issues/3287

frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/GitMacheteGraphTableManager.java
```diff
@@ -208,16 +217,10 @@ private void updateRepository(Path mainDirectoryPath, Path gitDirectoryPath, boo
...
-          graphTable.setMacheteFilePath(macheteFilePath);
-          gitMacheteRepositoryRef.set(gitMacheteRepositoryFactory.create(mainDirectoryPath, gitDirectoryPath, branchLayout));
-        } else {
-          LOG.warn("Skipping the repository update because no VCS repository is selected");
-        }
+        IBranchLayout branchLayout = createBranchLayout(macheteFilePath);
+        graphTable.setBranchLayout(branchLayout);
+        graphTable.setMacheteFilePath(macheteFilePath);
+        gitMacheteRepositoryRef.set(gitMacheteRepositoryFactory.create(mainDirectoryPath, gitDirectoryPath, branchLayout));
       } catch (Exception e) {
```
PawelLipski, line 224: Unrel: specify the exception type + remove warning suppression from the method

frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/GitMacheteGraphTableManager.java
```diff
@@ -92,55 +94,59 @@ private void subscribeToGitRepositoryChanges() {
...
-  @UIEffect
   private void refreshGraphTable(Path macheteFilePath, boolean isMacheteFilePresent) {
-    LOG.debug(() -> "Entering: macheteFilePath = ${macheteFilePath}, isMacheteFilePresent = ${isMacheteFilePresent}");
-    // isUnitTestMode() checks if IDEA is running as a command line applet or in unit test mode.
-    // No UI should be shown when IDEA is running in this mode.
-    if (!project.isInitialized() || ApplicationManager.getApplication().isUnitTestMode()) {
-      LOG.debug("Project is not initialized or application is in unit test mode. Returning.");
-      return;
-    }
+    SwingUtilities.invokeLater(() -> {
```
micpiotrowski, line 111: Isn't it enough to invoke on UI thread only lines 137-138 + 147-148 ?

PawelLipski: A similar approach has been taken on #257

PawelLipski:
Consider incorporating the changes from #257 here and then we can continue the review of this PR.

frontendUiRootImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/root/VcsRootComboBox.java
PawelLipski: Undone on #257 :/ Rule of the thumb:
* Locks (and many other concurrency primitives) suck :banana: 
* Futures (or more generally: schedulable tasks) rock :rocket:

frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/GitMacheteGraphTableManager.java
PawelLipski: I fixed all problems with `@UIEffect` recognized so far on #257... and I yearn for https://github.com/typetools/checker-framework/issues/3253 even more now

frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/GitMacheteGraphTableManager.java
PawelLipski: Done on a recent @micpiotrowski 's PR anyway

PR #254, opened 2020-05-04
Subscribe VcsComboBox to VCS mapping changes

mkondratek:


frontendUiRootImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/root/VcsRootComboBox.java
```diff
@@ -36,7 +55,13 @@ public VcsRootComboBox(@MinLen(1) List<GitRepository> repositories) {
 
   @Override
   @UIEffect
-  public void updateRepositories(@MinLen(1) List<GitRepository> repositories) {
+  public void updateRepositories() {
+    // GitUtil.getRepositories(project) should never return empty list because it means there's no git repository in an opened
```
PawelLipski, line 59: Is this comment actually true in the light of #255?

frontendUiRootImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/root/VcsRootComboBox.java
```diff
@@ -36,7 +55,13 @@ public VcsRootComboBox(@MinLen(1) List<GitRepository> repositories) {
...
+    // project, so Git Machete plugin shouldn't even be loaded in the first place (as ensured by GitMacheteVisibilityPredicate)
+    List<GitRepository> repositories = List.ofAll(GitUtil.getRepositories(project));
```
PawelLipski, line 61: I'd put a comment here similar to what's in `com.virtuslab.gitmachete.frontend.ui.impl.table.GitMacheteGraphTableManager#queueGraphTableRefreshOnDispatchThread`: you're accessing git repo list on UI thread, but that's likely still okay

frontendUiRootImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/root/VcsRootComboBox.java
```diff
@@ -45,14 +70,20 @@ public void updateRepositories(@MinLen(1) List<GitRepository> repositories) {
...
     }
 
+    this.setVisible(getModel().getItems().size() > 1);
+
     boolean selectedItemUpdateRequired = selected == null || !getModel().getItems().contains(selected);
-    if (selectedItemUpdateRequired) {
-      getModel().setSelectedItem(repositories.get(0));
+    if (repositories.isEmpty()) {
+      LOG.debug(() -> "No VCS roots found");
+      observers.forEach(o -> o.onSelectionChanged());
```
PawelLipski, line 78: Hmm why notify the observers here? So that they can learn that the selected repository is now null, right?

frontendUiRootImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/root/VcsRootComboBox.java
```diff
@@ -45,14 +70,20 @@ public void updateRepositories(@MinLen(1) List<GitRepository> repositories) {
...
+    } else if (selectedItemUpdateRequired) {
+      LOG.debug(() -> "Selecting first VCS root");
+      setSelectedItem(repositories.get(0));
     } else {
+      LOG.debug(() -> "Selecting previously selected VCS root");
+      // VcsRootComboBox#setSelectedItem is omitted to avoid unnecessary observers call
       getModel().setSelectedItem(selected);
```
PawelLipski, line 85: Hmm why `setSelectedItem` in some cases and `getModel().setSelectedItem` in others?

frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/GitMacheteGraphTableManager.java
```diff
@@ -181,17 +179,9 @@ public void queueRepositoryUpdateAndGraphTableRefresh() {
     LOG.debug("Entering");
 
     if (project != null && !project.isDisposed()) {
-      // GitUtil.getRepositories(project) should never return empty list because it means there is no git repository
-      // in an opened project, so Git Machete plugin shouldn't even be loaded in the first place
-      @SuppressWarnings("value:assignment.type.incompatible")
-      @MinLen(1)
-      // A bit of a shortcut: we're accessing git repositories even though we may be on UI thread here;
```
PawelLipski, line 188: Yeah, this comment should be moved to the new location alongside with the `GitUtil.getRepositories` invocation

frontendUiRootImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/root/VcsRootComboBox.java
mkondratek: removed and todo left

frontendUiRootImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/root/VcsRootComboBox.java
mkondratek: > _Slack discussion summary._
> 
> This seems to be dangerous call since `updateRepositories` perform calls to `setSeletedItem` (in some cases) and `setSelectedItem` calls observers of `VcsRootsComboBox` so `updateRepositories` in particular.
> 
> Even though the stack overflow is not possible now, this situation is **bugo-genic**.

https://github.com/VirtusLab/git-machete-intellij-plugin/pull/260/files#r420868858

mkondratek: Please rereview now. `updateRepositories` has been removed from `queueRepositoryUpdateAndGraphTableRefresh` as combobox updates are now handled by subscription to `VCS_REPOSITORY_MAPPING_UPDATED` topic.

Also making `updateRepositories` private probably resolves the above concerns.

frontendUiRootImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/root/VcsRootComboBox.java
mkondratek: The difference is following:
- `getModel().setSelectedItem(...)` does not call observers 
- `setSelectedItem(...)` calls observers

Omitting observers is desirable in the case when selection does not change.

frontendUiRootImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/root/VcsRootComboBox.java
```diff
@@ -7,36 +7,61 @@
...
   @Override
   public MutableCollectionComboBoxModel<GitRepository> getModel() {
     return (MutableCollectionComboBoxModel<GitRepository>) super.getModel();
   }
 
-  @Override
   @UIEffect
-  public void updateRepositories(@MinLen(1) List<GitRepository> repositories) {
+  private void updateRepositories() {
+    // A bit of a shortcut: we're accessing filesystem even though we may be on UI thread here;
```
PawelLipski, line 59: `s/may be/are`

PawelLipski: If method is `@UIEffect`, then we can assume that it's invoked on UI thread *only*.
If method is `@SafeEffect` (i.e. non-UI-thread), then it may be invoked outside UI thread, but also on UI thread.

PR #272, opened 2020-05-11
[frontendActions] Add machete file edit action

mkondratek:
In scope of #37: file type added, branch name completion
In scope of #277: as the file type is added the plugin shall be suggested 

___
- added open machete file action 
- add (create and) open machete file action to empty graph table text
- add machete file completion 
- add machete file type with an icon

PR #245, opened 2020-04-30
Feat/fetch and pull

mkondratek:
- add `fetch all remotes`
- add `pull selected / current`
- extract common parts to action base classes and introduce `IActionSourceAwareBranchNameProvider`

additionally:
- rewrite `git4idea.fetch.GitFetchSupportImpl`
this is because IntelliJ below version 201 does not support specified branch update

PR #273, opened 2020-05-11
Git reset --keep action

micpiotrowski:


frontendActions/src/main/resources/actions/ResourceBundle.properties
```diff
@@ -17,6 +17,10 @@ action.GitMachete.PushSelectedBranchAction.text=Push Selected Branch...
 action.GitMachete.PushSelectedBranchAction.GitMacheteContextMenu.text=_Push Branch...
 action.GitMachete.PushSelectedBranchAction.description=Push this branch using push dialog
 
+action.GitMachete.ResetSelectedBranchAction.text=Reset Selected Branch
```
PawelLipski, line 20: As discussed on Slack, `Checkout and...`, same for `Rebase selected branch`

src/main/java/icons/MacheteIcons.java
```diff
@@ -0,0 +1,9 @@
+package icons;
+
+import com.intellij.openapi.util.IconLoader;
+import javax.swing.Icon;
+
+@SuppressWarnings("regexp")
```
PawelLipski, line 6: :hushed: what's RegexpChecker screaming here?

src/main/resources/icons/reset.svg
```diff
@@ -0,0 +1,36 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
PawelLipski, line 1: Nice, where's that taken from?

src/main/java/icons/MacheteIcons.java
```diff
@@ -0,0 +1,9 @@
...
+public interface MacheteIcons {
```
PawelLipski, line 7: final class?

frontendActions/src/main/resources/actions/ResourceBundle.properties
PawelLipski, line 20: Also, I'd specify that it's about `reset to remote (tracking branch)` in every single user-facing message... just `reset branch` with no further context is very confusing

src/main/resources/META-INF/plugin.xml
```diff
@@ -93,6 +99,12 @@
                 <!--                <override-text place="GitMacheteToolbar"/>-->
             </action>
 
+            <action id="GitMachete.ResetCurrentBranchAction"
```
PawelLipski, line 102: I'd also rename the action classes (including the base) to make it clear that the branches are reset to their remote [tracking branch].

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/toolbar/ResetCurrentBranchAction.java
```diff
@@ -0,0 +1,34 @@
...
+ *  <li>{@link DataKeys#KEY_GIT_MACHETE_REPOSITORY}</li>
+ *  <li>{@link DataKeys#KEY_SELECTED_VCS_REPOSITORY}</li>
+ *  <li>{@link CommonDataKeys#PROJECT}</li>
+ * </ul>
+ */
+public class ResetCurrentBranchAction extends BaseResetBranchAction {
+  @Override
+  protected Option<String> getBranchName(AnActionEvent anActionEvent) {
+    return getGitMacheteRepository(anActionEvent)
+        .flatMap(IGitMacheteRepository::getCurrentBranchIfManaged)
```
PawelLipski, line 26: There should be a helper in `ActionUtils` to get current branch (name)

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/common/BaseResetBranchAction.java
```diff
@@ -0,0 +1,169 @@
...
+
+    Option<String> branchName = getBranchName(anActionEvent);
+
+    if (branchName.isEmpty()) {
+      presentation.setEnabled(false);
+      presentation.setDescription("Reset disabled due to undefined ${getActionType()} branch");
+      return;
+    }
+
+    Option<SyncToRemoteStatus> syncToRemoteStatus = getGitMacheteRepository(anActionEvent)
```
PawelLipski, line 70: Use helper for branch name

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/common/BaseResetBranchAction.java
```diff
@@ -0,0 +1,169 @@
...
+      LOG.warn("Skipping the action because name of branch to reset is undefined");
+      VcsNotifier.getInstance(project).notifyError(VCS_LOGGER_TITLE,
+          "Internal error occurred. For more information see IDE log file");
+    }
+  }
+
+  protected void doResetToRemoteWithKeep(Project project, GitRepository gitRepository, String branchName,
+      IGitMacheteRepository macheteRepository) {
+
+    new Task.Backgroundable(project, TASK_TITLE, true) {
```
PawelLipski, line 124: Name `true`

src/main/java/icons/MacheteIcons.java
micpiotrowski: It's screaming about interface name (should starts with `I`) but as you suggested I just transformed it to final class

src/main/resources/icons/reset.svg
micpiotrowski: It's a combination of two icons from IntelliJ  :)

src/main/java/icons/MacheteIcons.java
micpiotrowski: done

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/toolbar/ResetCurrentBranchAction.java
micpiotrowski: Oh, it seems to was there all the time  :)
Replaced.

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/common/BaseResetBranchAction.java
micpiotrowski: Can you explain what do you mean?
I have branch name in var, so I just use it.

PawelLipski: Ok, leave as is, I confused two things here

src/main/resources/icons/reset.svg
PawelLipski: haha okay :rofl:

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/common/BaseResetBranchToRemoteAction.java
```diff
@@ -157,11 +178,9 @@ public void run(ProgressIndicator indicator) {
           // If `changes` is null the whole root will be refreshed
```
PawelLipski, line 178: Do we need to explicitly trigger `GitMacheteRepository` refresh here, or will it always be triggered automatically anyway?

frontendActions/src/main/resources/actions/ResourceBundle.properties
```diff
@@ -5,9 +5,9 @@ action.GitMachete.CheckoutSelectedBranchAction.text=Checkout Selected Branch
 action.GitMachete.CheckoutSelectedBranchAction.GitMacheteContextMenu.text=_Checkout Branch
 action.GitMachete.CheckoutSelectedBranchAction.description=Checkout this branch
 
-action.GitMachete.RebaseSelectedBranchOntoParentAction.text=Rebase Selected Branch Onto Parent
-action.GitMachete.RebaseSelectedBranchOntoParentAction.GitMacheteContextMenu.text=_Rebase Branch Onto Parent
-action.GitMachete.RebaseSelectedBranchOntoParentAction.description=Rebase this branch onto parent
+action.GitMachete.RebaseSelectedBranchOntoParentAction.text=Checkout and Rebase Selected Branch Onto Parent
```
PawelLipski, line 8: Nit: pls check how other IntelliJ actions choose b/w upper/lower case in words like `and`/`to` in title-cased captions like this one

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/common/BaseResetBranchToRemoteAction.java
```diff
@@ -157,11 +178,9 @@ public void run(ProgressIndicator indicator) {
...
           GitUtil.refreshVfs(gitRepository.getRoot(), /* changes */ null);
         }
-      }
 
-      @Override
-      public void onSuccess() {
```
PawelLipski: No need to override `onSuccess` anymore?

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/common/BaseResetBranchToRemoteAction.java
```diff
@@ -119,36 +126,50 @@ public void actionPerformed(AnActionEvent anActionEvent) {
...
+            new GitBranchWorker(project, Git.getInstance(), uiHandler)
+                .checkout(branchName, /* detach */ false, List.of(gitRepository));
+
+            // Check again if we are in branch to reset to be sure that checkout was successful
+            // This time we are using git4idea because GitMacheteRepository is immutable and it would return previous branch
+            GitLocalBranch nullableLocalBranch = gitRepository.getCurrentBranch();
+            if (nullableLocalBranch == null || !nullableLocalBranch.getName().equals(branchName)) {
+              LOG.error("Checkout to branch ${branchName} before reset was unsuccessful");
+              VcsNotifier.getInstance(project).notifyError(VCS_LOGGER_TITLE,
+                  "Error occurred during checking out to branch that was to be reset! Operation aborted.");
```
PawelLipski, line 167: Nit:
```suggestion
                  "Error occurred during checkout of the branch that was to be reset! Operation aborted.");
```

`-ing -ing` is generally considered to sound weird/poorly in English AFAIK

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/toolbar/ResetCurrentBranchToRemoteAction.java
```diff
@@ -0,0 +1,30 @@
...
+ * Expects DataKeys:
+ * <ul>
+ *  <li>{@link DataKeys#KEY_GIT_MACHETE_REPOSITORY}</li>
+ *  <li>{@link DataKeys#KEY_SELECTED_VCS_REPOSITORY}</li>
+ *  <li>{@link CommonDataKeys#PROJECT}</li>
+ * </ul>
+ */
+public class ResetCurrentBranchToRemoteAction extends BaseResetBranchToRemoteAction {
+  @Override
+  protected Option<String> getBranchName(AnActionEvent anActionEvent) {
```
mkondratek, line 22: I added `ActionAwareBranchNameProvider` in #245. It would be wise to use it here too :slightly_smiling_face:

src/main/resources/icons/reset.svg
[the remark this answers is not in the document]
mkondratek: Looks like a lightning riding an arrow :rofl:

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/common/BaseResetBranchToRemoteAction.java
mkondratek: Good question. We are subscribed to the vcs changes somewhere. Does it work in this case?

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/common/BaseResetBranchToRemoteAction.java
```diff
@@ -0,0 +1,188 @@
...
+            // Checking out given branch
+            GitBranchUiHandlerImpl uiHandler = new GitBranchUiHandlerImpl(project, Git.getInstance(), indicator);
+            new GitBranchWorker(project, Git.getInstance(), uiHandler)
+                .checkout(branchName, /* detach */ false, List.of(gitRepository));
+
+            // Check again if we are in branch to reset to be sure that checkout was successful
+            // This time we are using git4idea because GitMacheteRepository is immutable and it would return previous branch
+            GitLocalBranch nullableLocalBranch = gitRepository.getCurrentBranch();
+            if (nullableLocalBranch == null || !nullableLocalBranch.getName().equals(branchName)) {
+              LOG.error("Checkout to branch ${branchName} before reset was unsuccessful");
```
mkondratek, line 165: This sentence sounds weird to me. Since this log is preceded by 
`LOG.debug(() -> "Resetting '${branchName}' branch");`
what about logging the fact that the checkout is needed (right in the place where we know that, so the line 156 of this file) and then, here we could just: 
`LOG.error("Checkout to branch ${branchName} failed");`

Alternatively, the following sounds more ~natural~ better to me:
`LOG.error("Reset-preceding checkout of branch ${branchName} failed");`

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/common/BaseResetBranchToRemoteAction.java
```diff
@@ -0,0 +1,188 @@
...
+
+          resetHandler.endOptions();
+
+          // Check if branch to reset is not current branch - if isn't then checkout
+          Option<String> currentBranchOption = getCurrentBranchNameIfManaged(anActionEvent);
+          if (currentBranchOption.isEmpty() || !currentBranchOption.get().equals(branchName)) {
+            // Checking out given branch
+            GitBranchUiHandlerImpl uiHandler = new GitBranchUiHandlerImpl(project, Git.getInstance(), indicator);
+            new GitBranchWorker(project, Git.getInstance(), uiHandler)
+                .checkout(branchName, /* detach */ false, List.of(gitRepository));
```
mkondratek, line 159: We can delegate this directly to `CheckoutSelectedBranchAction`, can't we?

frontendActions/src/main/resources/actions/ResourceBundle.properties
```diff
@@ -5,14 +5,22 @@ action.GitMachete.CheckoutSelectedBranchAction.text=Checkout Selected Branch
...
+action.GitMachete.RebaseSelectedBranchOntoParentAction.GitMacheteContextMenu.text=Checkout and _Rebase Branch Onto Parent
```
mkondratek, line 9: :tada:  **Fun fact**  :tada: 
about this `_` within action texts like `Checkout and _Rebase Branch Onto `
(that probably only me wasn't aware of)
___

It is better to see the context - the context menu. 

image

`_` underlines the following letter. It indicates the key shortcut to perform an action within the context menu. 

So in the example presented above:
- pressing **R** - performs `Rollback`
- pressing **D** - performs `Delete`

We should be aware of that to **avoid duplicate definitions**.

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/common/BaseResetBranchToRemoteAction.java
```diff
@@ -0,0 +1,188 @@
...
+  @UIEffect
+  public void update(AnActionEvent anActionEvent) {
+    super.update(anActionEvent);
+
+    Presentation presentation = anActionEvent.getPresentation();
+    if (!presentation.isEnabledAndVisible()) {
+      return;
+    }
+
+    Option<String> branchNameOption = getBranchName(anActionEvent);
```
mkondratek, line 66: nit: `var branchName` here for consistency

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/common/BaseResetBranchToRemoteAction.java
```diff
@@ -0,0 +1,188 @@
...
+    }
+  }
+
+  @Override
+  @UIEffect
+  public void actionPerformed(AnActionEvent anActionEvent) {
+    LOG.debug("Performing");
+
+    Project project = getProject(anActionEvent);
+    Option<GitRepository> selectedVcsRepository = getSelectedVcsRepository(anActionEvent);
```
mkondratek, line 102: I didn't expect that :wink:

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/common/BaseResetBranchToRemoteAction.java
```diff
@@ -0,0 +1,188 @@
...
+ *  <li>{@link DataKeys#KEY_GIT_MACHETE_REPOSITORY}</li>
+ *  <li>{@link CommonDataKeys#PROJECT}</li>
+ * </ul>
+ */
+public abstract class BaseResetBranchToRemoteAction extends GitMacheteRepositoryReadyAction {
+  private static final IPrefixedLambdaLogger LOG = PrefixedLambdaLoggerFactory.getLogger("frontendActions");
+  private static final String VCS_LOGGER_TITLE = "Resetting";
+  private static final String TASK_TITLE = "Resetting...";
+
+  protected abstract Option<String> getBranchName(AnActionEvent anActionEvent);
```
mkondratek, line 52: hmmmm :thinking:  actually an implementation of this method can use any data key. Let's lists the used one at least

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/common/BaseResetBranchToRemoteAction.java
```diff
@@ -0,0 +1,188 @@
...
+              anActionEvent);
+        } else {
+          LOG.error("Skipping the action because can't get Git Machete repository");
+          VcsNotifier.getInstance(project).notifyError(VCS_LOGGER_TITLE,
+              "Internal error occurred. For more information see IDE log file");
+        }
+      } else {
+        LOG.warn("Skipping the action because no VCS repository is selected");
+        VcsNotifier.getInstance(project).notifyWarning(VCS_LOGGER_TITLE,
+            "Skipping the action because no VCS repository is selected");
```
mkondratek, line 119: This is an inconsistency to the messages above and below. 
And I am not convinced that `Internal error occurred. For more information see IDE log file` is something that we wanna show. I think that an information with a cause like `because name of branch to reset is undefined` would be more helpfull to the user. 

Feel free to open an issue to clean up and provide some conventions to vcs notification and logs within the scope of `frontendActions`.

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/common/BaseResetBranchToRemoteAction.java
```diff
@@ -0,0 +1,188 @@
...
+            String message = "Branch '${branchName}' doesn't have remote tracking branch, so cannot be reset";
+            LOG.warn(message);
+            VcsNotifier.getInstance(project).notifyWarning(VCS_LOGGER_TITLE, message);
+          }
+
+          resetHandler.endOptions();
+
+          // Check if branch to reset is not current branch - if isn't then checkout
+          Option<String> currentBranchOption = getCurrentBranchNameIfManaged(anActionEvent);
+          if (currentBranchOption.isEmpty() || !currentBranchOption.get().equals(branchName)) {
```
mkondratek, line 155: Shouldn't it be the first thing we do in this scope?

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/common/BaseResetBranchToRemoteAction.java
mkondratek: please ignore

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/common/BaseResetBranchToRemoteAction.java
micpiotrowski: Yeah, it's seems to be unneeded.

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/common/BaseResetBranchToRemoteAction.java
micpiotrowski: No, it's no needed coz it's run after each `run` method invocation regardless of checkout and reset operations success, so it's not what we want.

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/common/BaseResetBranchToRemoteAction.java
micpiotrowski: Ok, after some refactor in `CheckoutSelectedBranchAction` it is done.

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/common/BaseResetBranchToRemoteAction.java
micpiotrowski: It can be but IMHO making sure that we can reset branch (it has remote tracking branch) first is better coz we don't force user to checkout unnecessarily in this case.

frontendActions/src/main/resources/actions/ResourceBundle.properties
micpiotrowski: It seems to be the same: words like `and`, `to`, `from`, `the` etc. are in lowercase

src/main/resources/icons/reset.svg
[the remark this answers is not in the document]
micpiotrowski: It'a a combination of left round arrow that is a symbol of undoing something and lighting that indicate something potential dangerous (despite of `--keep` option it is a branch resetting).
IMHO only arrow is not enough intuitive for that action.
If you have a better idea, please tell me. I can try to do that icon.

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/common/BaseResetBranchToRemoteAction.java
micpiotrowski: Created #320

frontendActions/src/main/resources/actions/ResourceBundle.properties
```diff
@@ -48,6 +52,10 @@ action.GitMachete.PushCurrentBranchAction.text=Push Current Branch...
 action.GitMachete.PushCurrentBranchAction.GitMacheteToolbar.text=Push Current Branch...
 action.GitMachete.PushCurrentBranchAction.description=Push current branch using push dialog
 
+action.GitMachete.ResetCurrentBranchAction.text=Reset Current Branch to Remote
+action.GitMachete.ResetCurrentBranchAction.GitMacheteContextMenu.text=_Reset Current Branch to Remote
```
PawelLipski, line 56: Pls keep all `GitMacheteContextMenu` entries together

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/common/BaseResetBranchToRemoteAction.java
```diff
@@ -0,0 +1,179 @@
...
+          }
+
+          GitCommandResult result = Git.getInstance().runCommand(resetHandler);
+          if (!result.success()) {
+            LOG.error(result.getErrorOutputAsJoinedString());
+            VcsNotifier.getInstance(project).notifyError(VCS_NOTIFIER_TITLE, result.getErrorOutputAsHtmlString());
+          }
+        }
+
+        // If we are here this mean that all went good
```
PawelLipski, line 173: Hmm how about the case where `!result.success()`, lines 168-169? We're also gonna get here, right?

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/toolbar/ResetCurrentBranchToRemoteAction.java
```diff
@@ -0,0 +1,30 @@
...
+import io.vavr.control.Option;
+
+import com.virtuslab.gitmachete.frontend.actions.common.BaseResetBranchToRemoteAction;
+import com.virtuslab.gitmachete.frontend.datakeys.DataKeys;
+
+/**
+ * Expects DataKeys:
+ * <ul>
+ *  <li>{@link DataKeys#KEY_GIT_MACHETE_REPOSITORY}</li>
+ *  <li>{@link DataKeys#KEY_SELECTED_VCS_REPOSITORY}</li>
```
PawelLipski, line 16: Do we actually need to provide `KEY_SELECTED_VCS_REPOSITORY` here?

PawelLipski: I'm btw gonna rework these `DataKey` specs, that's getting unmaintainable (move it to interfaces instead or whatnot)

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/common/BaseResetBranchToRemoteAction.java
micpiotrowski: o-o, yeah, my fault.

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/common/BaseResetBranchToRemoteAction.java
```diff
@@ -0,0 +1,179 @@
...
+          }
+
+          resetHandler.endOptions();
+
+          // Check if branch to reset is not current branch - if isn't then checkout
+          Option<String> currentBranchOption = getCurrentBranchNameIfManaged(anActionEvent);
+          if (currentBranchOption.isEmpty() || !currentBranchOption.get().equals(branchName)) {
+            LOG.debug(() -> "Checkout to branch '${branchName}' is needed");
+            // Checking out given branch
+            CheckoutSelectedBranchAction.doCheckout(branchName, gitRepository, project, indicator);
```
PawelLipski, line 151: Hmmm dunno, maybe just invoke `actionPerformed` on `CheckoutSelectedBranchAction` so as to not expose the inner workings (`doCheckout`) of `CheckoutSelectedBranchAction`?

PawelLipski: Ouch, `branchName`... we'd need to pass that via `AnActionEvent` then, that's too bad :/

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/toolbar/ResetCurrentBranchToRemoteAction.java
micpiotrowski: Yes, it is used by `ActionUtils.getSelectedVcsRepository()` invocation in `Base` class

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/common/BaseResetBranchToRemoteAction.java
```diff
@@ -0,0 +1,179 @@
...
+
+            // Check again if we are in branch to reset to be sure that checkout was successful
+            // This time we are using git4idea because GitMacheteRepository is immutable and it would return previous branch
+            GitLocalBranch nullableLocalBranch = gitRepository.getCurrentBranch();
```
PawelLipski, line 155: Can be `@Nullable GitLocalBranch localBranch` as well to highlight that the value may be null && avoid Hungarian notation

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/common/BaseResetBranchToRemoteAction.java
micpiotrowski: It's not good idea, because it run its action in backgroundable task, so do this async. but we need to do this synchronously.

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/common/BaseResetBranchToRemoteAction.java
```diff
@@ -0,0 +1,179 @@
...
+
+          Option<BaseGitMacheteBranch> branchOption = macheteRepository.getBranchByName(branchName);
+          assert branchOption.isDefined() : "Can't get branch '${branchName}' from Git Machete repository";
+          Option<IGitMacheteRemoteBranch> remoteTrackingBranchOption = branchOption.get().getRemoteTrackingBranch();
+          if (remoteTrackingBranchOption.isDefined()) {
+            resetHandler.addParameters(remoteTrackingBranchOption.get().getPointedCommit().getHash());
+          } else {
+            String message = "Branch '${branchName}' doesn't have remote tracking branch, so cannot be reset";
+            LOG.warn(message);
+            VcsNotifier.getInstance(project).notifyWarning(VCS_NOTIFIER_TITLE, message);
```
PawelLipski, line 141: Does it throw an error? Shouldn't we have an explicit `return` here?

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/common/BaseResetBranchToRemoteAction.java
```diff
@@ -0,0 +1,179 @@
...
+
+  protected void doResetToRemoteWithKeep(Project project, GitRepository gitRepository, String branchName,
+      IGitMacheteRepository macheteRepository, AnActionEvent anActionEvent) {
+
+    new Task.Backgroundable(project, TASK_TITLE, /* canBeCancelled */ true) {
+
+      @Override
+      public void run(ProgressIndicator indicator) {
+        LOG.debug(() -> "Resetting '${branchName}' branch");
+        try (AccessToken ignored = DvcsUtil.workingTreeChangeStarted(project, TASK_TITLE)) {
```
PawelLipski, line 129: Drop a comment on why this is necessary

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/common/BaseResetBranchToRemoteAction.java
micpiotrowski: Yeas, again, my fault...

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/common/BaseResetBranchToRemoteAction.java
micpiotrowski: TBH, I'm not 100% but from JB code it seems to be try with resources construction that notify listeners about batch changes and probably it's some kind of `mutex` for repo operations (see: https://github.com/JetBrains/intellij-community/blob/master/platform/dvcs-impl/src/com/intellij/dvcs/DvcsUtil.java#L171)
This piece of code is also JB actions, so I think that it can be important.

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/common/BaseResetBranchToRemoteAction.java
```diff
@@ -0,0 +1,183 @@
...
+  @UIEffect
+  public void actionPerformed(AnActionEvent anActionEvent) {
+    LOG.debug("Performing");
+
+    Project project = getProject(anActionEvent);
+    var gitRepository = getSelectedVcsRepository(anActionEvent);
+    var branchName = getNameOfBranchUnderAction(anActionEvent);
+    var macheteRepository = getGitMacheteRepository(anActionEvent);
+
+    if (branchName.isDefined()) {
```
PawelLipski, line 101: Maybe flip the ifs `to `if (NOT ...) { LOG...; return; }` to reduce nesting & keep warnings close to their respective conditions

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/common/BaseResetBranchToRemoteAction.java
```diff
@@ -0,0 +1,183 @@
...
+
+          GitCommandResult result = Git.getInstance().runCommand(resetHandler);
+          if (!result.success()) {
+            LOG.error(result.getErrorOutputAsJoinedString());
+            VcsNotifier.getInstance(project).notifyError(VCS_NOTIFIER_TITLE, result.getErrorOutputAsHtmlString());
+            return;
+          }
+        }
+
+        // If we are here this mean that all went good
```
PawelLipski, line 177: `means`

src/main/resources/icons/reset.svg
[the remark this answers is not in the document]
mkondratek: Don't get me wrong. I am not against this. I just wanted to know the reasoning behind it. It is hard to provide "descriptive" icons for actions like reset, rebase or even merge --ff-only. This is fine! :rocket:

frontendActions/src/main/resources/actions/ResourceBundle.properties
```diff
@@ -5,18 +5,22 @@ action.GitMachete.CheckoutSelectedBranchAction.text=Checkout Selected Branch
...
-action.GitMachete.RebaseSelectedBranchOntoParentAction.text=Rebase Selected Branch Onto Parent
-action.GitMachete.RebaseSelectedBranchOntoParentAction.GitMacheteContextMenu.text=_Rebase Branch Onto Parent
-action.GitMachete.RebaseSelectedBranchOntoParentAction.description=Rebase this branch onto parent
+action.GitMachete.RebaseSelectedBranchOntoParentAction.text=Checkout and Rebase Selected Branch Onto Parent
+action.GitMachete.RebaseSelectedBranchOntoParentAction.GitMacheteContextMenu.text=Checkout and _Rebase Branch Onto Parent
+action.GitMachete.RebaseSelectedBranchOntoParentAction.description=Checkout and rebase this branch onto parent
 
 action.GitMachete.PullSelectedBranchAction.text=Pull Selected Branch
-action.GitMachete.PullSelectedBranchAction.GitMacheteContextMenu.text=_Pull Branch
+action.GitMachete.PullSelectedBranchAction.GitMacheteContextMenu.text=Pull Branch
```
mkondratek, line 13: Let's mark `u`(pdate) with `_` here. 
`P_ull Branch`

frontendActions/src/main/resources/actions/ResourceBundle.properties
```diff
@@ -5,18 +5,22 @@ action.GitMachete.CheckoutSelectedBranchAction.text=Checkout Selected Branch
...
 action.GitMachete.PullSelectedBranchAction.description=Pull this branch
 
 action.GitMachete.PushSelectedBranchAction.text=Push Selected Branch...
 action.GitMachete.PushSelectedBranchAction.GitMacheteContextMenu.text=_Push Branch...
 action.GitMachete.PushSelectedBranchAction.description=Push this branch using push dialog
 
+action.GitMachete.ResetSelectedBranchAction.text=Checkout and Reset Selected Branch to Remote
+action.GitMachete.ResetSelectedBranchAction.GitMacheteContextMenu.text=Checkout and Reset Branch to Remote
```
mkondratek, line 21: What about marking some shortcut key here? Maybe `s`(oft)?
`Checkout and Re_set Branch to Remote`

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/toolbar/ResetCurrentBranchToRemoteAction.java
```diff
@@ -0,0 +1,30 @@
...
+ * </ul>
+ */
+public class ResetCurrentBranchToRemoteAction extends BaseResetBranchToRemoteAction {
+  @Override
+  public Option<String> getNameOfBranchUnderAction(AnActionEvent anActionEvent) {
+    return getCurrentBranchNameIfManaged(anActionEvent);
+  }
+
+  @Override
+  protected String getActionType() {
```
mkondratek, line 27: not needed as we can retrieve action place from `anActionEvent` (see other `IBranchNameProvider` implementations)

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/contextmenu/ResetSelectedBranchToRemoteAction.java
```diff
@@ -0,0 +1,31 @@
...
+ * </ul>
+ */
+public class ResetSelectedBranchToRemoteAction extends BaseResetBranchToRemoteAction {
+  @Override
+  public Option<String> getNameOfBranchUnderAction(AnActionEvent anActionEvent) {
+    return getSelectedBranchName(anActionEvent);
+  }
+
+  @Override
+  protected String getActionType() {
```
mkondratek, line 28: we dont need that, see comment above (or below, idk, reviewing upwards XD)

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/contextmenu/CheckoutSelectedBranchAction.java
```diff
@@ -78,15 +80,20 @@ public void actionPerformed(AnActionEvent anActionEvent) {
       new Task.Backgroundable(project, "Checking out") {
         @Override
         public void run(ProgressIndicator indicator) {
-          LOG.info("Checking out branch '${selectedBranchName.get()}'");
-          GitBranchUiHandlerImpl uiHandler = new GitBranchUiHandlerImpl(project, Git.getInstance(), indicator);
-          new GitBranchWorker(project, Git.getInstance(), uiHandler)
-              .checkout(selectedBranchName.get(), /* detach */ false, java.util.List.of(gitRepository.get()));
+          doCheckout(selectedBranchName.get(), gitRepository.get(), project, indicator);
```
mkondratek, line 83: :+1:

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/common/BaseResetBranchToRemoteAction.java
```diff
@@ -0,0 +1,185 @@
...
+ *  <li>{@link DataKeys#KEY_SELECTED_VCS_REPOSITORY}</li>
+ *  <li>{@link CommonDataKeys#PROJECT}</li>
+ * </ul>
+ */
+public abstract class BaseResetBranchToRemoteAction extends GitMacheteRepositoryReadyAction implements IBranchNameProvider {
+  private static final IPrefixedLambdaLogger LOG = PrefixedLambdaLoggerFactory.getLogger("frontendActions");
+  private static final String VCS_NOTIFIER_TITLE = "Resetting";
+  private static final String TASK_TITLE = "Resetting...";
+
+  protected abstract String getActionType();
```
mkondratek, line 49: :-1:

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/toolbar/ResetCurrentBranchToRemoteAction.java
mkondratek: So it seems that comments are ordered by datetime.... :thinking:  cool !

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/common/BaseResetBranchToRemoteAction.java
PawelLipski: :question:

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/common/BaseResetBranchToRemoteAction.java
```diff
@@ -0,0 +1,185 @@
...
+          GitCommandResult result = Git.getInstance().runCommand(resetHandler);
+          if (!result.success()) {
+            LOG.error(result.getErrorOutputAsJoinedString());
+            VcsNotifier.getInstance(project).notifyError(VCS_NOTIFIER_TITLE, result.getErrorOutputAsHtmlString());
+            return;
+          }
+        }
+
+        // If we are here this means that all went good
+        VcsNotifier.getInstance(project).notifySuccess("Branch '${branchName}' reset to remote");
```
PawelLipski, line 180: Hmmm when testing locally I've noticed that GitMacheteGraphTable isn't refreshed (the `(diverged from origin)` doesn't disappear). Maybe call refresh action here then?

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/common/BaseResetBranchToRemoteAction.java
PawelLipski: Oh ok, can be retrieved from action event

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/toolbar/ResetCurrentBranchToRemoteAction.java
micpiotrowski: Ok, I give up this distinction completely.
Now I just show branch name instead of it's type (current/selected)

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/contextmenu/ResetSelectedBranchToRemoteAction.java
micpiotrowski: as above

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/common/BaseResetBranchToRemoteAction.java
micpiotrowski: Strange, it's worked for me...

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/toolbar/ResetCurrentBranchToRemoteAction.java
PawelLipski: You can leave the distinction but just base it on `com.intellij.openapi.actionSystem.AnActionEvent#getPlace` rather than a bespoke method of yours

src/main/java/icons/MacheteIcons.java
```diff
@@ -0,0 +1,9 @@
+package icons;
```
PawelLipski, line 1: TBD with @mkondratek whether files like this one should be left in the root project or rather go into `frontendIcons` (as on #272)

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/common/BaseResetBranchToRemoteAction.java
PawelLipski: Yeah, might be undeterministic :(

src/main/java/icons/MacheteIcons.java
mkondratek: I would suggest to move it now. As a separate module it is more useful.

mkondratek: see #272

micpiotrowski: Ok, I moved it to `frontendIcons` module, but according to JB SDK (https://www.jetbrains.org/intellij/sdk/docs/reference_guide/work_with_icons_and_images.html) this class must be in `icons` package to can be simply used in `plugin.xml`

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/common/BaseResetBranchToRemoteAction.java
micpiotrowski: Restore previous refreshing methods invocation

frontendActions/src/main/resources/actions/ResourceBundle.properties
```diff
@@ -5,18 +5,22 @@ action.GitMachete.CheckoutSelectedBranchAction.text=Checkout Selected Branch
...
-action.GitMachete.PullSelectedBranchAction.GitMacheteContextMenu.text=_Pull Branch
+action.GitMachete.PullSelectedBranchAction.GitMacheteContextMenu.text=P_ull Branch
 action.GitMachete.PullSelectedBranchAction.description=Pull this branch
 
 action.GitMachete.PushSelectedBranchAction.text=Push Selected Branch...
 action.GitMachete.PushSelectedBranchAction.GitMacheteContextMenu.text=_Push Branch...
 action.GitMachete.PushSelectedBranchAction.description=Push this branch using push dialog
 
+action.GitMachete.ResetSelectedBranchAction.text=Checkout and Reset Selected Branch to Remote
+action.GitMachete.ResetSelectedBranchAction.GitMacheteContextMenu.text=Checkout and Re_set Branch to Remote
```
PawelLipski, line 21: Hmm I guess this action should not have `Checkout and` part in case of the current branch... not sure if that's covered

frontendActions/src/main/resources/actions/ResourceBundle.properties
```diff
@@ -5,18 +5,22 @@ action.GitMachete.CheckoutSelectedBranchAction.text=Checkout Selected Branch
 action.GitMachete.CheckoutSelectedBranchAction.GitMacheteContextMenu.text=_Checkout Branch
 action.GitMachete.CheckoutSelectedBranchAction.description=Checkout this branch
 
-action.GitMachete.RebaseSelectedBranchOntoParentAction.text=Rebase Selected Branch Onto Parent
-action.GitMachete.RebaseSelectedBranchOntoParentAction.GitMacheteContextMenu.text=_Rebase Branch Onto Parent
-action.GitMachete.RebaseSelectedBranchOntoParentAction.description=Rebase this branch onto parent
+action.GitMachete.RebaseSelectedBranchOntoParentAction.text=Checkout and Rebase Selected Branch Onto Parent
+action.GitMachete.RebaseSelectedBranchOntoParentAction.GitMacheteContextMenu.text=Checkout and _Rebase Branch Onto Parent
```
PawelLipski, line 9: As below, `Checkout and` redundant in case of current branch (can be resolved within a separate issue)

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/common/BaseResetBranchToRemoteAction.java
```diff
@@ -0,0 +1,197 @@
...
+                  "Error occurred during checkout of the branch that was to be reset! Operation aborted.");
+              return;
+            } else {
+              LOG.debug(() -> "Checkout to branch '${branchName}' successful");
+            }
+          }
+
+          GitCommandResult result = Git.getInstance().runCommand(resetHandler);
+
+          // git4idea methods are used instead of our RefreshStatusAction, to refresh state not only in our plugin,
```
PawelLipski, line 173: :+1:

frontendActions/src/main/resources/actions/ResourceBundle.properties
micpiotrowski: Issue #327 created

src/main/resources/META-INF/plugin.xml
```diff
@@ -60,6 +60,12 @@
                 <!--                <override-text place="GitMacheteContextMenu"/>-->
             </action>
 
+            <action id="GitMachete.ResetSelectedBranchAction"
+                    class="com.virtuslab.gitmachete.frontend.actions.contextmenu.ResetSelectedBranchToRemoteAction"
+                    icon="MacheteIcons.RESET">
```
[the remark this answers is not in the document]
mkondratek: _Slack discussion summary_

let it be for now

PR #275, opened 2020-05-12
Extract CONTRIBUTING.md from README.md

PawelLipski:
README now contains all the things that might in my assessment be ever needed to someone who **builds the project locally** but **does not contribute**.

PR #280, opened 2020-05-13
Fix Gradle options for local/non-CI builds

PawelLipski:
As suggested by @mkondratek on #275

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -94,7 +94,8 @@ subprojects {
       // Treat each compiler warning (esp. the ones coming from Checker Framework) as an error.
       options.compilerArgs << '-Werror'
       if (System.getenv('CI') != 'true') {
-        options.forkOptions.jvmArgs += ['-Xmx12g', '-XX:+HeapDumpOnOutOfMemoryError']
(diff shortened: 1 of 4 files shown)


build.gradle
```diff
@@ -79,6 +81,9 @@ subprojects {
     tasks.withType(JavaCompile) {
       // Treat each compiler warning (esp. the ones coming from Checker Framework) as an error.
       options.compilerArgs << '-Werror'
+      if (System.getenv('CI') != 'true') {
+        options.forkOptions.jvmArgs += ['-Xmx12g', '-XX:+HeapDumpOnOutOfMemoryError']
```
PawelLipski, line 85: Skipping `-XX:MaxPermSize` - deprecated since Java 8

mkondratek: IMHO this (local options) should be mentioned in the contribution guidelines anyway

PawelLipski: @mkondratek I reworked that slightly (also decreased the default limit, 12GB was kinda a looot), pls see the updated version

mkondratek: this is fine :fire:

PR #285, opened 2020-05-15
Push tag back to the repository on master builds

PawelLipski:


build.gradle
```diff
@@ -32,19 +32,27 @@ if (System.getenv('CI') == 'true') {
...
-  def maybeDirty = git.status().clean ? '' : '-dirty'
-  git.close()
-
   group 'com.virtuslab'
+
   apply from: "$rootDir/version.gradle"
-  version "${PLUGIN_VERSION}${maybeSnapshot}${shortCommitHash}${maybeDirty}"
+  def ciBranch = System.getenv('CIRCLE_BRANCH')
+  if (ciBranch == 'master') {
+    version PLUGIN_VERSION
```
PawelLipski, line 41: No longer attach build metadata to released versions, we'll have tags to identify them instead

.circleci/config.yml
```diff
@@ -75,6 +75,13 @@ jobs:
     steps:
       - attach_workspace:
           at: .
+      - deploy:
+          name: Push git tag
+          command: |
+            source scripts/utils.sh
+            tag=v$(extract_version_from_current_wd)
+            git tag "$tag"
+            git push https://${GITHUB_TOKEN}@github.com/Virtuslab/git-machete-intellij-plugin.git "$tag"
```
PawelLipski, line 84: `GITHUB_TOKEN` is a var set up manually in Circle CI by myself; as any other reasonable CI, Circle always masks the matches of secret values in the output + doesn't allow using secrets in env vars coming from forks (or otherwise... see https://medium.com/virtuslab/nifty-docker-tricks-for-your-ci-vol-2-c5191a67f1a4 + Ctrl-F `Once the image is built`)

build.gradle
```diff
@@ -32,19 +32,27 @@ if (System.getenv('CI') == 'true') {
...
+  } else {
+    // As for now, we're unfortunately not running builds for develop yet
```
mkondratek, line 43: _Slack discussion summary_

Let's consider running CI builds for all branches.

PawelLipski: Done, reconfigured CI, updating the comment...

.circleci/config.yml
PawelLipski: (Actually, as for now we even never build PRs coming from forks)

image

PR #292, opened 2020-05-18
Add basic thread-safe time measurement to the logger

PawelLipski:
image

### DIFF
--- a/logging/src/main/java/com/virtuslab/logger/PrefixedLambdaLogger.java
+++ b/logging/src/main/java/com/virtuslab/logger/PrefixedLambdaLogger.java
@@ -5,10 +5,13 @@ import java.io.StringWriter;
 import java.util.function.Supplier;
 
 import kr.pe.kwonnam.slf4jlambda.LambdaLogger;
+import org.checkerframework.checker.nullness.qual.Nullable;
 
 public class PrefixedLambdaLogger implements IPrefixedLambdaLogger {
   private final LambdaLogger logger;
   private final String className;
+  private final ThreadLocal<@Nullable String> bufferedTimerMessage;
+  private final ThreadLocal<@Nullable Long> timerStartMillis;
 
   PrefixedLambdaLogger(LambdaLogger logger) {
     this.logger = logger;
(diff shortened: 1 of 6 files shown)


logging/src/main/java/com/virtuslab/logger/PrefixedLambdaLogger.java
```diff
@@ -5,10 +5,13 @@
 import java.util.function.Supplier;
 
 import kr.pe.kwonnam.slf4jlambda.LambdaLogger;
+import org.checkerframework.checker.nullness.qual.Nullable;
 
 public class PrefixedLambdaLogger implements IPrefixedLambdaLogger {
   private final LambdaLogger logger;
   private final String className;
+  private final ThreadLocal<@Nullable String> bufferedTimerMessage;
```
PawelLipski, line 13: Tip: `ThreadLocal<T>` is easiest to conceptualize as `ConcurrentHashMap<Long /* thread id */, T>`

logging/src/main/java/com/virtuslab/logger/PrefixedLambdaLogger.java
```diff
@@ -34,7 +39,9 @@ private String getLogMessagePrefix() {
     // #6: method from which IMacheteLogger.<logMethod> was called;
     @SuppressWarnings("upperbound")
     StackTraceElement element = Thread.currentThread().getStackTrace()[6];
-    return className + "#" + element.getMethodName() + ": ";
+    String timerMessage = bufferedTimerMessage.get();
```
PawelLipski, line 42: ... and then `ThreadLocal#get` would be just sth like `.get(Thread.currentThread().getId())`

PawelLipski: Similarly with `set` and `remove`

scripts/utils.sh
```diff
@@ -16,7 +16,15 @@ function die() {
 function extract_version_from_gradle_file_stdin() {
   [[ -t 0 ]] && die "${FUNCNAME[0]}: expecting non-terminal stdin, aborting" || true
 
-  grep -Po "(?<=PLUGIN_VERSION = ').*(?=')"
+  # Let's avoid any external process calls (cat/grep) to speed things up.
```
mkondratek, line 19: :heart_decoration:

PR #294, opened 2020-05-18
Partially automate and document the intended flow for backports

PawelLipski:


.circleci/config.yml
```diff
@@ -75,18 +75,32 @@ jobs:
...
             source scripts/utils.sh
-            tag=v$(extract_version_from_current_wd)
-            git tag "$tag"
-            git push https://${GITHUB_TOKEN}@github.com/Virtuslab/git-machete-intellij-plugin.git "$tag"
-      - deploy:
-          name: Publish plugin to Jetbrains Marketplace
-          command: ./gradlew publishPlugin
+            version=$(extract_version_from_current_wd)
+            git checkout -b "backport/$version"
+            ssh-keyscan -t rsa github.com | ssh-keygen -lf -
```
PawelLipski, line 94: So that the pipeline don't stall on the prompt whether we trust the key

.circleci/config.yml
```diff
@@ -75,18 +75,32 @@ jobs:
...
+            hub pull-request \
```
PawelLipski, line 95: Authorized with `$GITHUB_TOKEN` as well

scripts/enforce-pre-release-version-correct-use
```diff
@@ -11,16 +11,23 @@ if [[ ${CI_PULL_REQUEST-} ]]; then
...
-  if [[ ( $CIRCLE_BRANCH == master || $CIRCLE_BRANCH == release/* || $CIRCLE_BRANCH == hotfix/* ) && $head_pre_release ]]; then
-    die 'Pre-release version is not allowed on master/release/hotfix branch tip'
-  elif [[ $CIRCLE_BRANCH != master && -z $head_pre_release ]]; then
-    die 'Non-master/release/hotfix branch tip must have a pre-release version (a.b.c-d, not a.b.c)'
+  if [[ $CIRCLE_BRANCH == master || $CIRCLE_BRANCH == release/* || $CIRCLE_BRANCH == hotfix/* || $CIRCLE_BRANCH == backport/* ]]; then
+    [[ -z $head_pre_release ]] || \
+      die 'Pre-release version is not allowed on master/release/hotfix/backport branch tip'
+  else
+    [[ $head_pre_release ]] || \
+      die 'Non-master/release/hotfix/backport branch tip must have a pre-release version (a.b.c-d, not a.b.c)'
```
mkondratek, line 31: It's a bit unclear what is `non-ed` here :confused:

.circleci/config.yml
mkondratek: wow, this is cool :rocket:

scripts/enforce-pre-release-version-correct-use
PawelLipski: Done, fixed... but I'm lazy, fixed that on the very end of PR chain (#329) :trollface:

PR #298, opened 2020-05-20
Capture stray unprocessed string interpolations

PawelLipski:


binding/src/main/java/com/virtuslab/binding/RuntimeBinding.java
```diff
@@ -31,17 +31,17 @@ private RuntimeBinding() {}
           .filter(c -> !c.isInterface() && !c.isAnonymousClass() && !c.isLocalClass() && !c.isMemberClass())
           .collect(Collectors.toSet());
       if (classes.isEmpty()) {
-        throw new ClassNotFoundException("No viable class implementing ${interfaze.getCanonicalName()} found");
+        throw new ClassNotFoundException("No viable class implementing " + interfaze.getCanonicalName() + " found");
```
PawelLipski, line 34: Those interpolations weren't resolved... no idea why

mkondratek: :thinking:

scripts/verify-no-string-interpolation-leftovers
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+
+set -e -o pipefail -u
+
+exit_code=0
+for class in $(find . -name '*.class' -type f); do
+  strings_output=$(strings $class)
```
PawelLipski, line 7: `javap` (java "deassembler") would be much more correct here, `strings` is just a nasty heurestics; I tried the former but it was prohibitively slow, esp. given that there's surprisingly no option to just list constant pool; the entire constant pool is only listed in the most compute-intensive `-verbose` mode

PR #300, opened 2020-05-20
Fix range of commits displayed when merge commits are present

PawelLipski:
Wrt. the original issue:

image

### DIFF
--- a/config/checkstyle/tree_walker_module_directives.xml
+++ b/config/checkstyle/tree_walker_module_directives.xml
@@ -136,12 +136,6 @@
 <module name="Regexp">
     <property name="illegalPattern" value="true"/>
     <property name="ignoreComments" value="true"/>
-    <property name="format" value="\bSneakyThrows\b"/>
-    <property name="message" value="String interpolation does not work in methods marked with lombok.SneakyThrows, use lombok.Lombok#sneakyThrow instead"/>
-</module>
-<module name="Regexp">
-    <property name="illegalPattern" value="true"/>
-    <property name="ignoreComments" value="true"/>
     <property name="format" value="\bString\.format\b"/>
     <property name="message" value="Use string interpolation instead (com.antkorwin:better-strings)"/>
 </module>
--- a/gitCoreJGit/src/main/java/com/virtuslab/gitcore/impl/jgit/BaseGitCoreBranch.java
+++ b/gitCoreJGit/src/main/java/com/virtuslab/gitcore/impl/jgit/BaseGitCoreBranch.java
@@ -62,9 +62,9 @@ public abstract class BaseGitCoreBranch implements IGitCoreBranch {
   protected @Unique RevWalk getTopoRevWalkFromPointedCommit() throws GitCoreException {
     @Unique RevWalk walk = new RevWalk(repo.getJgitRepo());
     walk.sort(RevSort.TOPO);
-    @Unique RevCommit commit = repo.gitCoreCommitToRevCommit(getPointedCommit());
+    ObjectId objectId = repo.gitCoreCommitToObjectId(getPointedCommit());
     try {
-      walk.markStart(commit);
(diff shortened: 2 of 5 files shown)


gitCoreJGit/src/main/java/com/virtuslab/gitcore/impl/jgit/BaseGitCoreBranch.java
```diff
@@ -74,12 +74,19 @@ public synchronized GitCoreCommit getPointedCommit() throws GitCoreException {
   @Override
   public List<IGitCoreCommit> deriveCommitsUntil(IGitCoreCommit upToCommit) throws GitCoreException {
     LOG.debug(() -> "Entering: branch = '${getFullName()}', upToCommit = ${upToCommit}");
+    @Unique RevWalk walk = getTopoRevWalkFromPointedCommit();
+    try {
+      walk.markUninteresting(walk.parseCommit(repo.gitCoreCommitToObjectId(upToCommit)));
```
PawelLipski, line 79:
```java
	/**
	 * Include {@link RevFlag#UNINTERESTING} boundary commits after all others.
	 * In {@link ObjectWalk}, objects associated with such commits (trees,
	 * blobs), and all other objects marked explicitly as UNINTERESTING are also
	 * included.
	 * <p>
	 * A boundary commit is a UNINTERESTING parent of an interesting commit that
	 * was previously output.
	 */
	BOUNDARY;
```

config/checkstyle/tree_walker_module_directives.xml
```diff
@@ -121,12 +121,6 @@
     <property name="format" value="\bpeek\b"/>
     <property name="message" value="Usage of the prohibited peek method"/>
 </module>
-<module name="Regexp">
-    <property name="illegalPattern" value="true"/>
-    <property name="ignoreComments" value="true"/>
-    <property name="format" value="\bSneakyThrows\b"/>
-    <property name="message" value="String interpolation does not work in methods marked with lombok.SneakyThrows, use lombok.Lombok#sneakyThrow instead"/>
```
mkondratek, line 128: Is it no longer true?

gitCoreJGit/src/main/java/com/virtuslab/gitcore/impl/jgit/BaseGitCoreBranch.java
```diff
@@ -74,12 +74,19 @@ public synchronized GitCoreCommit getPointedCommit() throws GitCoreException {
...
+    }
+    walk.sort(RevSort.BOUNDARY);
 
     LOG.debug("Starting revwalk");
-    return Iterator.ofAll(getTopoRevWalkFromPointedCommit())
+    return Iterator.ofAll(walk.iterator())
         .takeUntil(revCommit -> revCommit.getId().getName().equals(upToCommit.getHash().getHashString()))
         .map(revCommit -> {
-          LOG.debug(() -> revCommit.getId().getName());
+          LOG.debug(() -> "* " + revCommit.getId().getName());
```
mkondratek, line 89: string interpolations here?

config/checkstyle/tree_walker_module_directives.xml
PawelLipski: It was likely never true in the first place, I've only ever observed that in `binding` module and thought that's the fault of `@SneakyThrows`, but the actual reason was/is sth different... still dunno what specifically

gitCoreJGit/src/main/java/com/virtuslab/gitcore/impl/jgit/BaseGitCoreBranch.java
PawelLipski: IMO more readable without here :) string interpolations IMO make less sense when there's just one value to be interpolated at the beginning or end of a short string

PR #319, opened 2020-06-01
Add merge --ff-only action

mkondratek:
- add action to merge (`--ff-only`-like, via `fetch . child:parent`) 
- resolve fetch is already running issue
- add dark machete file icon


frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/common/ActionUtils.java
```diff
@@ -74,4 +79,29 @@ public static String syncToRemoteStatusRelationToReadableBranchDescription(SyncT
         Case($(), "in unknown status '${relation.toString()}' to its remote"));
     return "the branch is ${desc}";
   }
+
+  static Task.Backgroundable getFetchBackgroundable(Project project,
+      GitRepository gitRepository,
+      String refspec,
+      GitRemote remote,
+      final String taskTitle) {
+    return new Task.Backgroundable(project, taskTitle, true) {
```
PawelLipski, line 88: Name `true`

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/common/BaseMergeBranchIntoParentAction.java
```diff
@@ -0,0 +1,139 @@
...
+        // in case of root branch we do not want to show this option at all
+        presentation.setEnabledAndVisible(false);
+      }
+      return;
+    }
+
+    var gitMacheteNonRoot = gitMacheteBranch.get().asNonRootBranch();
+    var syncToParentStatus = gitMacheteNonRoot.getSyncToParentStatus();
+
+    if (SyncToParentStatus.InSync == syncToParentStatus) {
```
PawelLipski, line 75: Nit: flip comparison side for clarity

frontendActions/src/main/resources/actions/ResourceBundle.properties
```diff
@@ -10,13 +10,17 @@ action.GitMachete.RebaseSelectedBranchOntoParentAction.GitMacheteContextMenu.tex
...
 action.GitMachete.PullSelectedBranchAction.text=Pull Selected Branch
-action.GitMachete.PullSelectedBranchAction.GitMacheteContextMenu.text=_Pull Branch
+action.GitMachete.PullSelectedBranchAction.GitMacheteContextMenu.text=P_ull Branch
 action.GitMachete.PullSelectedBranchAction.description=Pull this branch
 
 action.GitMachete.PushSelectedBranchAction.text=Push Selected Branch...
 action.GitMachete.PushSelectedBranchAction.GitMacheteContextMenu.text=_Push Branch...
 action.GitMachete.PushSelectedBranchAction.description=Push this branch using push dialog
 
+action.GitMachete.MergeSelectedBranchIntoParentAction.text=Merge Selected Branch Into Parent
```
PawelLipski, line 20: Hmm how about renaming that to `Fast forward parent to match this/current/selected/etc. branch`?

PawelLipski: Coz `merge X into Y`:
1. IMO requires more thinking about what branch is changed and what branch remains unchanged
1. `merge` for most people suggests a non-fast-forward merge, where an actual  merge commit is created

PawelLipski: If you agree, pls cascade this change into other affected strings/names in the codebase

src/main/resources/META-INF/plugin.xml
```diff
@@ -103,6 +108,12 @@
                 <!--                <override-text place="GitMacheteToolbar"/>-->
             </action>
 
+            <action id="GitMachete.MergeCurrentBranchIntoParentAction"
+                    class="com.virtuslab.gitmachete.frontend.actions.toolbar.MergeCurrentBranchIntoParentAction"
+                    icon="AllIcons.Vcs.Merge">
+                <!--                <override-text place="GitMacheteToolbar"/>-->
+            </action>
+
             <action id="GitMachete.OpenMacheteFileAction"
```
PawelLipski, line 117: Hmm is there any reason for `edit` action being specifically b/w `merge` and `slide-out`? likely the actions should be rearranged (here & in context menu)

frontendActions/src/main/resources/actions/ResourceBundle.properties
mkondratek: Very good point. It's a perfect idea to reflect that it is not some gitish merge. It's our, cool, nocheckout-requiring, fast forward merge.

We could do a similar thing to the pull action. WDYT?

PawelLipski: Hmmm what would you suggest for pull action then? :thinking:

mkondratek: Fast forward current/selected branch to match remote?

mkondratek: `Fast forward this/current/selected/etc. (local?) branch to match remote`?

PawelLipski: Hmm the problem I see with `Fast forward ... (local) branch to match remote`:  that's a bit ambiguous; it's not clear from this phrasing whether the fetch happens first or not, as opposed to e.g `Pull (fast-forward only) this branch`, where it's clear that the fetch does happen indeed

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/common/BaseFastForwardParentToMatchBranchAction.java
```diff
@@ -0,0 +1,141 @@
...
+
+/**
+ * Expects DataKeys:
+ * <ul>
+ *  <li>{@link DataKeys#KEY_GIT_MACHETE_REPOSITORY}</li>
+ *  <li>{@link CommonDataKeys#PROJECT}</li>
+ * </ul>
+ */
+public abstract class BaseFastForwardParentToMatchBranchAction extends GitMacheteRepositoryReadyAction
+    implements
```
PawelLipski, line 35: Join lines (+maybe reconfigure spotless so that it stops happening if possible)

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/common/BaseFastForwardParentToMatchBranchAction.java
```diff
@@ -0,0 +1,141 @@
...
+        presentation.setText("Fast Forward Parent To Match Current Branch");
+      }
+
+      var parentName = gitMacheteNonRoot.getUpstreamBranch().getName();
+      presentation.setDescription("Fast forward branch '${parentName}' to match '${branchName.get()}'");
+
+    } else {
+      presentation.setEnabled(false);
+      var desc = Match(syncToParentStatus).of(
+          Case($(SyncToParentStatus.InSync), "in sync to its parent"),
```
PawelLipski, line 89: Impossible case?

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/common/BaseFastForwardParentToMatchBranchAction.java
```diff
@@ -0,0 +1,141 @@
...
+
+    if (gitMacheteBranch.isDefined()) {
+      if (gitRepository.isDefined()) {
+        assert gitMacheteBranch.get().isNonRootBranch() : "Provided machete branch to merge is a root";
+        doMerge(project, gitRepository.get(), gitMacheteBranch.get().asNonRootBranch());
+      } else {
+        LOG.warn("Skipping the action because no VCS repository is selected");
+      }
+    } else {
+      LOG.warn("Skipping the action because machete branch to merge is undefined");
```
PawelLipski, line 115: `s/merge/fast forward` here and in other left over places in this class

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/common/BaseFastForwardParentToMatchBranchAction.java
```diff
@@ -0,0 +1,141 @@
...
+      return;
+    } else if (parentTrackingInfo == null) {
+      LOG.warn("No branch tracking info for parent branch ${gitMacheteNonRootBranch.getUpstreamBranch().getName()}");
+      return;
+    }
+
+    var localFullName = trackingInfo.getLocalBranch().getFullName();
+    var parentLocalFullName = parentTrackingInfo.getLocalBranch().getFullName();
+
+    // On the other hand this refspec has no '+' sign.
```
PawelLipski, line 134: `On the other hand`... but from what? ;)

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/common/BaseFastForwardParentToMatchBranchAction.java
```diff
@@ -0,0 +1,141 @@
...
+    // This is because the fetch from local remotes to local heads must behave fast-forward-like.
+    var refspecRemoteLocal = "${localFullName}:${parentLocalFullName}";
+
+    // Remote set to '.' (dot) is just the local repository.
+    getFetchBackgroundable(project, gitRepository, refspecRemoteLocal, GitRemote.DOT, "Merging...").queue();
```
PawelLipski, line 139: Ditto, stray `Merging`

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/contextmenu/FastForwardParentToMatchSelectedBranchAction.java
```diff
@@ -0,0 +1,26 @@
...
+import com.intellij.openapi.actionSystem.CommonDataKeys;
+import io.vavr.control.Option;
+
+import com.virtuslab.gitmachete.frontend.actions.common.BaseFastForwardParentToMatchBranchAction;
+import com.virtuslab.gitmachete.frontend.datakeys.DataKeys;
+
+/**
+ * Expects DataKeys:
+ * <ul>
+ *  <li>{@link DataKeys#KEY_GIT_MACHETE_REPOSITORY}</li>
```
PawelLipski, line 15: Soon to be reworked by myself on a separate PR...

src/main/resources/META-INF/plugin.xml
```diff
@@ -111,6 +116,12 @@
                 <!--                <override-text place="GitMacheteToolbar"/>-->
             </action>
 
+            <action id="GitMachete.FastForwardParentToMatchCurrentBranchAction"
+                    class="com.virtuslab.gitmachete.frontend.actions.toolbar.FastForwardParentToMatchCurrentBranchAction"
+                    icon="AllIcons.Vcs.Merge">
```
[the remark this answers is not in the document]
PawelLipski: Note that we might want to add the equivalent of `git machete update/traverse --merge`, which does a possibly non-FF merge of **parent into child** (as opposed to FF-only merge of child into parent that we do here)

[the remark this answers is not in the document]
PawelLipski: (to satisfy the more rebase-fearing part of population)

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/common/ActionUtils.java
```diff
@@ -74,4 +79,29 @@ public static String syncToRemoteStatusRelationToReadableBranchDescription(SyncT
         Case($(), "in unknown status '${relation.toString()}' to its remote"));
     return "the branch is ${desc}";
   }
+
+  static Task.Backgroundable getFetchBackgroundable(Project project,
+      GitRepository gitRepository,
+      String refspec,
+      GitRemote remote,
+      final String taskTitle) {
+    return new Task.Backgroundable(project, taskTitle, /* canByCancelled */ true) {
```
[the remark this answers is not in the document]
PawelLipski: Also, I think that an package-private class extending `Task.Backgroundable` would be more natural here rather than a static util method

src/main/resources/META-INF/plugin.xml
mkondratek: _Slack discussion summary_
Target: 
image

src/main/resources/META-INF/plugin.xml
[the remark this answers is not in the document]
mkondratek: Yeah, the present icon is confusing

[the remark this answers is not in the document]
mkondratek: WDYT?

image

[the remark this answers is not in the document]
PawelLipski: Ok, let's got for this one

src/main/resources/META-INF/plugin.xml
```diff
@@ -119,7 +119,7 @@
 
             <action id="GitMachete.FastForwardParentToMatchCurrentBranchAction"
                     class="com.virtuslab.gitmachete.frontend.actions.toolbar.FastForwardParentToMatchCurrentBranchAction"
-                    icon="AllIcons.Vcs.Merge">
+                    icon="AllIcons.Diff.ApplyNotConflictsLeft">
```
PawelLipski, line 122: There's a problem with hijacking icons like that: they're subject to change b/w Intellij versions, and at some point they might be changed so that they still will be adequate for their original purpose, but no longer match our use case. Maybe instead copy the original icon to our project?... TBD how that works with the IntelliJ's Apache license.

PawelLipski: This comment likely already applies to a few other off-label/hijacked uses of icons in the plugin as well.

src/main/resources/META-INF/plugin.xml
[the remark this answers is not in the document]
PawelLipski: But with the reservations wrt. off-label icon use, see the other comment

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/contextmenu/FastForwardParentToMatchSelectedBranchAction.java
PawelLipski: PR #322

frontendActions/src/main/resources/actions/ResourceBundle.properties
PawelLipski: Resolved to use `Pull (fast-forward only)`

src/main/resources/META-INF/plugin.xml
mkondratek: For now, I would like to keep things like this. So let's use JB icons tath (presently) just looks (close to) accurate.

Your concerns reasonable (and actually they are not new in the scope of our plugin) and custom icon set deserve their own issue definitely (#326). Assuming that icons creation will take some time we shall carefully think through our targets, requirements, and options.

PawelLipski: Hmm there are in fact 2 separate issues here:
1. whether to provide our own icons or rely or IntelliJ's icons for certain actions
1. if we rely on IntelliJ's icons, then make sure that our UX isn't susceptible to IntelliJ changing the look of those icons b/w versions

PawelLipski: (2) is relatively easy to do, we can e.g. just copy the icons directly into our project

mkondratek: You are right. I will take care of the second issue on separate PR

mkondratek: Actually this seems to be a good point to do this since we have seen `the frontendIcons module` created.

src/main/resources/META-INF/plugin.xml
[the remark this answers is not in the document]
mkondratek: discussion resolved in https://github.com/VirtusLab/git-machete-intellij-plugin/pull/319#discussion_r435226926

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/toolbar/FetchAllRemotesAction.java
```diff
@@ -33,9 +33,12 @@ public void update(AnActionEvent anActionEvent) {
     super.update(anActionEvent);
 
     var project = getProject(anActionEvent);
+    var presentation = anActionEvent.getPresentation();
```
PawelLipski, line 36: Hmmm wait... 1) is this change the entire fix for #321? 2) if so, how does it work? :thinking:

mkondratek: yup, the description had been changed while running fetch but has not been changed after that back to the correct description. that's all

mkondratek: mutable objects

image

PawelLipski: Oh ok, I didn't notice the newly-added `else` ;)

src/main/resources/META-INF/plugin.xml
mkondratek: PR #328

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/common/BaseFastForwardParentToMatchBranchAction.java
```diff
@@ -0,0 +1,139 @@
...
+      var parentName = gitMacheteNonRoot.getUpstreamBranch().getName();
+      presentation.setDescription("Fast forward branch '${parentName}' to match '${branchName.get()}'");
+
+    } else {
+      presentation.setEnabled(false);
+      var desc = Match(syncToParentStatus).of(
+          Case($(SyncToParentStatus.InSyncButForkPointOff), "in sync to its parent but fork point is off"),
+          Case($(SyncToParentStatus.MergedToParent), "merged into parent"),
+          Case($(SyncToParentStatus.OutOfSync), "out of sync to its parent"),
+          Case($(), "in unknown status '${syncToParentStatus.toString()}' to its parent"));
```
micpiotrowski, line 92: IIRC this piece of code is also somewhere else. If it's true maybe extract it to utils?

mkondratek: I would rather keep it like that. 
- this is the only place where it is used 
- we do not want to perform too much call within `update` (performance)

PR #322, opened 2020-06-03
Redefine DataKey contracts with interfaces instead of Javadoc

PawelLipski:
Almost 200 lines less as a result :)


scripts/run-pre-build-checks
```diff
@@ -6,7 +6,6 @@ set -e
 x_option=${-//[^x]/}
 
 ./scripts/enforce-change-notes-updated-on-release-pr
-./scripts/enforce-data-keys-javadoc-for-actions
```
mkondratek, line 9: :+1:

mkondratek:
The `common` package seems to be overloaded now. Can we extract `.+Action` classes from it?

PawelLipski:
> The common package seems to be overloaded now. Can we extract .+Action classes from it?

Done, reorganized the package structure a bit

PR #328, opened 2020-06-04
Refactor/icons kidnapping

mkondratek:
> There's a problem with hijacking icons like that: they're subject to change b/w Intellij versions, and at some point they might be changed so that they still will be adequate for their original purpose, but no longer match our use case. Maybe instead copy the original icon to our project?... TBD how that works with the IntelliJ's Apache license.

https://github.com/VirtusLab/git-machete-intellij-plugin/pull/319#discussion_r435226926


PawelLipski:
Haha I don't think `kidnap` can be used in the figurative sense as easily as `hijack` can TBH ;D

image

image

frontendIcons/src/main/java/icons/MacheteIcons.java
```diff
@@ -6,6 +6,26 @@
 
 public final class MacheteIcons {
   private MacheteIcons() {}
-  public static final Icon ICON = IconLoader.getIcon("/macheteFileIcon.svg");
+
```
PawelLipski, line 9: While you're here, pls drop a comment why this must be `icons` and not e.g. `com.virtuslab.gitmachete.icons` package

mkondratek: @micpiotrowski

mkondratek: okay, I got it
https://www.jetbrains.org/intellij/sdk/docs/reference_guide/work_with_icons_and_images.html

PR #337, opened 2020-06-17
Add basic time measurement to frontendGraphImpl

PawelLipski:


frontendGraphApi/src/main/java/com/virtuslab/gitmachete/frontend/graph/api/repository/NullRepositoryGraph.java
```diff
@@ -3,13 +3,19 @@
...
 
 import com.virtuslab.gitmachete.frontend.graph.api.elements.GraphEdge;
 import com.virtuslab.gitmachete.frontend.graph.api.items.IGraphItem;
 import com.virtuslab.gitmachete.frontend.graph.api.render.parts.IRenderPart;
 
-public class NullRepositoryGraph implements IRepositoryGraph {
+public final class NullRepositoryGraph implements IRepositoryGraph {
+
+  @Getter
+  private static final NullRepositoryGraph instance = new NullRepositoryGraph();
```
micpiotrowski, line 16: Shouldn't `instance` be capitalized?

PawelLipski: Hmm judging by checkstyle settings, we decided at some point to make an exception for `instance`s and `cache`s - which kinda makes sense, they're more of "fields of the class" than actual "constants". Although I do agree that this distinction is rather vague

PR #353, opened 2020-06-23
Enclose demo table implementation within Impl module

mkondratek:
- merge `*tableDemo` module into `*tableImpl` ~(compile time reduction)~
- reduce logging in the actions' updates


PawelLipski:
> merge *tableDemo module into *tableImpl (compile time reduction)

Technically, it's not merging the modules that leads to compile time reduction, but rather cutting the dependency from `frontendActions` to `frontendUiTable*`

frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/DemoGraphTableFactory.java
```diff
@@ -0,0 +1,12 @@
+package com.virtuslab.gitmachete.frontend.ui.impl.table;
+
+import com.intellij.ui.table.JBTable;
+
+import com.virtuslab.gitmachete.frontend.ui.api.table.IDemoGraphTableFactory;
+
+public class DemoGraphTableFactory implements IDemoGraphTableFactory {
+  @Override
+  public JBTable create() {
```
PawelLipski, line 9: Since it's always returning the very same cached instance, maybe do rename `create` to `getInstance` and `(I)DemoGraphTableFactory` to `(I)DemoGraphTableProvider`

PawelLipski:
And now UI and Actions can compile in parallel:

image

PR #361, opened 2020-06-24
Rectify release-related scripts

PawelLipski:


scripts/enforce-pre-release-version-correct-use
```diff
@@ -7,27 +7,18 @@ source "$self_dir"/utils.sh
...
-      die 'Pre-release version is not allowed on master/release/hotfix/backport branch tip'
-  else
-    [[ $head_pre_release ]] || \
-      die 'Non-master/release/hotfix/backport branch tip must have a pre-release version (a.b.c-d, not a.b.c)'
-  fi
+if [[ $cb == master || $cb == release/* || $cb == hotfix/* || $cb == backport/* ]]; then
+  [[ -z $head_pre_release ]] || \
+    die 'Pre-release version is not allowed on master/release/hotfix/backport branch tip'
+elif [[ $cb != develop ]]; then
+  # Tip of develop can legitimately have either a pre-release version (when a regular PR has been merged)
```
PawelLipski, line 20: This hasn't been taken into account previously, hence this PR is a bugfix.

frontendExternalSystem/src/main/java/com/virtuslab/gitmachete/frontend/externalsystem/settings/MacheteExecutionSettings.java
```diff
@@ -8,7 +8,7 @@
 
 @Data
 @EqualsAndHashCode(callSuper = true)
-@ToString(callSuper = true)
+@ToString(callSuper = false)
```
PawelLipski, line 11: Unrel: `false` is the default, but let's still require it to be explicit in case a class extends sth different then Object.

scripts/enforce-pre-release-version-correct-use
```diff
@@ -7,27 +7,18 @@ source "$self_dir"/utils.sh
 
 parse_version_from_current_wd head
 
-if [[ ${CI_PULL_REQUEST-} ]]; then
```
PawelLipski, line 10: This check didn't add much value, removed

scripts/run-pre-build-checks
```diff
@@ -5,12 +5,13 @@ set -e
 # `$-` yields the list of the options of the current shell; we'll only interested in whether `x` is present
 x_option=${-//[^x]/}
 
-./scripts/enforce-change-notes-updated-on-release-pr
+# pass -x option if this script is run with -x option itself
+bash ${x_option:+-x} ./scripts/enforce-change-notes-updated-on-pr-to-master
```
PawelLipski, line 9: Let's turn on shell debug in case of the scripts that execute completely differently on CI than locally

PR #329, opened 2020-06-04
Prohibit Option#toString and io.vavr.collection.<X>#peek using FenumChecker

PawelLipski:
~Depends on changes from https://github.com/antkorwin/better-strings/pull/14/files~ => changes dependent on better-strings v0.4 were moved to #362

PR #362, opened 2020-06-24
Bump better-strings to v0.4 (x.toString() generated instead of String.valueOf(x))

PawelLipski:
Extracted from PR #329 so that it isn't blocked on v0.4 being published

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,17 +16,6 @@ jobs:
           name: Check formatting
           command: ./gradlew spotlessCheck
       - run:
-          name: '[TEMP] Publish better-strings annotation processor locally'
-          command: |
-            apt-get update
-            apt-get install -y maven
-
-            git clone https://github.com/PawelLipski/better-strings.git ~/better-strings
-            cd ~/better-strings
-            # https://github.com/antkorwin/better-strings/pull/14/files
-            git checkout 8d419868204fc7810de4d7685e7fd9b0793813f9
-            mvn install
-      - run:
           name: Compile production code
           # Given the RAM limits on CI (4GB), max-workers=2 is necessary to prevent OOMs.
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -16,6 +16,19 @@ jobs:
       - run:
           name: Check formatting
           command: ./gradlew spotlessCheck
+      - run:
+          name: '[TEMP] Publish better-strings annotation processor locally'
```
PawelLipski, line 20: To be removed once v0.4 is published

build.gradle
```diff
@@ -208,12 +208,22 @@ subprojects {
   // This needs to be enabled in each subproject by default because there's going to be no warning
   // if this annotation processor isn't run in any subproject (the strings will be just interpreted verbatim, without interpolation applied).
   // We'd only capture that in CI's post-compile checks by analyzing constants in class files.
+  repositories {
```
PawelLipski, line 211: To be removed once v0.4 is published

build.gradle
```diff
@@ -208,12 +208,22 @@ subprojects {
...
+
   dependencies {
     def betterStringsArtifact = [ group: 'com.antkorwin', name: 'better-strings', version: betterStringsVerson ]
     annotationProcessor(betterStringsArtifact)
     testAnnotationProcessor(betterStringsArtifact)
   }
 
+  gradle.projectsEvaluated {
+    tasks.withType(JavaCompile) {
+      options.compilerArgs += '-AcallToStringExplicitlyInInterpolations'
```
PawelLipski, line 223: This is the new option added in v0.4

build.gradle
PawelLipski: ...for which there's actually no rush for us thanks to jitpack.io <3

PR #367, opened 2020-06-24
Migrate from 2019.3 to 2020.1 earliest supported version

mkondratek:
- `override-text` enabled
- `externalSystemIconProvider` enabled

**IMPORTANT NOTE - GitFetchSupport**

It turned out that custom `GitFetchSupportImpl` cannot be replaced with the one provided by IntelliJ 2020.1.
The updated one (from 2020.1) does provide the ability to "update" non-current branches indeed.
However, it does not allow us to fetch the current branch. 
The option - `--update-head-ok` required by fetch cannot be delivered to the `git4idea.commands.GitImpl#fetch(...)`.

Because of that, the decision has been made to keep our custom implementation for now.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -82,7 +82,7 @@ allprojects {
     intellijVersions = [
       earliestSupported: '2020.1',
       // An arbitrary list of versions between earliest supported (excl.) and current (excl.), for binary compatibility checks
-      // inBetweenForCompatibilityChecks: [ '2020.1' ],
+      inBetweenForCompatibilityChecks: [],
       current: '2020.1.2',
     ]
     intellijVersions.buildTarget = intellijVersions.current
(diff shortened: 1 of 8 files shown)


build.gradle
```diff
@@ -80,9 +80,9 @@ allprojects {
 
     // Update the range of UI-tested versions in .circle/config.yml when any of the below is updated
     intellijVersions = [
-      earliestSupported: '2019.3',
+      earliestSupported: '2020.1',
       // An arbitrary list of versions between earliest supported (excl.) and current (excl.), for binary compatibility checks
-      inBetweenForCompatibilityChecks: [ '2020.1' ],
+      // inBetweenForCompatibilityChecks: [ '2020.1' ],
```
PawelLipski, line 85: Just leave an empty list here

PR #370, opened 2020-06-25
Highlighting of invalid indentation

micpiotrowski:
Maybe not the simplest and clearest one but it is: implementation of invalid indentation highlighting based on Machete Annotator class.
Probably better solution would be use a flexer to do this but something went wrong wen I try using it and I have no idea how to assign custom tooltip error text when mouse is hovered on selected error (see the first commit in this PR)

### DIFF
--- a/frontendFile/src/main/grammar/Machete.bnf
+++ b/frontendFile/src/main/grammar/Machete.bnf
@@ -15,7 +15,7 @@
 
 simpleFile ::= item_*
 
-// Even if we don't have comments in our machete file this element must be here - language parser definition
+// Even if we don't have comments in our machete file, the COMMENT element must be here - language parser definition
 // (in file MacheteParserDefinition) enforces us to define element that represent comments
 private item_ ::= (entry|COMMENT|(INDENTATION? EOL))
 
--- a/frontendFile/src/main/grammar/Machete.bnf
+++ b/frontendFile/src/main/grammar/Machete.bnf
@@ -17,7 +17,7 @@ simpleFile ::= item_*
 
 // Even if we don't have comments in our machete file this element must be here - language parser definition
 // (in file MacheteParserDefinition) enforces us to define element that represent comments
-private item_ ::= (entry|COMMENT|EOL)
+private item_ ::= (entry|COMMENT|(INDENTATION? EOL))
 
 entry ::= INDENTATION? branch CUSTOM_ANNOTATION?
 
--- a/frontendFile/src/main/java/com/virtuslab/gitmachete/frontend/file/highlighting/MacheteAnnotator.java
+++ b/frontendFile/src/main/java/com/virtuslab/gitmachete/frontend/file/highlighting/MacheteAnnotator.java
@@ -48,7 +48,7 @@ public class MacheteAnnotator implements Annotator {
 
   private void processIndentationElement(PsiElement element, AnnotationHolder holder) {
     PsiElement parent = element.getParent();
-    assert parent != null : "Element has not parent";
+    assert parent != null : "Element has no parent";
 
     if (parent instanceof MacheteFile) {
       return;
@@ -78,8 +78,11 @@ public class MacheteAnnotator implements Annotator {
 
     var thisIndentationText = element.getText();
 
-    if (thisIndentationText.chars().filter(c -> c != indentationParameters.indentationCharacter).count() > 0) {
-      holder.newAnnotation(HighlightSeverity.ERROR, "Indentation character does not match the first indented line")
+    var wrongIndentChar = thisIndentationText.chars().filter(c -> c != indentationParameters.indentationCharacter).findFirst();
+    if (wrongIndentChar.isPresent()) {
+      holder.newAnnotation(HighlightSeverity.ERROR,
+          "Indentation character (${indentCharToName((char)wrongIndentChar.getAsInt())}) "
+              + "does not match the indentation character in first indented line (${indentCharToName(indentationParameters.indentationCharacter)})")
           .range(element).create();
       return;
     }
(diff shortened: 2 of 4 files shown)


frontendFile/src/main/java/com/virtuslab/gitmachete/frontend/file/highlighting/MacheteAnnotator.java
```diff
@@ -37,4 +45,129 @@ public void annotate(PsiElement element, AnnotationHolder holder) {
           .range(branch).create();
     }
   }
+
+  private void processIndentationElement(PsiElement element, AnnotationHolder holder) {
+    PsiElement parent = element.getParent();
+    assert parent != null : "Element has not parent";
```
PawelLipski, line 51:
```suggestion
    assert parent != null : "Element has no parent";
```

frontendFile/src/main/java/com/virtuslab/gitmachete/frontend/file/highlighting/MacheteAnnotator.java
```diff
@@ -37,4 +45,129 @@ public void annotate(PsiElement element, AnnotationHolder holder) {
...
+      holder.newAnnotation(HighlightSeverity.ERROR,
+          "Indentation width is not multiple of ${indentationParameters.indentationWidth}" +
+              " as first indented line suggests")
+          .range(element).create();
+    }
+
+    thisLevel = thisIndentationText.length() / indentationParameters.indentationWidth;
+
+    if (hasPrevLevelCorrectWidth && thisLevel > prevLevel + 1) {
+      holder.newAnnotation(HighlightSeverity.ERROR, "Indentation level does not math")
```
PawelLipski, line 97: `match`

frontendFile/src/main/grammar/Machete.bnf
```diff
@@ -17,7 +17,7 @@ simpleFile ::= item_*
 
 // Even if we don't have comments in our machete file this element must be here - language parser definition
```
PawelLipski, line 18: Nit:
```suggestion
// Even if we don't have comments in our machete file, the COMMENT element must be here - language parser definition
```

frontendFile/src/main/java/com/virtuslab/gitmachete/frontend/file/highlighting/MacheteAnnotator.java
```diff
@@ -37,4 +45,129 @@ public void annotate(PsiElement element, AnnotationHolder holder) {
...
+      hasPrevLevelCorrectWidth = true;
+    } else {
+      var prevIndentationText = prevIndentationNodeOption.get().getText();
+      hasPrevLevelCorrectWidth = prevIndentationText.length() % indentationParameters.indentationWidth == 0;
+      prevLevel = prevIndentationText.length() / indentationParameters.indentationWidth;
+    }
+
+    var thisIndentationText = element.getText();
+
+    if (thisIndentationText.chars().filter(c -> c != indentationParameters.indentationCharacter).count() > 0) {
```
PawelLipski, line 81: Nit: `xd.filter(f).count > 0` can likely be replaced with `xd.exists(f)` or sth like that

frontendFile/src/main/java/com/virtuslab/gitmachete/frontend/file/highlighting/MacheteAnnotator.java
```diff
@@ -37,4 +45,129 @@ public void annotate(PsiElement element, AnnotationHolder holder) {
...
+      holder.newAnnotation(HighlightSeverity.ERROR, "Indentation character does not match the first indented line")
```
PawelLipski, line 82: Would be nice if the message also mentions what's the expected & what's the actual indent character

frontendFile/src/main/java/com/virtuslab/gitmachete/frontend/file/highlighting/MacheteAnnotator.java
PawelLipski: Maybe even for super-clarity sth like:
```suggestion
      holder.newAnnotation(HighlightSeverity.ERROR, "Too much indent on this line")
```

frontendFile/src/main/java/com/virtuslab/gitmachete/frontend/file/highlighting/MacheteAnnotator.java
micpiotrowski: `exists()` seems to not exists for `IntStream` :/

micpiotrowski: Oh, ok, I can use `anyMatch` instead

micpiotrowski: Anyway, I done this in another way

frontendFile/src/main/java/com/virtuslab/gitmachete/frontend/file/highlighting/MacheteAnnotator.java
```diff
@@ -78,8 +78,11 @@ private void processIndentationElement(PsiElement element, AnnotationHolder hold
 
     var thisIndentationText = element.getText();
 
-    if (thisIndentationText.chars().filter(c -> c != indentationParameters.indentationCharacter).count() > 0) {
-      holder.newAnnotation(HighlightSeverity.ERROR, "Indentation character does not match the first indented line")
+    var wrongIndentChar = thisIndentationText.chars().filter(c -> c != indentationParameters.indentationCharacter).findFirst();
```
PawelLipski, line 81: Hmm actually what happens when the file is empty or there's no indent on any line?

micpiotrowski: Nothing special - this piece of code will even not being executed

PR #379, opened 2020-06-29
Prohibit inner (non-static nested) classes until a bug in NullnessChecker is fixed

PawelLipski:
See https://github.com/typetools/checker-framework/issues/3407

### DIFF
--- a/frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/pushdialog/GitPushDialog.java
+++ b/frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/pushdialog/GitPushDialog.java
@@ -37,6 +37,9 @@ import com.intellij.util.ui.components.BorderLayoutPanel;
 import io.vavr.collection.List;
 import net.miginfocom.swing.MigLayout;
 import org.checkerframework.checker.guieffect.qual.UIEffect;
+import org.checkerframework.checker.initialization.qual.Initialized;
+import org.checkerframework.checker.initialization.qual.NotOnlyInitialized;
+import org.checkerframework.checker.initialization.qual.UnderInitialization;
 import org.checkerframework.checker.nullness.qual.Nullable;
 
 public final class GitPushDialog extends DialogWrapper implements VcsPushUi {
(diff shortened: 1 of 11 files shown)


scripts/prohibit-inner-classes
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+set -e -o pipefail -u
+
+self_dir=$(cd "$(dirname "$0")" &>/dev/null; pwd -P)
+source "$self_dir"/utils.sh
+
+if git grep -En '^ +[^/*"]*\bclass ' -- '*.java' | grep -v '\bstatic '; then
```
PawelLipski, line 8: It's likely doable with Checkstyle `Regexp`, but would require negative lookbehinds or some other brittle stuff

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/pushdialog/GitPushDialog.java
```diff
@@ -61,42 +64,39 @@ public GitPushDialog(
...
     this.isForcePushRequired = isForcePushRequired;
     this.project = project;
-    this.pushAction = new PushSwingAction();
-    this.pushAction.putValue(DEFAULT_ACTION, Boolean.TRUE);
-
-    var dialog = new MyVcsPushDialog(project, selectedRepositories.asJava(), pushSource) {
-      @Override
-      @UIEffect
-      public @Nullable VcsPushOptionValue getAdditionalOptionValue(PushSupport support) {
-        return GitPushDialog.this.getAdditionalOptionValue(support);
```
PawelLipski, line 71: @mkondratek note that this still wasn't correct and only passed thanks to another (or maybe the same?) bug in CF... `GitPushDialog.this` isn't `@Initialized` when `MyVcsPushDialog` is instantiated :/

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/pushdialog/GitPushDialog.java
```diff
@@ -61,42 +64,39 @@ public GitPushDialog(
...
-      }
-
-      @Override
-      @UIEffect
-      public JRootPane getRootPane() {
-        return GitPushDialog.this.getRootPane();
-      }
-    };
 
+    @SuppressWarnings("nullness:assignment.type.incompatible") @Initialized MyVcsPushDialog dialog = new MyVcsPushDialog(this,
```
PawelLipski, line 68: At least each place we stretch the type system is clearly visible now + range of warning suppressions is as narrow as possible

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/pushdialog/GitPushDialog.java
PawelLipski: https://github.com/typetools/checker-framework/issues/3408

PR #382, opened 2020-06-30
Rework branchLayoutApi/Impl to facilitate different layout implementations

PawelLipski:
Another step towards discover implementation. See #383 on how that's actually used.


branchLayoutApi/src/main/java/com/virtuslab/branchlayout/api/BranchLayout.java
```diff
@@ -0,0 +1,54 @@
...
+
+import io.vavr.Tuple;
+import io.vavr.collection.List;
+import io.vavr.collection.Map;
+import io.vavr.control.Option;
+import lombok.Getter;
+import org.checkerframework.checker.interning.qual.UsesObjectEquals;
+
+@UsesObjectEquals
+public class BranchLayout implements IBranchLayout {
```
PawelLipski, line 11: Possibly `IBranchLayout` can be completely removed now that `BranchLayout` is in API... not sure yet

branchLayoutImpl/src/main/java/com/virtuslab/branchlayout/impl/BranchLayoutEntry.java
```diff
@@ -20,6 +22,7 @@
   private final @Nullable String customAnnotation;
 
   @Getter(onMethod_ = {@Override})
+  @With(onMethod_ = {@Override})
```
mkondratek, line 25: :open_mouth:

PR #387, opened 2020-06-30
Rename GitMacheteRepository to GitMacheteRepositorySnapshot

PawelLipski:


backendApi/src/main/java/com/virtuslab/gitmachete/backend/api/IGitMacheteRepositorySnapshot.java
```diff
@@ -0,0 +1,20 @@
+package com.virtuslab.gitmachete.backend.api;
+
+import io.vavr.collection.List;
+import io.vavr.control.Option;
+
+import com.virtuslab.branchlayout.api.IBranchLayout;
+
+public interface IGitMacheteRepositorySnapshot {
```
mkondratek, line 8: Can we have a few words of description about what do we mean by snapshot and why do we reflect this? It is (only) intuitive for me but some clarification would be nice.

backendImpl/src/main/java/com/virtuslab/gitmachete/backend/impl/GitMacheteNonRootBranch.java
```diff
@@ -56,10 +56,10 @@ public GitMacheteNonRootBranch(
 
     // Note: since the class is final, `this` is already @Initialized at this point.
 
-    // This is a hack necessary to create an immutable cyclic structure (children pointing at the parent & parent
-    // pointing at the children).
+    // This is a hack necessary to create an immutable cyclic structure
```
mkondratek, line 59: I can see the same fragment of code for `GitMacheteRootBranch`... could we move it to base class? :thinking:

PR #389, opened 2020-06-30
Release 0.5.1 (includes fixes to push dialog and release process)

PawelLipski:


.circleci/config.yml
```diff
@@ -127,7 +127,7 @@ jobs:
               --labels=backport
       - deploy:
           name: Publish plugin to Jetbrains Marketplace
-          command: ./gradlew publishPlugin
+          command: ./gradlew --max-workers=2 publishPlugin
```
PawelLipski, line 130: That's a workaround - the project still gets unnecessarily recompiled to certain extent. Probably to actually make sure that it doesn't happen I'd need to fix our use GrammarKit plugin, but that's not as urgent.

CHANGE-NOTES.html
```diff
@@ -6,4 +6,5 @@
   <li>Added help action</li>
   <li>Improved displaying of status</li>
   <li>Improved push action to automatically suggest force-push when needed</li>
+  <li>Improved stability of push dialog (v0.5.1)</li>
```
PawelLipski, line 9: I think we can approach change-notes in two ways, depending on the case:
1. either scrap the content completely and replace with new changelog - makes sense for minor&major releases
1. or just append new pieces of changelog - makes sense for patch releases like this one

PawelLipski: The automated checks in CI only require that the CHANGE-NOTES.html are _modified_ compared to master.

config/checker/com.intellij.astub
```diff
@@ -210,6 +210,13 @@ package com.intellij.openapi.project;
 @UIPackage package com.intellij.openapi.roots.ui.configuration.libraryEditor;
 @UIPackage package com.intellij.openapi.roots.ui.util;
 @UIPackage package com.intellij.openapi.ui;
+
+class DialogWrapper {
+  // `this` in the below method is @UnknownInitialization since this methods is called from DialogWrapper's c'tor
+  @Nullable javax.swing.JRootPane getRootPane(@UnknownInitialization DialogWrapper this);
```
PawelLipski, line 216: Sorry @mkondratek if/that I suggested to remove that before - as a rule of the thumb, any method in any class that's invoked from any of this class's c'tors should be marked as `@UnknownInitialization ... this`

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/pushdialog/GitPushDialog.java
```diff
@@ -65,38 +63,32 @@ public GitPushDialog(
     this.isForcePushRequired = isForcePushRequired;
     this.project = project;
 
-    @SuppressWarnings("nullness:assignment.type.incompatible") @Initialized MyVcsPushDialog dialog = new MyVcsPushDialog(this,
-        project, selectedRepositories.asJava(), pushSource);
+    MyVcsPushDialog dialog = new MyVcsPushDialog(project, selectedRepositories.asJava(), pushSource);
```
PawelLipski, line 66: Note that this line (instantiation of inner classes from c'tor before enclosing object is fully initialized) only passes NullnessChecker because of https://github.com/typetools/checker-framework/issues/3407

PawelLipski: So... let us pray that this time it can't NPE in the runtime

scripts/prohibit-inner-classes
```diff
@@ -5,7 +5,8 @@ set -e -o pipefail -u
 self_dir=$(cd "$(dirname "$0")" &>/dev/null; pwd -P)
 source "$self_dir"/utils.sh
 
-if git grep -En '^ +[^/*"]*\bclass ' -- '*.java' | grep -v '\bstatic '; then
+excluded_java_file=./frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/pushdialog/GitPushDialog.java
```
PawelLipski, line 8: :hankey:

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/pushdialog/GitPushDialog.java
```diff
@@ -65,38 +63,32 @@ public GitPushDialog(
...
     this.pushController = dialog.getPushController();
     this.listPanel = pushController.getPushPanelLog();
 
-    @SuppressWarnings("nullness:assignment.type.incompatible") @Initialized Action action = new PushSwingAction(this,
-        isForcePushRequired);
-    this.pushAction = action;
+    this.pushAction = new PushSwingAction();
```
PawelLipski, line 70: Same here, it's a false negative for Checker here

micpiotrowski:
Oooh no, I'm too late :/

Issue #380, opened 2020-06-29
Indents in .git/machete are assumed by IDEA to be 2 spaces (?) regardless of what's in the file


PawelLipski:
- [ ] e.g. in tab-indented file, pressing a tab will insert 2 spaces, which will of course cause red squiggles

PawelLipski:
Also, whatever we choose to be the default indent for .git/machete, discover functionality should save the file with this very default as well.

Issue #372, opened 2020-06-25
Do something useful in case when can't get branch list in MacheteAnnotator class


micpiotrowski:
For now we just do nothing in case when we cant get branches connected with opened machete file (with it repository).
Maybe some kind of user facing message that we can't check branch names correctness will be nice.

Issue #324, opened 2020-06-03
`Nothing to show` should be replaced with `Loading...` when the graph table is loading


PawelLipski:
`Nothing to show` actually should **never** be shown: it should be either `Loading...`, or a suggested layout of branches.

Issue #312, opened 2020-05-27
Add an action that opens the plugin's tab


PawelLipski:
Note that we had `com.virtuslab.gitmachete.frontend.actions.ideprobe.OpenTabAction` until it's been ~moved to test code on #311~ completely removed on #317; most likely this could be much less reverted + assigned some keyboard shortcut.

PawelLipski:
From https://github.com/VirtusLab/git-machete-intellij-plugin/issues/274#issuecomment-629621134:

> Adding `(Open) Machete Tab` as an action to the main menu is an option. Probably under `VCS > Git`.

PR #395, opened 2020-07-01
Small v0.6.0 issues

micpiotrowski:
* [issue #324] `setTextForEmptyTable` is used before Git Machete repository refreshing (so also when it's first loaded) to set `Loading...` text
* [isseu #380] Add few lines of code to `MacheteAnnotator` that set proper TAB key behavior depending on detected file indentation. Maybe this is not the best place for that code but I can't find any better place  :/
* [issue #393] `ExecutionResult` was added as a object to hold execution result (exit code, stdout and stderr). Now `PreRebaseHookExecutor` returns `IExecutionResult` (not simply `Integer` as before). In messages inside exceptions thrown from `PreRebaseHookExecutor` stdout and stderr are included.
* [issue #312] Add `OpenMacheteTabAction` that open Git Machete tab. Action is placed in `VCS` -> `Git` menu (but we must consider other location according to commient in `plugin.xml`). `Ctrl + Alt + M` is a shortcut associated with this action.
* [issue #372] Maybe not the best solution (technically) but IMO more elegant (than e.g. VCS notifier). If it's too bad we can consider VCS notifier (that should be simpler - more elegant to done). For now I using something like tooltip that appear on editor with proper message.

### DIFF
--- a/backendApi/src/main/java/com/virtuslab/gitmachete/backend/api/hook/IExecutionResult.java
+++ b/backendApi/src/main/java/com/virtuslab/gitmachete/backend/api/hook/IExecutionResult.java
@@ -1,9 +1,7 @@
 package com.virtuslab.gitmachete.backend.api.hook;
 
-import io.vavr.control.Option;
-
 public interface IExecutionResult {
   int getExitCode();
-  Option<String> getStdout();
-  Option<String> getStderr();
+  String getStdout();
+  String getStderr();
 }
--- a/backendApi/src/main/java/com/virtuslab/gitmachete/backend/api/hook/IExecutionResult.java
+++ b/backendApi/src/main/java/com/virtuslab/gitmachete/backend/api/hook/IExecutionResult.java
@@ -1,0 +1,9 @@
+package com.virtuslab.gitmachete.backend.api.hook;
+
+import io.vavr.control.Option;
+
+public interface IExecutionResult {
+  int getExitCode();
+  Option<String> getStdout();
+  Option<String> getStderr();
+}
--- a/backendImpl/src/main/java/com/virtuslab/gitmachete/backend/impl/hooks/ExecutionResult.java
+++ b/backendImpl/src/main/java/com/virtuslab/gitmachete/backend/impl/hooks/ExecutionResult.java
@@ -1,0 +1,25 @@
+package com.virtuslab.gitmachete.backend.impl.hooks;
+
+import io.vavr.control.Option;
+import lombok.Data;
+import lombok.RequiredArgsConstructor;
+
+import com.virtuslab.gitmachete.backend.api.hook.IExecutionResult;
+
+@Data
+@RequiredArgsConstructor(staticName = "of")
+public class ExecutionResult implements IExecutionResult {
+  private final int exitCode;
+  private final String stdout;
+  private final String stderr;
+
+  @Override
+  public Option<String> getStdout() {
+    return Option.of(stdout);
+  }
+
+  @Override
+  public Option<String> getStderr() {
+    return Option.of(stderr);
+  }
+}
--- a/backendImpl/src/main/java/com/virtuslab/gitmachete/backend/impl/hooks/ExecutionResult.java
+++ b/backendImpl/src/main/java/com/virtuslab/gitmachete/backend/impl/hooks/ExecutionResult.java
@@ -1,9 +1,7 @@
 package com.virtuslab.gitmachete.backend.impl.hooks;
 
-import io.vavr.control.Option;
 import lombok.Data;
 import lombok.RequiredArgsConstructor;
-import org.checkerframework.checker.nullness.qual.Nullable;
 
 import com.virtuslab.gitmachete.backend.api.hook.IExecutionResult;
 
@@ -11,16 +9,6 @@ import com.virtuslab.gitmachete.backend.api.hook.IExecutionResult;
 @RequiredArgsConstructor(staticName = "of")
 public class ExecutionResult implements IExecutionResult {
   private final int exitCode;
-  private final @Nullable String stdout;
-  private final @Nullable String stderr;
-
-  @Override
-  public Option<String> getStdout() {
-    return Option.of(stdout);
-  }
-
-  @Override
-  public Option<String> getStderr() {
-    return Option.of(stderr);
-  }
+  private final String stdout;
+  private final String stderr;
 }
--- a/backendImpl/src/main/java/com/virtuslab/gitmachete/backend/impl/hooks/ExecutionResult.java
+++ b/backendImpl/src/main/java/com/virtuslab/gitmachete/backend/impl/hooks/ExecutionResult.java
@@ -3,6 +3,7 @@ package com.virtuslab.gitmachete.backend.impl.hooks;
 import io.vavr.control.Option;
 import lombok.Data;
 import lombok.RequiredArgsConstructor;
+import org.checkerframework.checker.nullness.qual.Nullable;
 
 import com.virtuslab.gitmachete.backend.api.hook.IExecutionResult;
 
@@ -10,8 +11,8 @@ import com.virtuslab.gitmachete.backend.api.hook.IExecutionResult;
 @RequiredArgsConstructor(staticName = "of")
 public class ExecutionResult implements IExecutionResult {
   private final int exitCode;
-  private final String stdout;
-  private final String stderr;
+  private final @Nullable String stdout;
+  private final @Nullable String stderr;
 
   @Override
   public Option<String> getStdout() {
--- a/build.gradle
+++ b/build.gradle
@@ -174,6 +174,7 @@ subprojects {
       // See https://github.com/diffplug/spotless/blob/master/ECLIPSE_SCREENSHOTS.md on importing and exporting settings from Eclipse
       eclipse().configFile "$rootDir/config/spotless/formatting-rules.xml"
       removeUnusedImports()
+      targetExclude 'build/generated/**/*.*'
     }
   }
 
--- a/frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/OpenMacheteTabAction.java
+++ b/frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/OpenMacheteTabAction.java
@@ -1,0 +1,40 @@
+package com.virtuslab.gitmachete.frontend.actions.base;
+
+import com.intellij.openapi.actionSystem.AnActionEvent;
+import com.intellij.openapi.project.DumbAwareAction;
+import com.intellij.openapi.wm.ToolWindow;
+import com.intellij.openapi.wm.ToolWindowId;
+import com.intellij.openapi.wm.ToolWindowManager;
+import lombok.CustomLog;
+import org.checkerframework.checker.guieffect.qual.UIEffect;
+
+import com.virtuslab.gitmachete.frontend.actions.expectedkeys.IExpectsKeyProject;
+import com.virtuslab.logger.IEnhancedLambdaLogger;
+
+@CustomLog
+public class OpenMacheteTabAction extends DumbAwareAction implements IExpectsKeyProject {
+  @Override
+  public IEnhancedLambdaLogger log() {
+    return LOG;
+  }
+
+  @Override
+  @UIEffect
+  public void actionPerformed(AnActionEvent e) {
+    LOG.debug("Performing");
+
+    ToolWindowManager toolWindowManager = ToolWindowManager.getInstance(getProject(e));
+    ToolWindow toolWindow = toolWindowManager.getToolWindow(ToolWindowId.VCS);
+    if (toolWindow == null) {
+      LOG.debug(() -> "VCS tool window does not exist");
+      return;
+    }
+
+    toolWindow.activate(() -> {
+      var contentManager = toolWindow.getContentManager();
+      var tab = contentManager.findContent("Git Machete");
+
+      contentManager.setSelectedContentCB(tab, /* requestFocus */ true);
+    });
+  }
+}
--- a/frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/vcsmenu/OpenMacheteTabAction.java
+++ b/frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/vcsmenu/OpenMacheteTabAction.java
@@ -2,6 +2,8 @@ package com.virtuslab.gitmachete.frontend.actions.vcsmenu;
 
 import com.intellij.openapi.actionSystem.AnActionEvent;
 import com.intellij.openapi.project.DumbAwareAction;
+import com.intellij.openapi.project.Project;
+import com.intellij.openapi.vcs.VcsNotifier;
 import com.intellij.openapi.wm.ToolWindow;
 import com.intellij.openapi.wm.ToolWindowId;
 import com.intellij.openapi.wm.ToolWindowManager;
@@ -23,16 +25,29 @@ public class OpenMacheteTabAction extends DumbAwareAction implements IExpectsKey
   public void actionPerformed(AnActionEvent e) {
     LOG.debug("Performing");
 
+    // Getting project from event and assigning it to variable is needed to avoid exception
+    // according to sharing data context between Swing events (esp. with #2 VcsNotifier call - inside lambda)
+    Project project = getProject(e);
+
     ToolWindowManager toolWindowManager = ToolWindowManager.getInstance(getProject(e));
     ToolWindow toolWindow = toolWindowManager.getToolWindow(ToolWindowId.VCS);
     if (toolWindow == null) {
       LOG.debug("VCS tool window does not exist");
+      VcsNotifier.getInstance(getProject(e)).notifyWarning(
+          "Could not open Git Machete tab", "Ensure that Git VCS is available");
       return;
     }
 
     toolWindow.activate(() -> {
       var contentManager = toolWindow.getContentManager();
       var tab = contentManager.findContent("Git Machete");
+
+      if (tab == null) {
+        LOG.debug("Machete tab does not exist");
+        VcsNotifier.getInstance(project).notifyWarning(
+            "Could not open Git Machete tab", "Ensure that Git VCS is available");
+        return;
+      }
 
       contentManager.setSelectedContentCB(tab, /* requestFocus */ true);
     });
--- a/frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/vcsmenu/OpenMacheteTabAction.java
+++ b/frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/vcsmenu/OpenMacheteTabAction.java
@@ -26,7 +26,7 @@ public class OpenMacheteTabAction extends DumbAwareAction implements IExpectsKey
     LOG.debug("Performing");
 
     // Getting project from event and assigning it to variable is needed to avoid exception
-    // according to sharing data context between Swing events (esp. with #2 VcsNotifier call - inside lambda)
+    // because the data context is shared between Swing events (esp. with #2 VcsNotifier call - inside lambda)
     Project project = getProject(e);
 
     ToolWindowManager toolWindowManager = ToolWindowManager.getInstance(getProject(e));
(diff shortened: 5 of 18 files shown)


backendImpl/src/main/java/com/virtuslab/gitmachete/backend/impl/hooks/ExecutionResult.java
```diff
@@ -0,0 +1,25 @@
...
+@Data
+@RequiredArgsConstructor(staticName = "of")
+public class ExecutionResult implements IExecutionResult {
+  private final int exitCode;
+  private final String stdout;
+  private final String stderr;
+
+  @Override
+  public Option<String> getStdout() {
+    return Option.of(stdout);
```
PawelLipski, line 18: Hmm since `stdout` and `stderr` are `@NonNull`... do even we need to return `Option<String>` instead of just `String`?

build.gradle
```diff
@@ -200,6 +200,7 @@ subprojects {
       // See https://github.com/diffplug/spotless/blob/master/ECLIPSE_SCREENSHOTS.md on importing and exporting settings from Eclipse
       eclipse().configFile "$rootDir/config/spotless/formatting-rules.xml"
       removeUnusedImports()
+      targetExclude 'build/generated/**/*.*'
```
PawelLipski, line 203: Whoops, right ;D

src/main/resources/META-INF/plugin.xml
```diff
@@ -163,6 +163,15 @@
...
             </action>
         </group>
+
+        <action id="GitMachete.OpenMacheteTabAction"
+                class="com.virtuslab.gitmachete.frontend.actions.base.OpenMacheteTabAction"
+                icon="MacheteIcons.MACHETE_FILE">
+            <!-- In future IntelliJ releases this group-id probably be `Git.MainMenu`
+                We can consider `VcsGroups` that seems to be more stable and place this action in VCS menu -->
+            <add-to-group group-id="GitRepositoryActions" anchor="last"/>
+            <keyboard-shortcut keymap="$default" first-keystroke="control alt M" />
```
PawelLipski, line 173: Hmmm note that `Ctrl-Alt-M` corresponds to `Extract method` in the standard settings, which is a relatively common keystroke (or at least common enough to get a few prospective users mad at us)... I think `Ctrl-Alt-Shift-M` would be safer

src/main/resources/META-INF/plugin.xml
```diff
@@ -163,6 +163,15 @@
                 <override-text place="GitMacheteToolbar"/>
             </action>
         </group>
+
+        <action id="GitMachete.OpenMacheteTabAction"
+                class="com.virtuslab.gitmachete.frontend.actions.base.OpenMacheteTabAction"
+                icon="MacheteIcons.MACHETE_FILE">
+            <!-- In future IntelliJ releases this group-id probably be `Git.MainMenu`
+                We can consider `VcsGroups` that seems to be more stable and place this action in VCS menu -->
```
PawelLipski, line 171: Nit:
```suggestion
            <!-- In future IntelliJ releases this group-id will probably be `Git.MainMenu`
                 We can consider `VcsGroups` that seems to be more stable and place this action in VCS menu -->
```

src/main/resources/META-INF/plugin.xml
PawelLipski: Also... since the keymap here is `$default`... will it cover Mac OS as well? :thinking: or does Mac OS have a separate keymap?

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/OpenMacheteTabAction.java
```diff
@@ -0,0 +1,40 @@
+package com.virtuslab.gitmachete.frontend.actions.base;
```
PawelLipski, line 1: Hmm I wouldn't put this to `base` package since this class is not meant to be a shared base for other actions... maybe add a `vcsmenu` (?) package

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/OpenMacheteTabAction.java
```diff
@@ -0,0 +1,40 @@
...
+
+  @Override
+  @UIEffect
+  public void actionPerformed(AnActionEvent e) {
+    LOG.debug("Performing");
+
+    ToolWindowManager toolWindowManager = ToolWindowManager.getInstance(getProject(e));
+    ToolWindow toolWindow = toolWindowManager.getToolWindow(ToolWindowId.VCS);
+    if (toolWindow == null) {
+      LOG.debug(() -> "VCS tool window does not exist");
```
PawelLipski, line 29: Nit: no gain from lazy evaluation here (nothing needs to be computed to render the log string), can as well be without `() -> `

frontendFile/src/main/java/com/virtuslab/gitmachete/frontend/file/highlighting/MacheteAnnotator.java
```diff
@@ -66,6 +69,14 @@ private void processIndentationElement(PsiElement element, AnnotationHolder hold
 
     IndentationParameters indentationParameters = findIndentationParameters(element);
 
+    // Potentially unrelated - it's responsible to change default TAB key behavior
```
PawelLipski, line 72: Hmm just curious if this piece can be executed in a different method? Since `processIndentationElement` looks like sth that's invoked multiple times during the edit, and this piece looks like sth that just needs to be invoked once-ish... but if there's no other way, pls leave a comment about that in the code

PawelLipski: Also, `s/to change/of changing`

frontendFile/src/main/java/com/virtuslab/gitmachete/frontend/file/highlighting/MacheteAnnotator.java
```diff
@@ -66,6 +69,14 @@ private void processIndentationElement(PsiElement element, AnnotationHolder hold
...
+    // (insert real TAB or insert some number of spaces) depending on detected file indentation.
+    // In case of space insertion indent width (size) is also set
```
PawelLipski, line 74: Very unlikely case, but I think a indent width >1 is also possible in case of tab

PawelLipski: (but nobody is gonna do that in real life anyway)

frontendFile/src/main/java/com/virtuslab/gitmachete/frontend/file/highlighting/MacheteAnnotator.java
PawelLipski: Oh ok, `Maybe this is not the best place for that code but I can't find any better place :/`... yeah, so pls leave a comment in the code then

build.gradle
PawelLipski: But actually... was spotless touching build/generated/ as well?

frontendFile/src/main/java/com/virtuslab/gitmachete/frontend/file/highlighting/MacheteAnnotator.java
micpiotrowski: Hmmm, I just thinking loud, maybe open machete file action is better place... to be check if its possible

frontendFile/src/main/java/com/virtuslab/gitmachete/frontend/file/highlighting/MacheteAnnotator.java
micpiotrowski: Yes, but it seems to be not supported by IntelliJ  :/

backendImpl/src/main/java/com/virtuslab/gitmachete/backend/impl/hooks/ExecutionResult.java
micpiotrowski: Ouh, they should be `@Nullable`...

PawelLipski: Hmmm since the current impl compiled without `@Nullable`... then maybe they don't need to be?

frontendFile/src/main/java/com/virtuslab/gitmachete/frontend/file/highlighting/MacheteAnnotator.java
PawelLipski: Hmm but pls consider that the tab doesn't need to be opened via the action, it can as well be opened directly from VCS tool window

backendImpl/src/main/java/com/virtuslab/gitmachete/backend/impl/hooks/ExecutionResult.java
micpiotrowski: IIRC, theoretically if during hook execution exception is thrown and stdout and stderr are not get then they are `null` and they are passed to `ExecutionResult` as `null`

build.gradle
micpiotrowski: Unfortunately, yes

backendImpl/src/main/java/com/virtuslab/gitmachete/backend/impl/hooks/ExecutionResult.java
micpiotrowski: I marked this fields as `@Nullable`

src/main/resources/META-INF/plugin.xml
micpiotrowski: IIUC this keymap should cover all platform (but we can specify separate keymaps for `Mac OS X`, `Mac OS X 10.5+`, `Default for XWin` - X Window, `Default for KDE` and `Default for GNOME`)

micpiotrowski: Shortcut changed to `control alt shift M`

backendImpl/src/main/java/com/virtuslab/gitmachete/backend/impl/hooks/ExecutionResult.java
PawelLipski: > IIRC, theoretically if during hook execution exception is thrown and stdout and stderr are not get then they are null and they are passed to ExecutionResult as null

Nooo that would mean that there's (yet another ;) ) false negative in `NullnessChecker` :)

PawelLipski: Neither `strippedStdout` nor `strippedStderr` can be null when they're passed to `ExecutionResult.of` in `com.virtuslab.gitmachete.backend.impl.hooks.PreRebaseHookExecutor#executeHookFor`

PawelLipski: If an exception happens before they're both assigned a non-null `new String`, then this exception is re-thrown as `GitMacheteException` and `ExecutionResult` is never created in the first place

src/main/resources/META-INF/plugin.xml
PawelLipski: Ok, let's leave as is as for now, `Ctrl-Alt-Shift-M` is still a valid keystroke on Mac OS (as in, `Ctrl` is still used for many shortcuts despite `Cmd` being more prominent in general0

frontendFile/src/main/java/com/virtuslab/gitmachete/frontend/file/highlighting/MacheteAnnotator.java
micpiotrowski: Ok, so for now comment leaved

frontendFile/src/main/java/com/virtuslab/gitmachete/frontend/file/highlighting/MacheteAnnotator.java
```diff
@@ -66,6 +69,16 @@ private void processIndentationElement(PsiElement element, AnnotationHolder hold
 
     IndentationParameters indentationParameters = findIndentationParameters(element);
 
+    // Potentially unrelated - it's responsible of changing default TAB key behavior
```
PawelLipski, line 72: Btw the dedent operation (Shift+Tab) still dedents by a single space even if the indent char is set to tab... not sure if there's a separate setting that controls that?

backendImpl/src/main/java/com/virtuslab/gitmachete/backend/impl/hooks/ExecutionResult.java
micpiotrowski: Yeah, you are right. I missed this exception rethrown. Fixed

frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/GitMacheteGraphTable.java
```diff
@@ -194,9 +194,18 @@ public void refreshModel() {
...
+    var statusText = getEmptyText().setText(upperText);
+    if (lowerText != null) {
+      statusText.appendSecondaryText(lowerText, attrs,
+          /* listener */ onClickRunnable != null ? __ -> onClickRunnable.run() : null);
+    }
+  }
+
+  @UIEffect
+  private void setTextForEmptyTable(String upperText) {
+    setTextForEmptyTable(upperText, null, null);
```
PawelLipski, line 208: Name the null params

backendApi/src/main/java/com/virtuslab/gitmachete/backend/api/hook/IExecutionResult.java
```diff
@@ -0,0 +1,7 @@
+package com.virtuslab.gitmachete.backend.api.hook;
+
+public interface IExecutionResult {
```
mkondratek, line 3: Nice

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/vcsmenu/OpenMacheteTabAction.java
```diff
@@ -0,0 +1,40 @@
...
+
+  @Override
+  @UIEffect
+  public void actionPerformed(AnActionEvent e) {
+    LOG.debug("Performing");
+
+    ToolWindowManager toolWindowManager = ToolWindowManager.getInstance(getProject(e));
+    ToolWindow toolWindow = toolWindowManager.getToolWindow(ToolWindowId.VCS);
+    if (toolWindow == null) {
+      LOG.debug("VCS tool window does not exist");
```
mkondratek, line 29: What about vcs notification? This one is probably gonna happen when git is not used within project. So maybe:
`Could not open Git Machete tab. Ensure that Git VCS is enabled.`?

frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/GitMacheteGraphTable.java
```diff
@@ -194,9 +194,18 @@ public void refreshModel() {
...
 
   @UIEffect
-  private void setTextForEmptyTable(String upperText, String lowerText, @UI Runnable onClickRunnableAction) {
+  private void setTextForEmptyTable(String upperText, @Nullable String lowerText, @Nullable @UI Runnable onClickRunnable) {
     var attrs = new SimpleTextAttributes(SimpleTextAttributes.STYLE_PLAIN, JBUI.CurrentTheme.Link.linkColor());
-    getEmptyText().setText(upperText).appendSecondaryText(lowerText, attrs, /* listener */ __ -> onClickRunnableAction.run());
+    var statusText = getEmptyText().setText(upperText);
+    if (lowerText != null) {
+      statusText.appendSecondaryText(lowerText, attrs,
+          /* listener */ onClickRunnable != null ? __ -> onClickRunnable.run() : null);
```
mkondratek, line 202: Hmmm is this lambda really needed? We can't just  pass runnable here, right?

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/vcsmenu/OpenMacheteTabAction.java
micpiotrowski: It sounds reasonable, so I done this. This is how it's working:
Screenshot from 2020-07-02 15-52-51
In addition - `toolWindow` seems to be **not** null even when there is no CVS repository available in opened project but then `tab` is `null`, so I added additional check.

frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/GitMacheteGraphTable.java
micpiotrowski: No, lambda is needed because `appendSecondaryText` need `ActionListener` as third argument (not `Runnable`) but thanks to `ActionListener` is a functional interface we can simply wrap our `Runnable` with lambda and then pass to `appendSecondaryText`

frontendFile/src/main/java/com/virtuslab/gitmachete/frontend/file/highlighting/MacheteAnnotator.java
```diff
@@ -28,16 +37,30 @@ public void annotate(PsiElement element, AnnotationHolder holder) {
...
   }
 
+  @UIEffect
+  private void showCantGetBranchesMessage(PsiFile file) {
+    Editor currentEditor = FileEditorManager.getInstance(file.getProject()).getSelectedTextEditor();
+    if (currentEditor == null) {
+      return;
+    }
+    HintManager.getInstance().showErrorHint(currentEditor,
+        "We can't get project branches, so we can't checking it names", HintManager.ABOVE);
```
PawelLipski, line 47: Maybe just:
```suggestion
        "Cannot get git branches", HintManager.ABOVE);
```

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/vcsmenu/OpenMacheteTabAction.java
```diff
@@ -23,17 +25,30 @@ public IEnhancedLambdaLogger log() {
   public void actionPerformed(AnActionEvent e) {
     LOG.debug("Performing");
 
+    // Getting project from event and assigning it to variable is needed to avoid exception
+    // according to sharing data context between Swing events (esp. with #2 VcsNotifier call - inside lambda)
```
PawelLipski, line 29:
```suggestion
    // because the data context is shared between Swing events (esp. with #2 VcsNotifier call - inside lambda)
```

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/vcsmenu/OpenMacheteTabAction.java
```diff
@@ -23,17 +25,30 @@ public IEnhancedLambdaLogger log() {
   public void actionPerformed(AnActionEvent e) {
     LOG.debug("Performing");
 
+    // Getting project from event and assigning it to variable is needed to avoid exception
```
PawelLipski, line 28: Exception or rather just a race condition (possibly not causing any expections)?

micpiotrowski: Unfortunately it caused exception

PawelLipski: Ok so for clarity pls include what exception specifically it was if you remember

micpiotrowski: I was just `Throwable` with proper message

frontendFile/src/main/java/com/virtuslab/gitmachete/frontend/file/highlighting/MacheteAnnotator.java
micpiotrowski: As agreed via Slack, we leave it as is for clarity

PawelLipski: Just fix the wording/grammar:
```suggestion
        "We can't get project branches, so we can't complete their names", HintManager.ABOVE);
```

micpiotrowski: done on following PR

PR #399, opened 2020-07-02
Automatically derive pre-release suffix as the number of commits since the latest tag

PawelLipski:
(Semi-)manual control of pre-release suffixes wasn't a great idea on my part... let's see how that works :trollface:


version.gradle
```diff
@@ -1,3 +1,3 @@
 ext {
-  PLUGIN_VERSION = '0.5.2-10'
+  PROSPECTIVE_RELEASE_VERSION = '0.5.2'
```
mkondratek, line 2: :tada:

PR #400, opened 2020-07-02
Rectify test repo setup scripts + add fake commit/reflog entry date setup

PawelLipski:
:100: :100: :100: :100: 
The improved (2.15.0-compatible) implementation of discover relies on reflog timestamps, so it's necessary to spread them over time a bit (so far they were all just landing within the same second/two seconds)

### DIFF
--- a/testCommon/src/test/resources/common.sh
+++ b/testCommon/src/test/resources/common.sh
@@ -10,11 +10,26 @@ EOF
 
 # All functions defined here are guaranteed to preserve the original current working directory.
 
-newrepo() {
+function set_fake_git_date() {
+  if (( $# != 1 )); then
+    echo "${FUNCNAME[0]} <date> needs 1 parameter, $# was given"
+    exit 100
+  fi
+
+  local date=$1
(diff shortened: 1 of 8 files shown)


testCommon/src/test/resources/common.sh
```diff
@@ -50,9 +65,10 @@ cmt() {
   touch $f
   git add $f
   git commit -m "$*"
+  set_fake_git_date 2020-01-$((++commit_day_of_month))
```
PawelLipski, line 68: Reflog in machete-sandbox now looks like:
image

PawelLipski: (2020-01-01 to 2020-01-12)

micpiotrowski: Not sure but maybe `commit_day_of_month` may be a argument of `commit()` and this can be more flexible then?

PawelLipski: Hmm TBH I don't think I want it particularly flexible... as for now just a simple 1-day-per-commit timeline should be enough

PR #403, opened 2020-07-03
Tidy up GitCoreRepository

PawelLipski:


build.gradle
```diff
@@ -4,7 +4,7 @@ plugins {
 
   id 'com.github.ben-manes.versions' version '0.28.0'
 
-  id 'com.diffplug.gradle.spotless' version '4.5.0' apply false
+  id 'com.diffplug.gradle.spotless' version '4.3.0' apply false
```
mkondratek, line 7: Why is that?

gitCoreApi/src/main/java/com/virtuslab/gitcore/api/IGitCoreBranchSnapshot.java
```diff
@@ -22,25 +25,25 @@
    */
   String getFullName();
 
-  IGitCoreCommit derivePointedCommit() throws GitCoreException;
+  IGitCoreCommit getPointedCommit();
```
mkondratek, line 28: aweseme!

build.gradle
PawelLipski: Idk why exactly, but 4.5.0 sometimes seemed to ignore un-formatted files (both in `apply` and in `check`)

PawelLipski: Possibly the combination of Gradle version (6.4) and plugin version was responsible? dunno

PR #407, opened 2020-07-03
[frontendActions] Add slide in new branch (closes #230)

### LINKED ISSUE #230: Consider Slide In & Add root actions
Currently available options allow only to remove (`slide out`) a non-root branch. 

Possibilities to add (aka `slide in`) a new subbranch (for current or selected) and to add a new root branch seem to be interesting features. The point is to handle machete file edits with plugin actions. 

## Scope of the task:
- discuss details together :palms_up_together: 
- add a `slide in` action (for current & selected dropdown) :droplet: 
- add `add root` action (from toolbar) :hammer: 
- add `remove root` action (probably from selected dropdown only) :wastebasket: 

### Notes:
If branches that do not exists in the repository are not gonna be properly presented to graph table (falling to VcsNotification error) then these (`slide in`, `add root`) actions should automaticaly revert instead of causing new error.

mkondratek:
- rearrange toolbar actions
- add slide in action
- make slide out text & desc consistent with the rest of actions
- disable toggle listing commits when no commits
- use bundle for texts/descs/notifications
- add slide in dialog 

_I strongly suggest reviewing it commit by commit instead as a whole._ 

- [x] resolve bundle files ambiguousness
- [ ] ~expose `getParent` from `IBranchLayoutEntry`~ NA #430
- [x] investigate capabilities of sliding in unmanaged branches


branchLayoutApi/src/main/java/com/virtuslab/branchlayout/api/BranchLayout.java
```diff
@@ -51,4 +54,49 @@ public IBranchLayout slideOut(String branchName) throws BranchLayoutException {
       return List.of(entry.withChildren(children.flatMap(child -> slideOut(child, entryToSlideOut))));
     }
   }
+
+  @Override
+  public IBranchLayout slideIn(String aboveBranchName, String branchName) throws BranchLayoutException {
+    var entry = findEntryByName(branchName).getOrNull();
+    if (entry != null) {
```
PawelLipski, line 61: So as to not create the impression that `entry` will correspond to `entry` param of the other `slideIn` method:
```suggestion
    if (findEntryByName(branchName).isDefined()) {
```

branchLayoutApi/src/main/java/com/virtuslab/branchlayout/api/IBranchLayout.java
```diff
@@ -8,5 +8,7 @@
 
   Option<IBranchLayoutEntry> findEntryByName(String branchName);
 
+  IBranchLayout slideIn(String aboveBranchName, String branchName) throws BranchLayoutException;
```
PawelLipski, line 11: To preserve consistent naming: `s/aboveBranchName/parentBranchName` (?)

scripts/prohibit-class-name-hardcode-in-strings
```diff
@@ -7,6 +7,6 @@ source "$self_dir"/utils.sh
 
 class_names=$(git ls-files '*.java' | xargs basename -s .java | sort | paste -sd'|')
 
-if git grep -EHn  '^[^"]*"[^"$]*\<('"$class_names"')\>' -- '*.java' ':(exclude)./frontendDefs'; then
+if git grep -EHn  '^[^"]*"[^"$]*\<('"$class_names"')\>' -- '*{^Bundle}.java' ':(exclude)./frontendDefs'; then
```
PawelLipski, line 10: Is `*{^Bundle}.java` the same as `:(exclude)*Bundle.java`? :thinking:

branchLayoutApi/src/main/java/com/virtuslab/branchlayout/api/IBranchLayout.java
PawelLipski: Also, for clarity, maybe `s/branchName/newBranchName`

branchLayoutApi/src/main/java/com/virtuslab/branchlayout/api/BranchLayout.java
```diff
@@ -51,4 +54,49 @@ public IBranchLayout slideOut(String branchName) throws BranchLayoutException {
...
+    if (entry == parent) {
+      return List.of(entry.withChildren(entry.getChildren().append(entryToSlideIn)));
+    } else {
+      return List.of(entry.withChildren(children.flatMap(child -> slideIn(child, entryToSlideIn, parent))));
+    }
+  }
+
+  @AllArgsConstructor
+  @SuppressWarnings("interning:not.interned") // to allow for `==` comparison in Lombok-generated `withChildren` method
+  private static final class SlidInEntry implements IBranchLayoutEntry {
```
PawelLipski, line 87: Hmm how about `@ToString` for this one? :thinking:

PawelLipski: Since after adding `@ToString` will be basically identical to `com.virtuslab.branchlayout.impl.BranchLayoutEntry`... possibly `BranchLayoutEntry` should go into API, just as `BranchLayout` did recently

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/common/GitMacheteBundle.java
```diff
@@ -0,0 +1,27 @@
+package com.virtuslab.gitmachete.frontend.actions.common;
+
+import java.util.function.Supplier;
+
+import com.intellij.DynamicBundle;
+import org.jetbrains.annotations.PropertyKey;
+
+public final class GitMacheteBundle extends DynamicBundle {
```
PawelLipski, line 8: Hmm I don't like the `actions.messages.GitMacheteBundle` vs `actions.ResourceBundle` distinction... it's actually not clear from the names why they should be split that way (if they should be split at all). I didn't change this on #422, TBD on this PR

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/toolbar/FetchAllRemotesAction.java
```diff
@@ -30,9 +31,9 @@ public void update(AnActionEvent anActionEvent) {
     var presentation = anActionEvent.getPresentation();
     if (GitFetchSupportImpl.fetchSupport(project).isFetchRunning()) {
       presentation.setEnabled(false);
-      presentation.setDescription("Fetch is already running...");
+      presentation.setDescription(GitMacheteBundle.message("action.fetch.description.disabled.aleardy-running"));
```
PawelLipski, line 34: Typo: `aleardy`

frontendActions/src/main/resources/actions/messages/GitMacheteBundle.properties
```diff
@@ -0,0 +1,97 @@
...
+action.checkout.description.disabled.currently-checked-out=Branch ''{0}'' is currently checked out
+action.checkout.description=Checkout branch ''{0}''
+action.checkout.task.title=Checking out
+
+action.fast-forward.description=Fast forward branch ''{0}'' to match ''{1}''
+action.fast-forward.description.root.branch=Root branch ''{0}'' cannot be fast-forwarded
+action.fast-forward.task.title=Fast Forwarding...
+action.fast-forward.text.current-branch=_Fast Forward Parent To Match Current Branch
+
+action.fetch.description=Fetch all remotes
```
PawelLipski, line 21: This string (and probably a few others) are repeated b/w here and ResourceBundle.properties

frontendActions/src/main/resources/actions/messages/GitMacheteBundle.properties
```diff
@@ -0,0 +1,97 @@
+action.description.disabled.branch.status={0} disabled because the branch is {1}
```
PawelLipski, line 1: Nice <3 (commenting on the file)

frontendActions/src/main/resources/actions/messages/GitMacheteBundle.properties
```diff
@@ -0,0 +1,97 @@
...
+fetch-backgroundable.fail=Fetch of refspec {0} failed
+fetch-backgroundable.success=Fetch of refspec {0} succeeded
+
+git4idea.repository.state.detached-head=in the detached head state
+git4idea.repository.state.ongoing.cherry-pick=during an ongoing cherry-pick
+git4idea.repository.state.ongoing.merge=during an ongoing merge
+git4idea.repository.state.ongoing.rebase=during an ongoing rebase
+git4idea.repository.state.ongoing.revert=during an ongoing revert
+
+synctoparentstatus.insync=in sync with its parent
```
PawelLipski, line 85: Hmm any reason why there are no hyphens in this key (and in all keys below)?

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseSlideInNewBranchBelowAction.java
```diff
@@ -0,0 +1,136 @@
...
+
+    var branchName = getNameOfBranchUnderAction(anActionEvent);
+    var branch = branchName.flatMap(bn -> getGitMacheteBranchByName(anActionEvent, bn));
+
+    if (branch.isEmpty()) {
+      presentation.setEnabled(false);
+      presentation.setDescription(GitMacheteBundle.message("action.slide-in.description.disabled.no-parent"));
+    } else {
+      // it is supposed to be defined since "branch" is defined
+      assert branchName.isDefined() : "Branch name is undefined";
```
PawelLipski, line 58: Nit:
```suggestion
    var branchName = getNameOfBranchUnderAction(anActionEvent);

    if (branchName.isEmpty()) {
      presentation.setEnabled(false);
      presentation.setDescription(GitMacheteBundle.message("action.slide-in.description.disabled.no-parent"));
    } else {
    var branch = branchName.flatMap(bn -> getGitMacheteBranchByName(anActionEvent, bn));
```

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseSlideInNewBranchBelowAction.java
```diff
@@ -0,0 +1,136 @@
...
+  }
+
+  private static String getMessageOrEmpty(Throwable t) {
+    return t.getMessage() != null ? t.getMessage() : "";
+  }
+
+  @Nullable
+  String createOrCheckoutNewBranch(Project project, GitRepository gitRepository, String startPoint, String title) {
+    var repositories = List.of(gitRepository).asJava();
+    var options = new GitNewBranchDialog(project,
```
PawelLipski, line 117: Whoops internal API... does it seem to be stable...?

PawelLipski: Also, the problem with using a **new branch** dialog here is that we don't allow for sliding in an existing branch, which is kinda useful... although this feature can be provided on a separate PR

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseSlideOutBranchAction.java
```diff
@@ -55,13 +56,19 @@ public void update(AnActionEvent anActionEvent) {
 
     if (branch.isEmpty()) {
       presentation.setEnabled(false);
-      presentation.setDescription("Slide out disabled due to undefined branch");
+      presentation
+          .setDescription(GitMacheteBundle.message("action.description.disabled.undefined.machete.branch", "Slide out"));
```
PawelLipski, line 60: I'd say `action.description.disabled.undefined-machete-branch`

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/ISyncToRemoteStatusDependentAction.java
```diff
@@ -17,6 +20,10 @@ default String getDescriptionActionName() {
     return getActionName();
   }
 
+  default String getEnabledDescriptionBundleKey() {
```
PawelLipski, line 23: Slightly reworked on #422

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/toolbar/ToggleListingCommitsAction.java
```diff
@@ -38,16 +39,30 @@ public void update(AnActionEvent anActionEvent) {
     if (branchLayout.isDefined()) {
       boolean anyChildBranchExists = branchLayout.get().getRootEntries()
           .exists(rootBranch -> rootBranch.getChildren().nonEmpty());
+
       if (anyChildBranchExists) {
-        presentation.setEnabled(true);
-        presentation.setDescription("Toggle listing commits");
+        boolean anyCommitExists = getGitMacheteRepositorySnapshot(anActionEvent)
```
PawelLipski, line 44: Nice, I noticed that some time ago as well... not sure if I opened any issue though

frontendActions/src/main/resources/actions/messages/GitMacheteBundle.properties
```diff
@@ -0,0 +1,97 @@
...
+action.slide-in.text=Slide _In New Branch Below Current Branch
+action.slide-in.notification.fail=Slide in of new branch <b>{0}</b> failed
+
+action.slide-out.description=Slide out ''{0}''
+action.slide-out.description.root.branch=Root branch ''{0}'' cannot be slid out
+action.slide-out.notification.fail=Slide out of <b>{0}</b> failed
+action.slide-out.notification.success=Branch <b>{0}</b> slid out
+
+action.toggle-listing-commits.description=Toggle listing commits
+action.toggle-listing_commits.description.disabled.no-commits=Toggle listing commits disabled: no commits present
```
PawelLipski, line 70: Stray underscore in the key

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/toolbar/ToggleListingCommitsAction.java
```diff
@@ -38,16 +39,30 @@ public void update(AnActionEvent anActionEvent) {
...
+          presentation
+              .setDescription(GitMacheteBundle.message("action.toggle-listing_commits.description.disabled.no-commits"));
+        }
       } else {
         presentation.setEnabled(false);
-        presentation.setDescription("Toggle listing commits disabled: no child branches present");
+        presentation
+            .setDescription(GitMacheteBundle.message("action.toggle-listing-commits.description.disabled.no-child-branches"));
       }
     } else {
```
PawelLipski, line 63: Maybe rework nested if-elses into if-returns to keep each piece of error handling closer to its corresponding condition

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/contextmenu/SlideInNewBranchBelowSelectedBranchAction.java
```diff
@@ -0,0 +1,16 @@
+package com.virtuslab.gitmachete.frontend.actions.contextmenu;
+
+import com.intellij.openapi.actionSystem.AnActionEvent;
+import io.vavr.control.Option;
+
+import com.virtuslab.gitmachete.frontend.actions.base.BaseSlideInNewBranchBelowAction;
+import com.virtuslab.gitmachete.frontend.actions.expectedkeys.IExpectsKeySelectedBranchName;
+
+public class SlideInNewBranchBelowSelectedBranchAction extends BaseSlideInNewBranchBelowAction
```
PawelLipski, line 9: Yeah, as mentioned elsewhere... it should ideally be `slide in new or existing branch`

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseSlideInNewBranchBelowAction.java
PawelLipski: (or even under a seperate issue... coz we might need to postpone that into v0.7.0)

PawelLipski: (unless it's easy to add to the existing impl)

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseSlideInNewBranchBelowAction.java
```diff
@@ -0,0 +1,136 @@
...
+
+        var newBranchLayout = Try.of(() -> branchLayout.get().slideIn(gitMacheteParentBranch.get().getName(), branchName))
+            .onFailure(
+                t -> notifier.notifyError(/* title */ GitMacheteBundle.message("action.slide-in.notification.fail", branchName),
+                    getMessageOrEmpty(t)))
+            .toOption();
+
+        newBranchLayout.map(nbl -> Try.of(() -> {
+          branchLayoutWriter.write(macheteFilePath, nbl, /* backupOldLayout */ true);
+          return "ok";
```
PawelLipski, line 103: `Try.run` - and then this `return` might become unnecessary

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseResetBranchToRemoteAction.java
```diff
@@ -114,24 +116,26 @@ public void actionPerformed(AnActionEvent anActionEvent) {
...
           assert branchOption.isDefined() : "Can't get branch '${branchName}' from Git Machete repository";
-          Option<IGitMacheteRemoteBranch> remoteTrackingBranchOption = branchOption.get().getRemoteTrackingBranch();
+          var remoteTrackingBranchOption = branchOption.get().getRemoteTrackingBranch();
           if (remoteTrackingBranchOption.isDefined()) {
             resetHandler.addParameters(remoteTrackingBranchOption.get().getPointedCommit().getHash());
           } else {
             String message = "Branch '${branchName}' doesn't have remote tracking branch, so cannot be reset";
             log().warn(message);
-            VcsNotifier.getInstance(project).notifyWarning(VCS_NOTIFIER_TITLE, message);
+            VcsNotifier.getInstance(project).notifyWarning(VCS_NOTIFIER_TITLE,
```
PawelLipski, line 137: Join lines

scripts/prohibit-class-name-hardcode-in-strings
mkondratek: yup, but I couldn't combine it with `:(exclude)./frontendDefs` :confused:

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseResetBranchToRemoteAction.java
PawelLipski: Done for all on #423 xD

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseRebaseBranchOntoParentAction.java
```diff
@@ -58,52 +59,56 @@ public void update(AnActionEvent anActionEvent) {
...
 
     } else if (state.get() != Repository.State.NORMAL) {
 
       var stateName = Match(state.get()).of(
-          Case($(Repository.State.GRAFTING), "during an ongoing cherry-pick"),
-          Case($(Repository.State.DETACHED), "in the detached head state"),
-          Case($(Repository.State.MERGING), "during an ongoing merge"),
-          Case($(Repository.State.REBASING), "during an ongoing rebase"),
-          Case($(Repository.State.REVERTING), "during an ongoing revert"),
+          Case($(Repository.State.GRAFTING), GitMacheteBundle.message("git4idea.repository.state.ongoing.cherry-pick")),
```
PawelLipski, line 67: I don't think these keys should include any mention of `git4idea`, esp. for the things that aren't git4idea-specific like repository state

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseSlideInNewBranchBelowAction.java
mkondratek: whoops. probably a very accurate point about the stability :confused: 

It is likely that another checkbox `setTrackingCheckbox` will be added in the near future.
So the next custom dialog is required... :rocket:

branchLayoutApi/src/main/java/com/virtuslab/branchlayout/api/BranchLayout.java
```diff
@@ -51,4 +54,49 @@ public IBranchLayout slideOut(String branchName) throws BranchLayoutException {
...
+    }
+    var entryToSlideIn = new SlidInEntry(branchName, null, List.empty());
+    return new BranchLayout(rootEntries.flatMap(rootEntry -> slideIn(rootEntry, entryToSlideIn, parentEntry)));
+  }
+
+  @SuppressWarnings("interning:not.interned") // to allow for `entry == entryToSlideOut`
+  private List<IBranchLayoutEntry> slideIn(
+      IBranchLayoutEntry entry,
+      IBranchLayoutEntry entryToSlideIn,
+      IBranchLayoutEntry parent) {
```
PawelLipski, line 76: Hmm consider exposing `getParent` from the `BranchLayoutEntry`... just as done for `GitMacheteNonRootBranch`, or on #415 for `MyBranchLayoutEntry`

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseSlideInNewBranchBelowAction.java
mkondratek: > Also, the problem with using a **new branch** dialog here is that we don't allow for sliding in an existing branch, which is kinda useful... although this feature can be provided on a separate PR

What do we mean by `sliding in an existing branch`? What's to be existing? Name within the branch layout? A local branch absent in there?

mkondratek: Probably both cases shall be handled :thinking:

PawelLipski: I specifically meant a machete-unmanaged local branch

frontendActions/src/main/resources/actions/messages/GitMacheteBundle.properties
mkondratek: just checking your vigilance :trollface:

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseSlideInNewBranchBelowAction.java
mkondratek: nice !

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseRebaseBranchOntoParentAction.java
mkondratek: good point

src/main/resources/META-INF/plugin.xml
```diff
@@ -113,21 +119,17 @@
...
                     icon="MacheteIcons.FETCH">
                 <override-text place="GitMacheteToolbar"/>
             </action>
 
-            <action id="GitMachete.PullCurrentBranchAction"
-                    class="com.virtuslab.gitmachete.frontend.actions.toolbar.PullCurrentBranchAction"
-                    icon="MacheteIcons.PULL">
+            <separator/>
+
+            <action id="GitMachete.RebaseCurrentBranchOntoParentAction"
```
PawelLipski, line 130: Hmm now fetch and push actions are separated by rebase, which looks a bit odd:
image

PawelLipski: Actually... now that use a snapshot version built from this PR, I think this single button could change between reset/pull/push/nothing, depending on the state of the branch (but this would break our rule that we don't hide actions from the toolbar, just disable them) :thinking:

mkondratek: wow great idea :rocket: but the statement about hiding-disabling is true too :confused: 

We could implement features like this and hide them behind some registry key - feature flipper :rofl: 

Anyway, I would rather leave it for another issue & PR

mkondratek: > Hmm now fetch and push actions are separated by rebase, which looks a bit odd:
> image

it looks like they were pointing the rebase :rofl:

PawelLipski: #427 created

branchLayoutApi/src/main/java/com/virtuslab/branchlayout/api/BranchLayout.java
mkondratek: Actually I see no use for `getParent` in my case. The one passed here is the very same entry that has been found before within `IBranchLayout slideIn(String parentBranchName, String newBranchName)`.

mkondratek: issue created https://github.com/VirtusLab/git-machete-intellij-plugin/issues/430

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseSlideInNewBranchBelowAction.java
mkondratek: > Whoops internal API... does it seem to be stable...?

Tested on 202.* EAP. Works fine because the new property has the default value set. I think we can leave it like this (without custom dialog impl)

image

README.md
```diff
@@ -6,7 +6,7 @@
 
 ![](src/main/resources/META-INF/pluginIcon.svg)
 
-This is a port of [git-machete](https://github.com/VirtusLab/git-machete) into an IntelliJ plugin.
+This is a port of [git-machete](https://github.com/VirtusLab/git-machete#git-machete) into an IntelliJ plugin.
```
PawelLipski, line 9: :+1:

docs/slide-in-use-cases.md
```diff
@@ -0,0 +1,64 @@
+# Overview through Slide In use cases
+
+It turned out that the sliding-in feature is more complex than it had appeared in the first place.
+This paper is an overview through the use cases.
```
PawelLipski, line 4: `paper` is maybe too far-fetched a word for this doc ;)

docs/slide-in-use-cases.md
```diff
@@ -0,0 +1,64 @@
...
+| --- | --- | --- |
+| **Working tree changed** | no checkout | branch entry **added*** under given parent |
+| Branch contains **some commits** | no checkout | branch entry **added*** under given parent |
+| Branch contains **no commits** | checkbox dependent | branch entry **added*** under given parent |
+
+\* added in terms of [slide in logic](#slide-in---logic)
+
+## Additional notes
+
+Slide In a branch under its own entry must not be allowed.
```
PawelLipski, line 64:
```suggestion
Slide In a branch under itself must not be allowed.
```

docs/slide-in-use-cases.md
```diff
@@ -0,0 +1,64 @@
...
+
+
+**Overwriting a parent branch must not affect the branch layout**.
+
+
+The other cases are described below.
+
+| Branch State | Checkout | Branch Layout |
+| --- | --- | --- |
+| **Working tree changed** | no checkout | branch entry **added*** under given parent |
```
PawelLipski, line 56: Nit:
```suggestion
| **Working tree changed** | no checkout | branch entry **added**\* under given parent |
```
here and in the next rows

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseSlideInNewBranchBelowAction.java
```diff
@@ -81,21 +81,38 @@ public void actionPerformed(AnActionEvent anActionEvent) {
...
-    var branchName = createOrCheckoutNewBranch(project, selectedVcsRepository.get(), gitMacheteParentBranch.get().getName(),
+    var parentName = gitMacheteParentBranch.get().getName();
+    var branchName = createOrCheckoutNewBranch(project, selectedVcsRepository.get(), parentName,
         GitMacheteBundle.message("action.GitMachete.BaseSlideInBranchBelowAction.dialog.title"));
     if (branchName == null) {
       log().debug("Name of branch to slide in is null: most likely the action has been canceled from dialog");
       return;
     }
 
+    var entryAlreadyExistsBelowGivenParent = branchLayout.get().findEntryByName(parentName)
```
PawelLipski, line 101: `branchLayout.get()` appears a couple times, maybe change `branchLayout`'s declaration to `var branchLayout = getBranchLayout(anActionEvent).getOrNull()`.

PawelLipski: Also, pls add a TODO for #430 here

docs/slide-in-use-cases.md
```diff
@@ -0,0 +1,64 @@
...
+## Slide In - Logic
+
+- When the **entry does not exist**, it must be added (slid in) under the given branch.
+- When the **entry exists under the given parent**, branch layout must not be affected.
+- When the **entry exists under other parent**, it must be first removed (slid out),
+and then added (slid in) under the given parent.
+
+## Slide In - New Branches
+
+This section applies to **UC1** and **UC4**.
```
PawelLipski, line 36: Since the cases are naturally grouped by `Local branch exists` (so UC1&UC4 vs UC2&UC3) rather than by `Branch entry exists`, maybe reorder them (as, in swap the columns of the table)

docs/slide-in-use-cases.md
```diff
@@ -0,0 +1,64 @@
...
+
+
+Use case **UC1** seems to be the most common.
+They are simply added in the way specified by the [slide in logic](#slide-in---logic).
+
```
PawelLipski, line 41: Hmm a word about UC4 missing?

docs/slide-in-use-cases.md
```diff
@@ -0,0 +1,64 @@
...
+
+
+## Slide In - With "Overwrite existing branch"
+
+This section applies to **UC2** and **UC3**.
+
+
+**Overwriting a parent branch must not affect the branch layout**.
```
PawelLipski, line 49: Hmm why would we want to "overwrite the (commit pointed by?) parent branch"? If I add Y under X... why would X be overwritten in the process?

docs/slide-in-use-cases.md
PawelLipski: Also, not nit: what do you specifically mean by `Working tree changed`?

docs/slide-in-use-cases.md
```diff
@@ -0,0 +1,64 @@
...
+
+
+The other cases are described below.
+
+| Branch State | Checkout | Branch Layout |
+| --- | --- | --- |
+| **Working tree changed** | no checkout | branch entry **added*** under given parent |
+| Branch contains **some commits** | no checkout | branch entry **added*** under given parent |
```
PawelLipski, line 57: `Branch contains **some commits**`... as in, branch tip is not equal to its fork point? :thinking:

branchLayoutApi/src/main/java/com/virtuslab/branchlayout/api/BranchLayout.java
```diff
@@ -51,4 +51,30 @@ public IBranchLayout slideOut(String branchName) throws BranchLayoutException {
...
+  @Override
+  public IBranchLayout slideIn(String parentBranchName, String newBranchName) throws BranchLayoutException {
+    if (findEntryByName(newBranchName).isDefined()) {
+      throw new BranchLayoutException("Child branch entry '${newBranchName}' already exists");
+    }
+    var parentEntry = findEntryByName(parentBranchName).getOrNull();
+    if (parentEntry == null) {
+      throw new BranchLayoutException("Parent branch entry '${parentBranchName}' does not exists");
+    }
+    var entryToSlideIn = new BranchLayoutEntry(newBranchName, null, List.empty());
```
PawelLipski, line 64: Name the `null`, it's not instantly obvious

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/toolbar/FetchAllRemotesAction.java
```diff
@@ -43,7 +45,9 @@ public void actionPerformed(AnActionEvent anActionEvent) {
     var project = getProject(anActionEvent);
     var gitRepository = getSelectedGitRepository(anActionEvent);
 
-    new Task.Backgroundable(project, "Fetching...", /* canBeCancelled */ true) {
+    new Task.Backgroundable(project, GitMacheteBundle.message("action.GitMachete.FetchAllRemotesAction.task-title"), /*
```
PawelLipski, line 48: Formatting fixed on #422

docs/slide-in-use-cases.md
```diff
@@ -0,0 +1,64 @@
...
+| 1 | false | false | ~ 75% |
+| 2 | false | true | ~ 25% |
+| 3 | true | true | < 1% |
+| 4 | true | false | < 1% |
+
+## Slide In - Logic
+
+- When the **entry does not exist**, it must be added (slid in) under the given branch.
+- When the **entry exists under the given parent**, branch layout must not be affected.
+- When the **entry exists under other parent**, it must be first removed (slid out),
```
PawelLipski, line 31: Also, an uncovered case here:

image

**entry exists and is already root**, TBD what should be done (I think we could allow for attaching a root under a non-root branch, this might be useful e.g. for attaching `develop` under `master`)

PawelLipski: Also, I don't think the error message should say "...of **new** branch master" since it has already existed

PawelLipski: And also... I don't think more complex objects like `BranchLayout` should be passed to error messages like this one... I can address that on my follow-up PR

docs/slide-in-use-cases.md
PawelLipski: Does this row specifically refer to the case when this kind of message appears?

image

PawelLipski: Although I have no idea based on what criteria git4idea decided that "some commits can be lost"

PawelLipski: Steps to reproduce the above:
1. set up machete-sandbox from git-machete-blog-2
2. add `foo123` under `call-ws`
3. try sliding in under `hotfix/add-trigger`

docs/slide-in-use-cases.md
```diff
@@ -0,0 +1,64 @@
...
+and then added (slid in) under the given parent.
```
PawelLipski, line 32: In terms of the user-facing naming: I think the user-facing dialog should be titled sth like `Add new branch or reattach existing branch under <parent-branch-name>`, the current `Slide In New Branch` is somewhat confusing wrt. those two distinct behaviors

docs/slide-in-use-cases.md
PawelLipski: (Note that as a last resort there's also a nuke-it-all solution: forbid reattaching existing branches completely and just limit ourselves to adding new branches)

docs/slide-in-use-cases.md
mkondratek: There are some factors here.
- we do not know whether the inserted branch name will be the name of an existing branch or a new one
- we shall care about consistency with the existing within IntelliJ `GitNewBranchDialog`
  - the most common case is creating a new branch 
  - it provides additional options (checkout, overwrite, (and in the near future) setTracking)

The dialog title can be changed easily to whatever we want. I am opting for `Slide In Branch Below <parent-name>`. 

Note the text field label `New branch name:` which references the `new`. Sadly, this one cannot be changed without reimplementation of the dialog.

docs/slide-in-use-cases.md
mkondratek: > Does this row specifically refer to the case when this kind of message appears?
> 
> image

yup, this is the expected behavior 

This (and "Working tree changed" aka uncommitted changes) may cause failed checkout notification (but successfully slide in an existing branch entry!).

mkondratek: My (probably wrong) thinking was that we can cover all cases 
\+ more by enhancing the existing new branch dialog by performing our slide in related operations (consistency, the same checkboxes options provided)
\- than creating a custom slide in dialog (unachievable consistency, more work to provide the checkbox options)

I would like to rethink things (on Monday) and consider **nuke-it-all solution**.

mkondratek: One more thing about the present solution cause it may not have been clear. 

Sliding in of the existing branch is now possible only with **overwrite** option (without it the `Create` button is disabled for existing branches).

docs/slide-in-use-cases.md
mkondratek: Probably already answered below (added to doc too).

The overwrite option is required for existing branches. Without it the dialog's "Create" button is disabled.

docs/slide-in-use-cases.md
mkondratek: Thinking out loud.

We could create a `Slide In Dialog` that takes the name (any valid) and slides in under the specified parent.
Then, for the NON-EXISTING BRANCHES ONLY show the `Create New Branch Dialog` using the provided name as the initial field value.
- a user may change the name within the second dialog (it would be rather hard to forbid the edition)
- we omit the problematic overwriting case
- TBD what to do in case of second dialog cancelation: do we want the whole action to be atomic or not

mkondratek:
As discussed f2f...

The updated approach is to have a separate dialog for slide in. 
The new branch dialog shall appear for branches that do not exist within the local repo.

mkondratek:
image
image

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseSlideInBranchBelowAction.java
```diff
@@ -0,0 +1,219 @@
...
+    } else {
+      createNewBranch(project, repositories, startPoint, options);
+    }
+
+    return options.getName();
+  }
+
+  @AllArgsConstructor
+  @SuppressWarnings("interning:not.interned") // to allow for `==` comparison in Lombok-generated `withChildren` method
+  private static final class SlideInEntry implements IBranchLayoutEntry {
```
PawelLipski, line 204: Do we still need this one? Can we use `BranchLayoutEntry` instead?

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/dialogs/SlideInDialog.kt
```diff
@@ -0,0 +1,61 @@
...
+import com.intellij.ui.DocumentAdapter
+import com.intellij.ui.layout.CellBuilder
+import com.intellij.ui.layout.ValidationInfoBuilder
+import com.intellij.ui.layout.panel
+import com.virtuslab.gitmachete.frontend.actions.common.GitMacheteBundle.message
+import kotlin.apply
+import kotlin.text.isEmpty
+import kotlin.text.trim
+
+class SlideInDialog constructor(project: Project,
```
PawelLipski, line 15: Nit: merge lines?

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/dialogs/SlideInDialog.kt
```diff
@@ -0,0 +1,61 @@
...
+                                private var parentName: String)
+  : DialogWrapper(project, true) {
+
+  private var branchName = ""
```
PawelLipski, line 19: Nit: drop a comment that this field is only ever meant to be written on UI thread (IIUC), so no synchronization is needed

scripts/enforce-indent-two-spaces-outside-java
```diff
@@ -20,7 +20,7 @@ EOF
 )"
 
 exit_code=0
-for file in $(git ls-files ':!*.flex' ':!gradlew' ':!*.java' ':!plugin-verifier.gradle' ':!scripts/setup-multiroot-sandbox' ':!*.svg' ':!testCommon/src/test/resources/*.sh' ':!*.xml'); do
+for file in $(git ls-files ':!*.flex' ':!gradlew' ':!*.java' ':!*.kt' ':!plugin-verifier.gradle' ':!scripts/setup-multiroot-sandbox' ':!*.svg' ':!testCommon/src/test/resources/*.sh' ':!*.xml'); do
```
PawelLipski, line 23: Does spotless work on Kotlin files as well?

frontendActions/build.gradle
```diff
@@ -16,3 +16,9 @@ dependencies {
 
 addIntellijToCompileClasspath(withGit4Idea: true)
 applySubtypingChecker()
+
+apply plugin: 'kotlin'
+compileKotlin {
+  destinationDir = compileJava.destinationDir
+  kotlinOptions.jvmTarget = "1.8"
```
PawelLipski, line 23: Hmm Java 8 and not Java 11? :thinking:

frontendActions/build.gradle
```diff
@@ -16,3 +16,9 @@ dependencies {
 
 addIntellijToCompileClasspath(withGit4Idea: true)
 applySubtypingChecker()
+
+apply plugin: 'kotlin'
```
PawelLipski, line 20: `apply plugin: 'org.jetbrains.kotlin.jvm'` for consistency with declaration in top-level build.gradle

branchLayoutApi/src/main/java/com/virtuslab/branchlayout/api/IBranchLayout.java
```diff
@@ -8,5 +8,7 @@
 
   Option<IBranchLayoutEntry> findEntryByName(String branchName);
 
+  IBranchLayout slideIn(String aboveBranchName, IBranchLayoutEntry entryToSlideIn) throws BranchLayoutException;
```
PawelLipski, line 11: `s/above/parent`?

branchLayoutApi/src/main/java/com/virtuslab/branchlayout/api/BranchLayout.java
```diff
@@ -51,4 +51,59 @@ public IBranchLayout slideOut(String branchName) throws BranchLayoutException {
...
+      return removeEntry(rootEntries, branchName);
+    }
+  }
+
+  private static List<IBranchLayoutEntry> removeEntry(List<IBranchLayoutEntry> entries, String branchName) {
+    return entries.reject(e -> e.getName().equals(branchName))
+        .map(e -> e.withChildren(removeEntry(e.getChildren(), branchName)));
+  }
+
+  @SuppressWarnings("interning:not.interned") // to allow for `entry == parent`
```
PawelLipski, line 97: Looks that it's not needed anymore

frontendActions/build.gradle
PawelLipski: If Java 8 is needed indeed, pls add a comment why this discrepancy

frontendActions/build.gradle
```diff
@@ -16,3 +16,9 @@ dependencies {
...
+compileKotlin {
+  destinationDir = compileJava.destinationDir
```
PawelLipski, line 22: Why is this needed? :thinking:

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseSlideInBranchBelowAction.java
PawelLipski: Oh ok, `BranchLayoutEntry` isn't available in the API

PawelLipski: Hmm not sure if I brought this up before... but you could consider exposing `BranchLayoutEntry` in the API as well (and then here an extra class won't be needed)

docs/slide-in-use-cases.md
```diff
@@ -0,0 +1,58 @@
...
+| --- | :---: |
+| **Working tree changed** (uncommitted changes) | Fail |
+| Branch contains **some commits** | Fail |
+| Branch contains **no commits** | Success |
+
+![Checkout Failed Notification](checkout-failed.png "Checkout Failed Notification")
+
+**IMPORTANT NOTE:**
+The failure of the overwrite does not mean the failure of the slide in.
+Even though the notification like above may appear when the specified circumstances are met the branch slide will take place.
```
PawelLipski, line 58:
```suggestion
The branch will still be slid in even if the overwrite fails.
```

docs/slide-in-use-cases.md
```diff
@@ -0,0 +1,58 @@
+# Overview through Slide In use cases
+
+It turned out that the sliding-in feature is more complex than it had appeared in the first place.
+This document is an overview through the use cases.
+It should help to implement and understand the feature.
+
+## Slide In Branch Dialog
+
+The picture below shows the **Slide In Branch Dialog**.
+This is a regular **Create New Branch Dialog** with a changed title.
```
PawelLipski, line 10: I don't think this is up to date

docs/slide-in-use-cases.md
```diff
@@ -0,0 +1,58 @@
...
+## Slide In - Logic
+
+This specification covers all of the cases.
+
+- When the **entry does not exist**, it must be added under the given branch.
+- When the **entry exists under the given parent**, branch layout must not be affected.
+- When the **entry exists under other parent or is root**, it must be reattached under the given parent.
+
+**IMPORTANT NOTE:**
+Sliding in an entry under itself (directly and indirectly, as a deeper child node) must not be allowed.
```
PawelLipski, line 36: (?)
```suggestion
Sliding in an entry under itself or any of its descendants in the branch layout must not be allowed.
```

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/dialogs/SlideInDialog.kt
mkondratek: I thought about it in the first place but splitting constructor args seems to be a convention. 
I see two reasons to keep it like it is now:
- the other Kotlin constructors I saw within IntelliJ code have the split params
- (probably the key reason) the argument preceded with `var` or `val` becomes a field and splitting them increases readability

Kotlin note:
`val` - constant 
`var` - regular variable

frontendActions/build.gradle
mkondratek: Without setting this:
image

But it turned out that both `1.8` and `11` work fine so I am leaving the later (to avoid ambiguity)

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseSlideInBranchBelowAction.java
mkondratek: dropped revert XD it is in the API again

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/dialogs/SlideInDialog.kt
mkondratek: whoops, I could not force kotlin formatter to use "my" approach. finally lines merged

frontendActions/build.gradle
mkondratek: not needed

docs/slide-in-use-cases.md
mkondratek: actually this section is irrelevant now

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/dialogs/SlideInDialog.kt
PawelLipski: Seems they're not merged now ;)

build.gradle
```diff
@@ -10,7 +10,7 @@ plugins {
   id 'org.checkerframework' version '0.5.5' apply false
   id 'org.jetbrains.grammarkit' version '2020.2.1' apply false
   id 'org.jetbrains.intellij' version '0.4.21' apply false
-  id "org.jetbrains.kotlin.jvm" version "1.3.72" apply false
+  id "org.jetbrains.kotlin.jvm" version '1.3.72' apply false
```
PawelLipski, line 13: Nit: `'`

scripts/enforce-newline-at-eof
```diff
@@ -7,7 +7,7 @@ source "$self_dir"/utils.sh
 
 found=0
 # Java sources can be skipped since Spotless takes care of them.
```
PawelLipski, line 9:
```suggestion
# Java and Kotlin sources can be skipped since Spotless takes care of them.
```

frontendActions/build.gradle
```diff
@@ -17,8 +17,7 @@ dependencies {
 addIntellijToCompileClasspath(withGit4Idea: true)
 applySubtypingChecker()
 
-apply plugin: 'kotlin'
+apply plugin: 'org.jetbrains.kotlin.jvm'
 compileKotlin {
-  destinationDir = compileJava.destinationDir
-  kotlinOptions.jvmTarget = "1.8"
+  kotlinOptions.jvmTarget = "11"
```
PawelLipski, line 22:
```suggestion
  kotlinOptions.jvmTarget = sourceCompatibility.toString() // from java-library plugin
```

frontendActions/src/main/resources/GitMacheteBundle.properties
```diff
@@ -0,0 +1,222 @@
...
+action.GitMachete.BaseSlideInBranchBelowAction.description=Slide in branch below ''{0}''
+action.GitMachete.BaseSlideInBranchBelowAction.description.disabled.no-parent=Slide in disabled due to undefined parent branch
+action.GitMachete.BaseSlideInBranchBelowAction.dialog.create-new-branch.title=Create New Branch Below {0}
+action.GitMachete.BaseSlideInBranchBelowAction.dialog.slide-in.label=Branch name:
+action.GitMachete.BaseSlideInBranchBelowAction.dialog.slide-in.label.parent=Parent Branch:
+action.GitMachete.BaseSlideInBranchBelowAction.dialog.slide-in.ok-button=Slide In
+action.GitMachete.BaseSlideInBranchBelowAction.dialog.slide-in.title=Slide In Branch Below
+action.GitMachete.BaseSlideInBranchBelowAction.notification.fail=Slide in of branch <b>{0}</b> failed
+action.GitMachete.BaseSlideInBranchBelowAction.notification.fail.branch-layout-write=New branch layout write failed
+action.GitMachete.BaseSlideInBranchBelowAction.notification.message.branch-name-equals-parent=Cannot slide in the branch under itself
```
PawelLipski, line 71: I don't think `branch-name-equals-parent` and `Cannot slide in the branch under itself` refer to the same case

docs/slide-in-use-cases.md
```diff
@@ -0,0 +1,25 @@
+# Overview through Slide In use cases
```
PawelLipski, line 1: Wow, that's become pretty short finally :hushed:

docs/slide-in-use-cases.md
```diff
@@ -0,0 +1,25 @@
...
+
+It turned out that the sliding-in feature is more complex than it had appeared in the first place.
```
PawelLipski, line 3: I don't think that it can be called `complex` anymore ;D

frontendActions/src/main/resources/GitMacheteBundle.properties
```diff
@@ -0,0 +1,222 @@
...
+GitMachete.FetchBackgroundable.notification.fail=Fetch of refspec {0} failed
+GitMachete.FetchBackgroundable.notification.success=Fetch of refspec {0} succeeded
+
+
+# Base actions
+
+action.GitMachete.BaseFastForwardParentToMatchBranchAction.description=Fast forward branch ''{0}'' to match ''{1}''
+action.GitMachete.BaseFastForwardParentToMatchBranchAction.description.disabled.branch-status=Fast forward disabled because the branch is {0}
+action.GitMachete.BaseFastForwardParentToMatchBranchAction.description.disabled.root-branch=Root branch ''{0}'' cannot be fast-forwarded
+action.GitMachete.BaseFastForwardParentToMatchBranchAction.description.sync-to-parent-status.in-sync=in sync with its parent
```
PawelLipski, line 15: This one looks unused

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/dialogs/SlideInDialog.kt
```diff
@@ -0,0 +1,70 @@
...
+    setOKButtonMnemonic('I'.toInt())
+    init()
+  }
+
+  fun showAndGetBranchName() = if (showAndGet()) branchName.trim() else null
+
+  override fun createCenterPanel() =
+      panel {
+        row(
+            message(
```
PawelLipski, line 34: Nit: increase max line length? if easily doable

frontendActions/src/main/resources/GitMacheteBundle.properties
```diff
@@ -0,0 +1,222 @@
...
+
+action.GitMachete.BaseSlideInBranchBelowAction.description=Slide in branch below ''{0}''
+action.GitMachete.BaseSlideInBranchBelowAction.description.disabled.no-parent=Slide in disabled due to undefined parent branch
+action.GitMachete.BaseSlideInBranchBelowAction.dialog.create-new-branch.title=Create New Branch Below {0}
+action.GitMachete.BaseSlideInBranchBelowAction.dialog.slide-in.label=Branch name:
+action.GitMachete.BaseSlideInBranchBelowAction.dialog.slide-in.label.parent=Parent Branch:
+action.GitMachete.BaseSlideInBranchBelowAction.dialog.slide-in.ok-button=Slide In
+action.GitMachete.BaseSlideInBranchBelowAction.dialog.slide-in.title=Slide In Branch Below
+action.GitMachete.BaseSlideInBranchBelowAction.notification.fail=Slide in of branch <b>{0}</b> failed
+action.GitMachete.BaseSlideInBranchBelowAction.notification.fail.branch-layout-write=New branch layout write failed
```
PawelLipski, line 70: `Writing new branch layout failed`

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseSlideInBranchBelowAction.java
```diff
@@ -0,0 +1,198 @@
...
+    }.queue();
+  }
+
+  private static String getMessageOrEmpty(Throwable t) {
+    return t.getMessage() != null ? t.getMessage() : "";
+  }
+
+  @Nullable
+  String createOrCheckoutNewBranch(Project project, GitRepository gitRepository, String startPoint, String initialName) {
+    var repositories = List.of(gitRepository).asJava();
```
PawelLipski, line 167: Nit: `new java.util.ArrayList(gitRepository)`? (or whatever the c'tor/factory method is)

branchLayoutApi/src/main/java/com/virtuslab/branchlayout/api/BranchLayout.java
```diff
@@ -51,4 +51,58 @@ public IBranchLayout slideOut(String branchName) throws BranchLayoutException {
...
+    var parentEntry = findEntryByName(parentBranchName).getOrNull();
+    if (parentEntry == null) {
+      throw new BranchLayoutException("Parent branch entry '${parentBranchName}' does not exists");
+    }
+    var entry = findEntryByName(entryToSlideIn.getName());
+    var entryAlreadyExists = entry.isDefined();
+
+    if (entry.map(e -> isDescendant(e, parentEntry)).getOrElse(false)) {
+      throw new BranchLayoutException(
+          "Entry '${parentEntry.getName()}' is an descendant of entry '${entryToSlideIn.getName()}'");
```
PawelLipski, line 66: `a descendant`

PawelLipski: While you're here... there's also `a unambiguous` in gitCoreJGit/src/main/java/com/virtuslab/gitcore/impl/jgit/GitCoreRepository.java:311 ;D

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/dialogs/SlideInDialog.kt
mkondratek: the funny thing is that spotless does not allow that 

`not easily doable`

docs/slide-in-use-cases.md
```diff
@@ -0,0 +1,24 @@
...
+| 3 | true  | false | ~ 25% |
+| 4 | true  | true  | < 1%  |
+
+## Slide In - Logic
+
+This specification covers all of the cases.
+
+- When the **entry does not exist**, it must be added under the given branch.
+- When the **entry exists under the given parent**, branch layout must not be affected.
+- When the **entry exists under other parent or is root**, it must be reattached under the given parent.
```
micpiotrowski, line 21: I think that some kind of question dialogbox will be nice in this case - this is something that (esp. beginning users) can overlook and this potentially can destroy branches tree.

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseSlideInBranchBelowAction.java
```diff
@@ -0,0 +1,199 @@
...
+    }
+  }
+
+  @Override
+  @UIEffect
+  public void actionPerformed(AnActionEvent anActionEvent) {
+    var project = getProject(anActionEvent);
+    var selectedVcsRepository = getSelectedGitRepository(anActionEvent).getOrNull();
+    var gitMacheteParentBranch = getNameOfBranchUnderAction(anActionEvent)
+        .flatMap(bn -> getGitMacheteBranchByName(anActionEvent, bn)).getOrNull();
```
micpiotrowski, line 79: Potentially this mapping can be omitted: I know, it's for checking that selected branch is present in repo but (just thinking loud) if this branch is provided by (IIRC) `GraphTable` so this branch must exist in repo (at least in moment of repo refreshing) since it was displayed in graph, so I think we cant trust the data provided by `GraphTable`.
WDYT?

frontendActions/src/main/resources/GitMacheteBundle.properties
```diff
@@ -0,0 +1,221 @@
...
+action.GitMachete.BaseResetBranchToRemoteAction.action-name=Re_set to Remote
+action.GitMachete.BaseResetBranchToRemoteAction.description-action-name=Reset to remote
+action.GitMachete.BaseResetBranchToRemoteAction.notification.success=Branch ''{0}'' reset to remote
+action.GitMachete.BaseResetBranchToRemoteAction.notification.title=Resetting
+action.GitMachete.BaseResetBranchToRemoteAction.task-title=Resetting...
+
+
+action.GitMachete.BaseSlideInBranchBelowAction.description=Slide in branch below ''{0}''
+action.GitMachete.BaseSlideInBranchBelowAction.description.disabled.no-parent=Slide in disabled due to undefined parent branch
+action.GitMachete.BaseSlideInBranchBelowAction.dialog.create-new-branch.title=Create New Branch Below {0}
```
micpiotrowski, line 63: This `Below` is unclear to me: it can suggest that branch will actually be under selected branch in graph but will be siblings of the selected branch, not its child. So maybe something like `Create New Child Branch` would be more accurate.

mkondratek: A usual new branch dialog:
image

I would rather keep this title (actually fix it, `s/Below/From/`).
- this is not OUR dialog; let's make it's title consistent 
- we have OUR dialog now that refers to slide in and machete logic; there is no need to reflect this reference here
- the idea in the current approach is to simply delegate branch creation to `git4idea` (if needed)

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseFastForwardParentToMatchBranchAction.java
```diff
@@ -43,21 +44,26 @@ public void onUpdate(AnActionEvent anActionEvent) {
     var branchName = getNameOfBranchUnderAction(anActionEvent);
     if (branchName.isEmpty()) {
       presentation.setEnabled(false);
-      presentation.setDescription("Fast forward disabled due to undefined branch name");
+      presentation
+          .setDescription(
```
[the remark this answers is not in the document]
PawelLipski: Can be ignored on this PR, I reworked lines like this on #422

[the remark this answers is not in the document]
mkondratek: fixed on #422

frontendActions/src/main/resources/GitMacheteBundle.properties
micpiotrowski: Ouh, my fault. Not that line. I mean line 61 (action name and description)

docs/slide-in-use-cases.md
mkondratek: have you seen the slide in dialog? I think you might have omitted it.
image

It clearly indicates what is to be slide in below what. 
Even though the newbie misses the **bold** parent he/she can easily undo things within (conveniently accessible) machete file. 

Adding the next dialog `Are you sure that you are sure?` is unnecessary.

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseSlideInBranchBelowAction.java
mkondratek: It took me a while to understand your point but... XD yeah you are right

docs/slide-in-use-cases.md
mkondratek: + actually the comment about dialog name applies here

micpiotrowski: Yes, I have seen this dialog, but IMHO i someone will try slide-in branch that is present under other parent this is not something fully intuitive that this branch will be reattached. I mean (esp. in large branch tree) this is simple to accidentally omit already added branch and trying to slide-in it again. But ok, I do not insist, maybe it's just my misconception that the user should be treated as a potential idiot xD

branchLayoutApi/src/main/java/com/virtuslab/branchlayout/api/BranchLayout.java
```diff
@@ -51,4 +51,58 @@ public IBranchLayout slideOut(String branchName) throws BranchLayoutException {
       return List.of(entry.withChildren(children.flatMap(child -> slideOut(child, entryToSlideOut))));
     }
   }
+
+  @Override
+  public IBranchLayout slideIn(String parentBranchName, IBranchLayoutEntry entryToSlideIn) throws BranchLayoutException {
```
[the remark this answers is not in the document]
mkondratek: Ofc doable but what for?

The first thing is that passing `IBranchLayoutEntry` makes sense. 
This is because we slide in the whole entry (with its children entries).

Note that slide in is not "similar" to slide out (children's behavior :baby: :baby: ).

The second thing that passing `parentBranchName` is better than parent entry. 
This is because we want our entry to come from `this` branch layout (therefore we `findEntryByName`).
Passing an entry would increase complexity (shall we compare entry by reference, by name, by children?). 

I think that the present solution is justified.

docs/slide-in-use-cases.md
mkondratek: I agree with your **first experience**.

Probably our different expectations (it was @PawelLipski suggestion to reattach) should be considered... :thinking: 
My idea is to add a checkbox (may be enabled by default) for the reattach (simply slide out and slide in without children otherwise). 

This is a candidate for another issue, WDYT?

frontendActions/src/main/resources/GitMacheteBundle.properties
mkondratek: _slack discussion summary_

leaving as it is since `child` would in indicate that the branch is a child (but may be root)

docs/slide-in-use-cases.md
micpiotrowski: Yes, checkbox is definitely good idea (maybe even better than messagebox) and yes, this can be candidate for other issue

mkondratek: issue #439  created

PR #409, opened 2020-07-03
Rework test setups so that they lead to different results under 2.14.0 vs 2.15.2 `git machete discover`

PawelLipski:
We're now compatible with 2.14.0 and not with 2.15.2 (which skips some useless branches from the result) - and the purpose of this PR is to provide "red" test cases for TDD of migration to 2.15.2.

### DIFF
--- a/testCommon/src/test/resources/setup-for-diverged-and-older-than.sh
+++ b/testCommon/src/test/resources/setup-for-diverged-and-older-than.sh
@@ -31,10 +31,12 @@ create_repo machete-sandbox
     commit Call web service
     commit 1st round of fixes
     push
-  create_branch drop-constraint # not added to definition file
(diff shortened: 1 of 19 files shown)


testCommon/src/test/resources/setup-for-diverged-and-older-than.sh
```diff
@@ -31,10 +31,12 @@ create_repo machete-sandbox
     commit Call web service
     commit 1st round of fixes
     push
-  create_branch drop-constraint # not added to definition file
-    commit Drop unneeded SQL constraints
   git checkout call-ws
     commit 2nd round of fixes
+  git checkout develop
+    # call-ws is merged to develop, and would have no downstream branches in the discovered layout,
```
mkondratek, line 37: :rocket:

PR #412, opened 2020-07-03
Fix assert exception when IDE is closed

micpiotrowski:
`BaseProjectKeyAvailabilityAssuranceAction` was added. This action can be used as replacement of `DumbAwareAction`.
It provide `canBeUpdated` method that check if instance of current class is `IExpectsKeyProject` and if is, then returns if `project` key is available. If current class is not `IExpectsKeyProject` then always returns `true`.
In `IExpectsKeyProject` new method was added: `tryGetProject` returns `Option<Project>` and if `project` key is empty then returns `Option.none()`.
Also all actions that overrides `update` method was revamped to check if `canBeUpdated` if `false` and then returns.


frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseFastForwardParentToMatchBranchAction.java
```diff
@@ -35,6 +35,10 @@ public IEnhancedLambdaLogger log() {
   public void update(AnActionEvent anActionEvent) {
     super.update(anActionEvent);
 
+    if (!canBeUpdated(anActionEvent)) {
```
mkondratek, line 38: Why don't we override the `update` method within `BaseProjectKeyAvailabilityAssuranceAction` and disable the action there, on a missing project? We could depend on inheritance instead of "IF-ology". 
Note that most of the actions already have the following check:
```
if (!presentation.isEnabledAndVisible()) {
  return;
}
```

micpiotrowski: Because we must stop processing `update` method immediately not disable the action and processing rest of method anyway, so we can't do that in this way, or I missing something.
`presentation.isEnabledAndVisible` check isn't enough.

PawelLipski: Just thinking aloud... what if we made `update` final in this class and made it:
1. call `super.update` (just as now)
1. do `canBeUpdated` check (just as now)
1. can a template method to be filled up by subclasses (TBD)?

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseProjectKeyAvailabilityAssuranceAction.java
```diff
@@ -0,0 +1,18 @@
+package com.virtuslab.gitmachete.frontend.actions.base;
+
+import com.intellij.openapi.actionSystem.AnActionEvent;
+import com.intellij.openapi.project.DumbAwareAction;
+
+import com.virtuslab.gitmachete.frontend.actions.expectedkeys.IExpectsKeyProject;
+
+public abstract class BaseProjectKeyAvailabilityAssuranceAction extends DumbAwareAction {
+  protected boolean canBeUpdated(AnActionEvent anActionEvent) {
+    if (this instanceof IExpectsKeyProject) {
```
PawelLipski, line 10: Hmmm actually... do currently we have any action that would NOT implement `IExpectsKeyProject`? Possibly we could make `BaseProjectKeyAvailabilityAssuranceAction` implement `IExpectsKeyProject`?

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/toolbar/ToggleListingCommitsAction.java
```diff
@@ -62,7 +68,13 @@ public final void actionPerformed(AnActionEvent e) {
 
   @UIEffect
   public boolean isSelected(AnActionEvent anActionEvent) {
-    return getGraphTable(anActionEvent).isListingCommits();
+    var presentation = anActionEvent.getPresentation();
+    return Toggleable.isSelected(presentation);
+  }
+
+  @UIEffect
+  public boolean isSelected(Presentation presentation) {
```
PawelLipski, line 76: I think this can be private

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseFastForwardParentToMatchBranchAction.java
PawelLipski: This way there's no risk that someone forgets about calling `super.update` and/or `canBeUpdated` in their subclass `update` impl

micpiotrowski: Great idea!

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseProjectKeyAvailabilityAssuranceAction.java
micpiotrowski: Probably all actions would implement `IExpectsKeyProject` but it is unnecessary and IMO it's distinction is more clear.

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseProjectKeyAvailabilityAssuranceAction.java
```diff
@@ -0,0 +1,28 @@
...
+      if (projectOption.isEmpty()) {
+        return;
+      }
+    }
+
+    onUpdate(anActionEvent);
+  }
+
+  @UIEffect
+  protected abstract void onUpdate(AnActionEvent anActionEvent);
```
PawelLipski, line 27: Actually, to reduce the risk of someone forgetting to call `super.onUpdate`, I'd define `onUpdate` in `BaseGitMacheteRepositoryReadyAction` instead, and mark `update` as final there. This way nobody will need to remember to call `super.onUpdate` since it's gonna be always called from `BaseGitMacheteRepositoryReadyAction#update`.

PawelLipski: The only problem will be then that classes that extend `BaseProjectKeyAvailabilityAssuranceAction` but not `BaseGitMacheteRepositoryReadyAction`... but IIRC there's currently just one (?) such class anyway, so we can call `super.update` explicitly from that class.

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/toolbar/ToggleListingCommitsAction.java
PawelLipski: Hmm seems it's still public (?)

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseProjectKeyAvailabilityAssuranceAction.java
micpiotrowski: Ok, but what about actions that are not `BaseGitMacheteRepositoryReadyAction`? What about `BaseProjectKeyAvailabilityAssuranceAction`? Should only `update` method be there and then actions that derived from `BaseProjectKeyAvailabilityAssuranceAction` should `@Override` `update` method and call `super#update`?

micpiotrowski: As agreed via Slack we leave this as is for consistency: in every action we override `onUpdate` method and call `super#onUpdate`

micpiotrowski: + grep check for call `super`

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/toolbar/ToggleListingCommitsAction.java
micpiotrowski: Now it's `private` (after force-push)  ;)

micpiotrowski: Oh, I corrected first one and forgot correct the second. Done

scripts/enforce-call-super-for-overrided-onupdate
```diff
@@ -0,0 +1,27 @@
...
+  if [[ $? -ne 0 ]]
+  then
+    echo Call 'super.onUpdate' not found in "$0"
+    return 255
+  fi
+}
+
+export -f is_call_super_found
+
+git grep --files-with-matches "void onUpdate" | grep --invert-match "BaseProjectKeyAvailabilityAssuranceAction.java" | xargs -n 1 bash -c "is_call_super_found $@" 2> /dev/null
```
PawelLipski, line 22: Note that `git grep --files-with-matches "void onUpdate"` includes this very script as well, possibly an `:!scripts/` would be useful

scripts/run-pre-build-checks
```diff
@@ -6,6 +6,7 @@ set -e
 x_option=${-//[^x]/}
 
 # pass -x option if this script is run with -x option itself
+./scripts/enforce-call-super-for-overrided-onupdate
```
PawelLipski, line 9: `overridden`

scripts/enforce-call-super-for-overrided-onupdate
PawelLipski: Same for `grep --invert-match "BaseProjectKeyAvailabilityAssuranceAction.java"`, can be just replaced with `:!*/BaseProjectKeyAvailabilityAssuranceAction.java`

PawelLipski: `xargs -n 1 bash` results in ~10 bash processes getting spawned (one for each file)... possibly this can be simplified to just `xargs git grep --files-without-match 'extends BaseProjectKeyAvailabilityAssuranceAction\|super.onUpdate'` (and then the auxiliary function isn't necessary)

scripts/enforce-call-super-for-overrided-onupdate
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+
+function is_call_super_found {
+  # If class directly extends BaseProjectKeyAvailabilityAssuranceAction then call super.onUpdate is no needed
+  grep "extends BaseProjectKeyAvailabilityAssuranceAction" "$0" > /dev/null
```
PawelLipski, line 5: (regardless of the comments below) doesn't `$0` resolve to shell name?

```
$ function foo() { echo dollar0=$0; }
$ foo hello
dollar0=bash
```

scripts/enforce-call-super-for-overrided-onupdate
```diff
@@ -0,0 +1,27 @@
...
+  fi
+}
+
+export -f is_call_super_found
+
+git grep --files-with-matches "void onUpdate" | grep --invert-match "BaseProjectKeyAvailabilityAssuranceAction.java" | xargs -n 1 bash -c "is_call_super_found $@" 2> /dev/null
+
+if [[ $? -ne 0 ]]
+then
+  echo "You must call 'super.onUpdate' if you override 'onUpdate' method and this class is not directly extends 'BaseProjectKeyAvailabilityAssuranceAction'!"
```
PawelLipski, line 26: This doesn't fail with exit code != 0 and is thus easy to overlook, pls use `die` from utils.sh as in other scripts

PawelLipski: `s/is not directly extends/does not directly extend`

scripts/enforce-call-super-for-overrided-onupdate
micpiotrowski: Ok, bo how to join `':!*/BaseProjectKeyAvailabilityAssuranceAction.java'` and `':!scripts/'`?
They can't be regex IIRC

scripts/enforce-call-super-for-overrided-onupdate
micpiotrowski: In this case it wasn't (I also don't know why)

scripts/enforce-call-super-for-overrided-onupdate
PawelLipski: Hmm doesn't
```
git grep --files-with-matches 'void onUpdate' -- ':!scripts/' ':!*/BaseProjectKeyAvailabilityAssuranceAction.java'```
work?

File: scripts/enforce-call-super-for-overridden-onupdate [hunk 9]
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+set -e -o pipefail -u
+
+self_dir=$(cd "$(dirname "$0")" &>/dev/null; pwd -P)
+source "$self_dir"/utils.sh
+
+if git grep --files-with-matches "void onUpdate" ':!scripts/' | grep --invert-match "BaseProjectKeyAvailabilityAssuranceAction.java" | xargs git grep --files-without-match 'extends BaseProjectKeyAvailabilityAssuranceAction\|super.onUpdate'
```
PawelLipski, line 8: As mentioned, `git grep --files-with-matches 'void onUpdate' -- ':!scripts/' ':!*/BaseProjectKeyAvailabilityAssuranceAction.java' | xargs ...`?

PawelLipski: Also, nit: `super.onUpdate` => `super\.onUpdate`

scripts/enforce-call-super-for-overrided-onupdate
micpiotrowski: Hmm, strange: I was sure that I tested this and it didn't work! But yes, it works, thanks!

PR #414, opened 2020-07-04
Marking current branch even when repository operation is in progress

micpiotrowski:
Now current branch is marking (underline) even when repository operation is in progress (e.g. rebasing). In addition operation name is displaying before branch name.
`GitCoreRepository#deriveCurrentBranch` was revamped to handle detached head (and get correct head name).
`GitCoreRepository#deriveRepositoryState` was added.


gitCoreJGit/src/main/java/com/virtuslab/gitcore/impl/jgit/GitCoreRepository.java
```diff
@@ -414,6 +443,18 @@ public boolean isAncestor(IGitCoreCommit presumedAncestor, IGitCoreCommit presum
     });
   }
 
+  @Override
+  public GitCoreRepositoryState deriveRepositoryState() {
+    return Match(jgitRepo.getRepositoryState()).of(
```
PawelLipski, line 448: Nit: maybe spotless formatting can be somehow turned off for a code region... it'd feel much better if it was consistently sth like
```java
        Case($(isIn(...)),
            ...),
```

gitCoreJGit/src/main/java/com/virtuslab/gitcore/impl/jgit/GitCoreRepository.java
```diff
@@ -414,6 +443,18 @@ public boolean isAncestor(IGitCoreCommit presumedAncestor, IGitCoreCommit presum
...
+        Case($(isIn(RepositoryState.CHERRY_PICKING, RepositoryState.CHERRY_PICKING_RESOLVED)),
+            GitCoreRepositoryState.CHERRY_PICK),
+        Case($(isIn(RepositoryState.MERGING, RepositoryState.MERGING_RESOLVED)), GitCoreRepositoryState.MERGING),
+        Case($(isIn(RepositoryState.REBASING, RepositoryState.REBASING_INTERACTIVE, RepositoryState.REBASING_MERGE,
+            RepositoryState.REBASING_REBASING)), GitCoreRepositoryState.REBASING),
+        Case($(isIn(RepositoryState.REVERTING, RepositoryState.REVERTING_RESOLVED)), GitCoreRepositoryState.REVERTING),
+        Case($(), GitCoreRepositoryState.NORMAL));
```
PawelLipski, line 455: Hmm a case for `APPLY` and `BISECTING` is missing (TBH git-machete CLI handles the former but not the latter)

gitCoreJGit/src/main/java/com/virtuslab/gitcore/impl/jgit/GitCoreRepository.java
```diff
@@ -130,13 +137,35 @@ private ObjectId gitCoreCommitToObjectId(IGitCoreCommit commit) throws GitCoreEx
...
+
+    String currentBranchName = null;
+
     if (ref.isSymbolic()) {
-      String currentBranchName = Repository.shortenRefName(ref.getTarget().getName());
+      currentBranchName = Repository.shortenRefName(ref.getTarget().getName());
+    } else {
+      Path headNamePath = jgitRepo.getDirectory().toPath().resolve("rebase-apply").resolve("head-name");
+      if (!headNamePath.toFile().isFile()) {
+        headNamePath = jgitRepo.getDirectory().toPath().resolve("rebase-merge").resolve("head-name");
```
PawelLipski, line 152: Hmm wait but do we have a guarantee that `.git/rebase-merge/head-name` exists? What if we're just in detached HEAD state and not rebasing?

frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/cell/BranchOrCommitCellRendererComponent.java
```diff
@@ -72,7 +74,8 @@ public BranchOrCommitCellRendererComponent(
       int row,
       int column) {
 
-    this.graphTable = table;
+    assert table instanceof GitMacheteGraphTable : "Table variable is not instance of ${GitMacheteGraphTable.class.getSimpleName()}";
```
PawelLipski, line 77: Hmm that's not the cleanest approach... what about baking `ongoingRepositoryOperation` into `RepositoryGraph`, `BranchItem` or maybe even `GitMacheteBranch` instead?

PawelLipski: Yeah, possibly adding `Option<OngoingRepositoryOperation> getOngoingRepositoryOperation()` to `IGitMacheteBranch` would be pretty clear... but then likely also `isCurrentBranch` would be needed to be added to `IGitMacheteBranch` as well (and possibly `IGitMacheteRepositorySnapshot#getCurrentBranchIfManaged` could be removed)

frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/cell/BranchOrCommitCellRendererComponent.java
```diff
@@ -103,6 +106,23 @@ public BranchOrCommitCellRendererComponent(
     int textPadding = calculateTextPadding(graphTable, maxGraphNodePositionInRow);
     appendTextPadding(textPadding);
 
+    if (graphItem.isBranchItem() && graphItem.asBranchItem().isCurrentBranch()) {
+      var gitMacheteRepositorySnapshot = graphTable.getGitMacheteRepositorySnapshot();
+      if (gitMacheteRepositorySnapshot != null) {
+        var ongoingRepositoryOperation = gitMacheteRepositorySnapshot.getOngoingRepositoryOperation();
+        if (ongoingRepositoryOperation != OngoingRepositoryOperation.NO_OPERATION) {
+          var ongoingOperationName = Match(ongoingRepositoryOperation).of(
+              Case($(OngoingRepositoryOperation.CHERRY_PICK), "CHERRY-PICK"),
```
PawelLipski, line 115: `"CHERRY-PICKING"` for consistency?

frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/cell/BranchOrCommitCellRendererComponent.java
```diff
@@ -103,6 +106,23 @@ public BranchOrCommitCellRendererComponent(
...
+              Case($(OngoingRepositoryOperation.MERGING), "MERGING"),
+              Case($(OngoingRepositoryOperation.REBASING), "REBASING"),
+              Case($(OngoingRepositoryOperation.REVERTING), "REVERTING"),
+              Case($(), ""));
+          SimpleTextAttributes attributes = SimpleTextAttributes.ERROR_ATTRIBUTES;
+          append(ongoingOperationName + CELL_TEXT_FRAGMENTS_SPACING, attributes);
```
PawelLipski, line 121: Hmm do we want to append the spacing even if `ongoingOperationName` is empty? :thinking:

frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/cell/BranchOrCommitCellRendererComponent.java
micpiotrowski: But `ongoingRepositoryOperation` is something common for whole repository, so baking it into any structure representing branch is IMO not adequate.

frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/cell/BranchOrCommitCellRendererComponent.java
micpiotrowski: It's likely to be not a case - this `Match` construction cover all possible cases, so `Case($(), ""))` is only for safety and should never happen.

PawelLipski: Aaah ok, right

frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/cell/BranchOrCommitCellRendererComponent.java
PawelLipski: Hmm it could theoretically be a property of the current branch... although I do admit that it's indeed more a property of a repository :/ but still, I'm against doing `instanceof` here... how about just changing `graphTable` field type to `GitMacheteGraphTable`?

PawelLipski: Also, is current impl working with `DemoGraphTable`? :thinking:

gitCoreJGit/src/main/java/com/virtuslab/gitcore/impl/jgit/GitCoreRepository.java
```diff
@@ -130,13 +137,35 @@ private ObjectId gitCoreCommitToObjectId(IGitCoreCommit commit) throws GitCoreEx
     } catch (IOException e) {
```
mkondratek, line 137:
```suggestion
  @Override
  public Option<IGitCoreLocalBranch> deriveCurrentBranch() throws GitCoreException {
    Ref ref = Try.of(() -> jgitRepo.getRefDatabase().findRef(Constants.HEAD))
        .getOrElseThrow(e -> new GitCoreException("Cannot get current branch", e));

    if (ref == null) {
      throw new GitCoreException("Error occurred while getting current branch ref");
    }

    Option<String> currentBranchName = null;

    if (ref.isSymbolic()) {
      currentBranchName = Option.of(Repository.shortenRefName(ref.getTarget().getName()));
    } else {
      final Path headNamePath = Option.of(jgitRepo.getDirectory().toPath().resolve("rebase-apply").resolve("head-name"))
          .map(file -> file.toFile().isFile()
              ? file
              : jgitRepo.getDirectory().toPath().resolve("rebase-merge").resolve("head-name"))
          .get();

      currentBranchName = Try.of(() -> Stream.ofAll(Files.readAllLines(headNamePath)))
          .getOrElseThrow(e -> new GitCoreException("Error occurred while getting current branch ref", e))
          .headOption()
          .map(Repository::shortenRefName);
    }

    return currentBranchName.flatMap(this::deriveLocalBranchByName);
  }
```

What about vavring it a bit?

gitCoreJGit/src/main/java/com/virtuslab/gitcore/impl/jgit/GitCoreRepository.java
micpiotrowski: Good spot! Fixed.

backendApi/src/main/java/com/virtuslab/gitmachete/backend/api/OngoingRepositoryOperation.java
```diff
@@ -1,5 +1,5 @@
 package com.virtuslab.gitmachete.backend.api;
 
 public enum OngoingRepositoryOperation {
-  NO_OPERATION, CHERRY_PICK, MERGING, REBASING, REVERTING
+  NO_OPERATION, CHERRY_PICKING, MERGING, REBASING, REVERTING, APPLY, BISECTING
```
PawelLipski, line 4: `APPLYING` for consistency

scripts/prohibit-java-code-commented-out
```diff
@@ -5,6 +5,6 @@ set -e -o pipefail -u
 self_dir=$(cd "$(dirname "$0")" &>/dev/null; pwd -P)
 source "$self_dir"/utils.sh
 
-if git --no-pager grep -En '^ *// ?(@.*|.*\{|.*\);)$' -- '*.java'; then
+if git --no-pager grep -En --perl-regexp '^ *// ?(@((?!formatter).)*|.*\{|.*\);)$' -- '*.java'; then
```
PawelLipski, line 8: You can take advantage of the fact that annotations in Java are virtually always upper-case:

```suggestion
if git --no-pager grep -En '^ *// ?(@[A-Z].*|.*\{|.*\);)$' -- '*.java'; then
```

PawelLipski: (also, if you anyway stick to `-P`/`--perl-regexp`, then `-E`/`--extended-regexp` is redundant (`-P` is practically a superset of `-E`)

gitCoreJGit/src/main/java/com/virtuslab/gitcore/impl/jgit/GitCoreRepository.java
```diff
@@ -133,45 +134,42 @@ private ObjectId convertGitCoreCommitToObjectId(IGitCoreCommit commit) throws Gi
...
-      } catch (IOException e) {
-        throw new GitCoreException("Error occurred while getting current branch ref", e);
+      // To allow map to null
+      @SuppressWarnings("nullness") Option<Path> headNamePath = Option
+          .of(jgitRepo.getDirectory().toPath().resolve("rebase-apply").resolve("head-name"))
+          .map(path -> path.toFile().isFile()
+              ? path
+              : jgitRepo.getDirectory().toPath().resolve("rebase-merge").resolve("head-name"))
+          .map(path -> path.toFile().isFile() ? path : null)
+          .flatMap(Option::of);
```
PawelLipski, line 158:
```suggestion
      Option<Path> headNamePath = Stream.of("rebase-apply", "rebase-merge")
              .map(dir -> jgitRepo.getDirectory().toPath().resolve(dir).resolve("head-name"))
              .find(path -> path.toFile().isFile());
```

PawelLipski: (`import io.vavr.collection.Stream`)

gitCoreJGit/src/main/java/com/virtuslab/gitcore/impl/jgit/GitCoreRepository.java
```diff
@@ -133,45 +134,42 @@ private ObjectId convertGitCoreCommitToObjectId(IGitCoreCommit commit) throws Gi
...
-      targetBranch = deriveLocalBranchByName(currentBranchName).getOrNull();
-    } else {
-      targetBranch = null;
     }
 
+    @Nullable IGitCoreLocalBranchSnapshot targetBranch = currentBranchName
+        .map(CheckedFunction1.liftTry(this::deriveLocalBranchByName))
+        .getOrElse(Try.success(Option.none()))
+        .getOrElseThrow(e -> (GitCoreException) e)
+        .getOrNull();
```
PawelLipski, line 172: Actually... I'd argue this part is more readable when un-vavrified:

```suggestion
    IGitCoreLocalBranchSnapshot targetBranch; 
    if (currentBranchName.isDefined()) {
      targetBranch = deriveLocalBranchByName(currentBranchName.get()).getOrNull();
    } else {
      targetBranch = null;
    }
```

gitCoreJGit/src/main/java/com/virtuslab/gitcore/impl/jgit/GitCoreRepository.java
```diff
@@ -133,45 +134,42 @@ private ObjectId convertGitCoreCommitToObjectId(IGitCoreCommit commit) throws Gi
...
 
     if (ref == null) {
       throw new GitCoreException("Error occurred while getting current branch ref");
     }
+
     var reflog = deriveReflogByRefFullName(Constants.HEAD);
-    IGitCoreLocalBranchSnapshot targetBranch;
 
-    String currentBranchName = null;
+    Option<String> currentBranchName = Option.none();
```
PawelLipski, line 146: Possibly this will be even clearer with `String currentBranchName = null;` (since you can do without any `.map` or `.flatMap` on `currentBranchName` that would merit using an `Option`, just`.isDefined()` or `.get()`

frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/cell/BranchOrCommitCellRendererComponent.java
micpiotrowski: As agreed via Slack I'll try to implement this in way with common interface

scripts/prohibit-java-code-commented-out
micpiotrowski: I thin it's more clear when it is as is. It indicates what I exactly want to exclude from this check.

gitCoreJGit/src/main/java/com/virtuslab/gitcore/impl/jgit/GitCoreRepository.java
micpiotrowski: Nice! :heart:

gitCoreJGit/src/main/java/com/virtuslab/gitcore/impl/jgit/GitCoreRepository.java
micpiotrowski: ok

backendImpl/src/main/java/com/virtuslab/gitmachete/backend/impl/GitMacheteRepository.java
```diff
@@ -313,8 +318,17 @@ IGitMacheteRepositorySnapshot createGitMacheteRepository(IBranchLayout branchLay
           ? currentBranchIfManaged.getName()
           : "<none> (unmanaged branch or detached HEAD)"));
 
+      var ongoingOperation = Match(gitCoreRepository.deriveRepositoryState()).of(
+          Case($(GitCoreRepositoryState.CHERRY_PICK), OngoingRepositoryOperation.CHERRY_PICKING),
+          Case($(GitCoreRepositoryState.MERGING), OngoingRepositoryOperation.MERGING),
+          Case($(GitCoreRepositoryState.REBASING), OngoingRepositoryOperation.REBASING),
+          Case($(GitCoreRepositoryState.REVERTING), OngoingRepositoryOperation.REVERTING),
+          Case($(GitCoreRepositoryState.APPLY), OngoingRepositoryOperation.APPLY),
```
PawelLipski, line 326: `APPLYING` for all uses?

backendApi/src/main/java/com/virtuslab/gitmachete/backend/api/OngoingRepositoryOperation.java
micpiotrowski: Ok, but I'm not sure if it's correct: in `RepositoryState` from JGit it is `APPLY` when others are with `-ING` :thinking:

backendApi/src/main/java/com/virtuslab/gitmachete/backend/api/OngoingRepositoryOperation.java
```diff
@@ -0,0 +1,5 @@
+package com.virtuslab.gitmachete.backend.api;
+
+public enum OngoingRepositoryOperation {
+  NO_OPERATION, CHERRY_PICKING, MERGING, REBASING, REVERTING, APPLYING, BISECTING
```
mkondratek, line 4: It is rather a philosophical question but... how should the enums be ordered? Lexicographically or by "usage frequency"? :thinking: 

Let it be anyway

micpiotrowski: Hmm, it's a good question. Actually, in this case I'd prefer `NO_OPERATION` first as it's something special and then other actions in lexical order.
I will change this in that way.

PR #422, opened 2020-07-07
Wire up I18n Formatter Checker to verify correct bundle keys & formatting for user-facing messages

PawelLipski:
We can consider incorporating this one directly into #407.


PawelLipski:
> Unrel: It seems like GH has problem with proper commits range determination when base branch is force-pushed

@micpiotrowski yeah they would in fact need to do (at least) sth like `git merge-base --fork-point`, but it looks like they're just doing `git merge-base`

scripts/run-pre-build-checks
```diff
@@ -13,7 +13,7 @@ bash ${x_option:+-x} ./scripts/enforce-change-notes-updated-on-pr-to-master
 ./scripts/enforce-newline-at-eof
 ./scripts/enforce-settings-gradle-match-directories
 ./scripts/enforce-to-string-call-super-in-subclasses
-./scripts/enforce-unique-key-shortcuts-for-action-texts
+############## ./scripts/enforce-unique-key-shortcuts-for-action-texts
```
PawelLipski, line 16: @mkondratek since #407 most shortcuts are repeated, but maybe it's okay... TBD what should be done with that

PawelLipski: (and this script was not failing since the path to bundle was no longer correct)

mkondratek: yeah, I think we can live without this script

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseFastForwardParentToMatchBranchAction.java
```diff
@@ -77,34 +77,29 @@ public void onUpdate(AnActionEvent anActionEvent) {
...
 
       var parentName = gitMacheteNonRoot.getUpstreamBranch().getName();
-      presentation.setDescription(GitMacheteBundle
-          .message("action.GitMachete.BaseFastForwardParentToMatchBranchAction.description", parentName, branchName.get()));
+      presentation.setDescription(
+          format(getString("action.GitMachete.BaseFastForwardParentToMatchBranchAction.description"),
+              parentName, branchName.get()));
 
     } else {
+      // @formatter:off
```
mkondratek, line 89: :open_mouth:

frontendResourceBundles/src/main/resources/GitMacheteBundle.properties
```diff
@@ -3,8 +3,8 @@
 action.GitMachete.description.disabled.undefined.branch-name={0} disabled due to undefined branch name
 action.GitMachete.description.disabled.undefined.machete-branch={0} disabled due to undefined machete branch
 
-GitMachete.FetchBackgroundable.notification.fail=Fetch of refspec {0} failed
-GitMachete.FetchBackgroundable.notification.success=Fetch of refspec {0} succeeded
+action.GitMachete.FetchBackgroundable.notification.fail=Fetch of refspec {0} failed
```
mkondratek, line 6: `FetchBackgroundable` is not an action :confused:

PawelLipski: Hmmm it's true... but TBH shouldn't it then be `GitMachete.action.xxx` or `GitMachete.FetchBackgroundable.xxx`? Actually, do we even need that `GitMachete` prefix for anything? :thinking:

scripts/run-pre-build-checks
PawelLipski: Removed

frontendResourceBundles/src/main/resources/GitMacheteBundle.properties
mkondratek: We need `GitMachete` for bundles that are referenced by `plugin.xml` only

PawelLipski: _We need_ = _without that prefix, we could get a name conflict with a property of the same key defined somewhere else in the classpath_, right?

PawelLipski: Or is there any other reason?

PawelLipski: Also, replacing `action.GitMachete.xxx` with `GitMachete.action.xxx` everywhere should still be okay wrt. plugin.xml, right?

PR #443, opened 2020-07-23
[frontendActions][slide-out] Enforce refresh execution after branch deletion

mkondratek:
- execute repository update after branch deletion

### DIFF
--- a/frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseSlideOutBranchAction.java
+++ b/frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseSlideOutBranchAction.java
@@ -132,8 +132,7 @@ public abstract class BaseSlideOutBranchAction extends BaseGitMacheteRepositoryR
     new Task.Backgroundable(project, "Deleting branch if required...") {
       @Override
       public void run(ProgressIndicator indicator) {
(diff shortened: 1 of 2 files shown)


frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseSlideOutBranchAction.java
```diff
@@ -132,20 +132,16 @@ private void doSlideOut(AnActionEvent anActionEvent, IGitMacheteNonRootBranch br
...
+            () -> getGraphTable(anActionEvent).queueRepositoryUpdateAndModelRefresh());
       }
     }.queue();
   }
 
   /**
    * This method must NOT be called on UI thread.
    */
-  private void deleteBranchIfRequired(AnActionEvent anActionEvent, String branchName) {
+  private void deleteBranchIfRequired(AnActionEvent anActionEvent, String branchName, Runnable andThen) {
```
PawelLipski, line 144: Document (javadoc) that `andThen` will NOT be executed on UI thread (IIUC)

mkondratek: Actually we can just depend on the topic `GitRepository.GIT_REPO_CHANGE` in case of branch deletion

PR #458, opened 2020-07-29
Prevent against resetting to an outdated commit of remote tracking branch

PawelLipski:
This was relatively hard to come across since the bug only manifested itself when a `git fetch` happened between the latest graph table refresh and running the reset action.


backendImpl/src/main/java/com/virtuslab/gitmachete/backend/impl/GitMacheteRepository.java
```diff
@@ -76,7 +76,7 @@ public IGitMacheteRepositorySnapshot createSnapshotForLayout(IBranchLayout branc
     LOG.startTimer().debug("Entering");
     try {
       var aux = new CreateGitMacheteRepositoryAux(gitCoreRepository, statusHookExecutor, preRebaseHookExecutor);
-      var result = aux.createGitMacheteRepository(branchLayout);
+      var result = aux.createSnapshot(branchLayout);
```
PawelLipski, line 79: And also some unrelated naming fixes.

PR #473, opened 2020-07-31
Bump dependencies + optimize awk invocations + apply other misc fixes

PawelLipski:


scripts/enforce-indent-two-spaces-outside-java
```diff
@@ -2,13 +2,12 @@
 
 set -e -o pipefail -u
 
-self_dir=$(cd "$(dirname "$0")" &>/dev/null; pwd -P)
-source "$self_dir"/utils.sh
-
 awk_script="$(cat <<'EOF'
 
+  FNR == 1 { prev_non_empty = "" }
```
PawelLipski, line 7: `FNR` = number of line in the current file
`NR` = number of line, overall

PR #475, opened 2020-08-01
Enforce property keys contain valid class names

PawelLipski:


frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/toolbar/SlideInBranchBelowCurrentBranchAction.java
```diff
@@ -5,7 +5,7 @@
 
 import com.virtuslab.gitmachete.frontend.actions.base.BaseSlideInBranchBelowAction;
 
-public class SlideInBranchBelowCurrentAction extends BaseSlideInBranchBelowAction {
+public class SlideInBranchBelowCurrentBranchAction extends BaseSlideInBranchBelowAction {
```
micpiotrowski, line 8: While we are here: many of `toolbar` actions (`FastForwardParentToMatchCurrentBranchAction`, `PullCurrentBranchFastForwardOnlyAction`, `ResetCurrentBranchToRemoteAction`, `SlideInBranchBelowCurrentBranchAction`, `SlideOutCurrentBranchAction`) seems to be unused, so probably should be deleted.

micpiotrowski: But ok, maybe it's more reasonable to delete (or not) this actions on PR #484

PawelLipski: Yeah, I've already brought this up with @mkondratek, let's see which of the current-branch actions actually survive #484

PR #452, opened 2020-07-28
Readme revamp

micpiotrowski:

PR #476, opened 2020-08-03
Regenerate sample workflow GIF

micpiotrowski:


mkondratek:
IMHO multiple task-specific gifs would be better than one containing "everything" but let it be for now

PawelLipski:
Yeah, we might want to ditch/shorten this one once #452 is done

PR #482, opened 2020-08-04
Reflect gray/green difference in light theme

mkondratek:
| - | old | new |
| - | - | - |
| **dark** | Screenshot from 2020-08-04 12-07-29 | image |
| **light** | image | image |

### DIFF
--- a/testCommon/src/test/resources/common.sh
+++ b/testCommon/src/test/resources/common.sh
@@ -10,26 +10,11 @@ EOF
 
 # All functions defined here are guaranteed to preserve the original current working directory.
 
-function set_fake_git_date() {
-  if (( $# != 1 )); then
-    echo "${FUNCNAME[0]} <date> needs 1 parameter, $# was given"
-    exit 100
-  fi
-
-  local date=$1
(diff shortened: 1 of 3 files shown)


testCommon/src/test/resources/common.sh
```diff
@@ -10,27 +10,12 @@ EOF
 
 # All functions defined here are guaranteed to preserve the original current working directory.
 
-function set_fake_git_date() {
```
PawelLipski, line 13: Whoops... integration test for discover rely on this functionality :/

testCommon/src/test/resources/setup-for-diverged-and-older-than.sh
```diff
@@ -60,6 +60,8 @@ create_repo machete-sandbox
   sed 's/^  //' <<< "$machete_file" > .git/machete
 )
 
+sleep 1
```
PawelLipski, line 63: Hmm how about using `set_fake_git_date` instead to achieve the same effect in this script?

PawelLipski:
@mkondratek pls remove the setup script changes from this PR, I'm fix that on a separate PR

PR #487, opened 2020-08-04
Make toggle listing commits disabled when no non-root branch has any commits, not just children of roots

PawelLipski:


frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/toolbar/ToggleListingCommitsAction.java
```diff
@@ -56,9 +56,8 @@ public void onUpdate(AnActionEvent anActionEvent) {
     }
 
     boolean anyCommitExists = getGitMacheteRepositorySnapshot(anActionEvent)
-        .map(repo -> repo.getRootBranches()
-            .flatMap(root -> root.getChildBranches())
-            .exists(b -> b.getCommits().nonEmpty()))
+        .map(repo -> repo.getManagedBranches()
```
PawelLipski, line 59: Here's the actual change, the rest are unrelated piggy-backed fixes :sweat_smile:

PR #495, opened 2020-08-06
Bugfix/slide in with existing remote

mkondratek:
- FIX: When origin/Y exists but not Y, sliding in Y under X creates Y of X's commit and not of origin/Y's commit
- awaiting local branch with exponential timeout

---

presently experiencing an issue with 
`4:52 PM	The following branches defined by machete file do not belong to the local repository:: build-chain`
appearing on the slide in with checkout & tracking branch setting

Fixed with an exponential timeout. 

---

additional problem - compatibility problem appeared 
`shouldSetTracking` (from `GitNewBranchDialog`) does not exist yet in 2020.1
 assuming the following:
- remote exists <=> set tracking 
- do not show the should set tracking checkbox

Issue #496 opened.


frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseSlideInBranchBelowAction.java
```diff
@@ -111,16 +120,18 @@ public void actionPerformed(AnActionEvent anActionEvent) {
...
     if (localBranch == null) {
-      var createNewBranchDialogBranchName = createOrCheckoutNewBranch(project, selectedVcsRepository, parentName,
-          slideInOptions.getName());
-      if (!slideInOptions.getName().equals(createNewBranchDialogBranchName)) {
-        createNewBranchDialogBranchName = createNewBranchDialogBranchName != null
-            ? createNewBranchDialogBranchName
+      Tuple2<@Nullable String, Runnable> branchNameAndPreSlideInRunnable = getBranchNameAndPreSlideInRunnable(project,
+          selectedVcsRepository, parentName, slideInOptions.getName());
+      preSlideInRunnable = branchNameAndPreSlideInRunnable._2();
+      if (!slideInOptions.getName().equals(branchNameAndPreSlideInRunnable._1())) {
```
PawelLipski, line 128: Given that `_1` is referenced 3 times... extract a local var

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseSlideInBranchBelowAction.java
```diff
@@ -138,10 +149,15 @@ public void actionPerformed(AnActionEvent anActionEvent) {
       return;
     }
 
+    final var finalPreSlideInRunnable = preSlideInRunnable;
     new Task.Backgroundable(project, getString("action.GitMachete.BaseSlideInBranchBelowAction.task-title")) {
```
PawelLipski, line 153: Extract to a separate file, this `actionPerformed` has become huuuge

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseSlideInBranchBelowAction.java
```diff
@@ -199,6 +215,34 @@ public void run(ProgressIndicator indicator) {
...
+      private void waitForLocalBranch() {
+        Supplier<@Nullable GitLocalBranch> findLocalBranch = () -> selectedVcsRepository.getBranches()
+            .findLocalBranch(slideInOptions.getName());
+
+        try {
+          //  6 attempts, usually 3 are enough
+          final int TIMEOUT = 2048;
+          long WAIT_DURATION = 64;
+          while (findLocalBranch.get() == null && WAIT_DURATION <= TIMEOUT) {
+            synchronized (this) {
```
PawelLipski, line 228: Hmm why specifically `synchronized` around `wait`? As in, what's specifically the shared resource you're trying to protect from concurrent access here?

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseSlideInBranchBelowAction.java
```diff
@@ -230,27 +277,58 @@ String createOrCheckoutNewBranch(Project project, GitRepository gitRepository, S
...
-      createNewBranch(project, repositories, startPoint, options);
+    Runnable preSlideInRunnable = () -> {};
+    var remoteBranch = getGitRemoteBranch(project, gitRepository, branchName);
+
+    if (options.shouldCheckout() && remoteBranch != null) {
+      preSlideInRunnable = () -> checkoutRemoteBranch(project, repositories, remoteBranch.getName());
+
+    } else if (!options.shouldCheckout() && remoteBranch != null) {
+      var refspec = "refs/remotes/${remoteBranch.getName()}:refs/heads/${branchName}";
+      preSlideInRunnable = () -> new FetchBackgroundable(project, gitRepository, GitRemote.DOT.getName(), refspec,
```
PawelLipski, line 303: Use the constant from `FetchBackgroundable`

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseSlideInBranchBelowAction.java
```diff
@@ -230,27 +277,58 @@ String createOrCheckoutNewBranch(Project project, GitRepository gitRepository, S
...
+      assert remote != null : "remote is null";
+
+      if (remotes.size() > 1) {
+        VcsNotifier.getInstance(project).notifyInfo(format(
+            getString("action.GitMachete.BaseSlideInBranchBelowAction.notification.message.multiple-remotes"),
+            remote.getName()));
+      }
+
+      var remoteBranchName = "${remote.getName()}/${branchName}";
+      return gitRepository.getBranches().findRemoteBranch(remoteBranchName);
```
PawelLipski, line 330: Hmm I don't think that's the most convenient logic for the user: I'd rather only take into account the remotes that have a matching branch, and from them pick the first one in case there's more than one

PawelLipski: (or actually, from the remote repos that have a matching branch, I'd prefer `origin` first if applicable)

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseSlideInBranchBelowAction.java
```diff
@@ -230,27 +277,58 @@ String createOrCheckoutNewBranch(Project project, GitRepository gitRepository, S
...
 
-    return options.getName();
+    return Tuple.of(options.getName(), preSlideInRunnable);
+  }
+
+  @Nullable
+  private static GitRemoteBranch getGitRemoteBranch(Project project, GitRepository gitRepository, String branchName) {
+    var remotes = gitRepository.getRemotes();
+    if (remotes.size() > 0) {
+      var remote = Option.ofOptional(remotes.stream().findFirst()).getOrNull();
```
PawelLipski, line 320: Nit:
```suggestion
      var remote = remotes.stream().findFirst().orElse(null);
```

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseSlideInBranchBelowAction.java
mkondratek: `wait()` must be called within a `synchronized` block.
Otherwise, we will get an `IllegalMonitorStateException`.

Ref:
https://stackoverflow.com/questions/2779484/why-must-wait-always-be-in-synchronized-block

PawelLipski:
@mkondratek pls take a look at the following repo: [scalapy-tensorflow.zip](https://github.com/VirtusLab/git-machete-intellij-plugin/files/5051343/scalapy-tensorflow.zip) :thinking: 

I've checked out develop and clicked `Slide in below current`; then inserted `separate-examples` as branch name.
Unfortunately, the new branch has been checked out from `develop`, not `origin/separate-examples`:
image

Plugin version `0.7.0-49-SNAPSHOT+git.45b6df3`

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/backgroundables/SlideInBackgroundable.java
```diff
@@ -0,0 +1,166 @@
...
+
+    try {
+      //  6 attempts, usually 3 are enough
+      final int TIMEOUT = 2048;
+      long WAIT_DURATION = 64;
+      while (findLocalBranch.get() == null && WAIT_DURATION <= TIMEOUT) {
+        synchronized (this) {
+          wait(WAIT_DURATION);
+        }
+        WAIT_DURATION *= 2;
```
micpiotrowski, line 138: What are the benefits of exponential waiting time here?
Couldn't it be just const time - then it can be simplified (e.g. using Java's [Timer](https://docs.oracle.com/javase/7/docs/api/java/util/Timer.html) class)

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/backgroundables/SlideInBackgroundable.java
```diff
@@ -0,0 +1,166 @@
...
+    this.parentName = parentName;
+    this.notifier = VcsNotifier.getInstance(project);
+    this.branchLayoutWriter = branchLayoutWriter;
+  }
+
+  @Override
+  public void run(ProgressIndicator indicator) {
+    preSlideInRunnable.run();
+
+    waitForLocalBranch();
```
micpiotrowski, line 66: IIUC in `preSlideInRunnable` only `FetchBackgroundable` is nonblockin operation.
So, if we create something like `BaseFetch` or just `Fetch` that will **not** be `Task.Backgroundable` (so be blocking fetch operation) then we can probably get ride of `waitForLocalBranch` (maybe use only some kind of simple check if branch exists)

frontendResourceBundles/src/main/resources/GitMacheteBundle.properties
```diff
@@ -95,6 +95,9 @@ action.GitMachete.BaseSlideInBranchBelowAction.notification.title.slide-in-fail=
 action.GitMachete.BaseSlideInBranchBelowAction.notification.message.entry-does-not-exist=Branch <b>{0}</b> does not exist
 action.GitMachete.BaseSlideInBranchBelowAction.notification.message.entry-is-descendant-of=Branch <b>{0}</b> cannot be slid under branch <b>{1}</b> since <b>{0}</b> is already an ancestor of <b>{1}</b>
 action.GitMachete.BaseSlideInBranchBelowAction.notification.message.entry-is-root=Root branch <b>{0}</b> cannot be slid in
+action.GitMachete.BaseSlideInBranchBelowAction.notification.message.multiple-remotes=More than one remote found. Checking out from the first remote ({0}).
```
micpiotrowski, line 98:
```suggestion
action.GitMachete.BaseSlideInBranchBelowAction.notification.message.multiple-remotes=More than one remote with selected branch found. Checking out from the first remote ({0}).
```

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/backgroundables/SlideInBackgroundable.java
```diff
@@ -0,0 +1,166 @@
...
+  public SlideInBackgroundable(
+      Project project,
+      GitRepository gitRepository,
+      IBranchLayout branchLayout,
+      IBranchLayoutWriter branchLayoutWriter,
+      Runnable preSlideInRunnable,
+      SlideInOptions slideInOptions,
+      String parentName) {
+    super(project, getString("action.GitMachete.BaseSlideInBranchBelowAction.task-title"));
+    this.preSlideInRunnable = preSlideInRunnable;
```
PawelLipski, line 53: Nit: keep field decl & c'tor param decl & param->field assignment order consistent

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/backgroundables/SlideInBackgroundable.java
PawelLipski: Hmm @mkondratek  this indicates that an extra docs is needed why we actually need to wait here... and which of the git4idea operations are *sneakily-asynchronous* (i.e. asynchronous without clearly indicating that fact by e.g. returning a `Future`) :/

PawelLipski: I didn't think of that but it was only b/c Mikołaj told me in person; so your doubts @micpiotrowski are justified: even though there are more non-blocking/async operations (like `checkoutRemoteBranch` IIRC), it's not clear from just looking at the `BaseSlideInBranchBelowAction`/`SlideInBackgroundable` code

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/backgroundables/SlideInBackgroundable.java
```diff
@@ -0,0 +1,166 @@
...
+    Supplier<@Nullable GitLocalBranch> findLocalBranch = () -> gitRepository.getBranches()
+        .findLocalBranch(slideInOptions.getName());
+
+    try {
+      //  6 attempts, usually 3 are enough
+      final int TIMEOUT = 2048;
+      long WAIT_DURATION = 64;
+      while (findLocalBranch.get() == null && WAIT_DURATION <= TIMEOUT) {
+        synchronized (this) {
+          wait(WAIT_DURATION);
```
PawelLipski, line 136: `wait` without a corresponding `notify`? :thinking: maybe just `Thread.sleep` is enough

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/backgroundables/SlideInBackgroundable.java
PawelLipski: Yeah, can be both const and exponential here, difference is gonna be rather minuscule anyway ;)

PawelLipski: Hm `Timer` would introduce yet another thread on which the stuff would execute... which I'd rather avoid here unless absolutely needed :/

PawelLipski: (since each `Timer` has an associated thread... also, I'd rather avoid using non-IDE-managed threads - so in practice, only UI thread or background tasks look like a viable option)

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseSlideInBranchBelowAction.java
```diff
@@ -230,27 +157,66 @@ String createOrCheckoutNewBranch(Project project, GitRepository gitRepository, S
...
+            getString("action.GitMachete.BaseSlideInBranchBelowAction.notification.message.multiple-remotes"),
+            chosen._1().getName()));
+
+        // guaranteed by filter above
+        var gitRemoteBranchOption = chosen._2();
+        assert gitRemoteBranchOption.isDefined() : "remote branch is undefined";
+        return gitRemoteBranchOption.get();
+      }
+    }
+    return null;
```
PawelLipski, line 220:
```suggestion
private static @Nullable GitRemoteBranch getGitRemoteBranch(Project project, GitRepository gitRepository, String branchName) {
    var remotes = gitRepository.getRemotes();
    if (remotes.isEmpty()) {
      return null;
    }
    var remotesWithBranch = List.ofAll(remotes).flatMap(r -> {
      var remoteBranchName = "${r.getName()}/${branchName}";
      var remoteBranch = gitRepository.getBranches().findRemoteBranch(remoteBranchName);
      return remoteBranch != null ? Option.some(Tuple.of(r, remoteBranch)) : Option.none();
    }).sortBy(t -> !t._1().getName().equals("origin"));

    var chosen = remotesWithBranch.head();
    if (remotesWithBranch.size() > 1) {
      VcsNotifier.getInstance(project).notifyInfo(format(
              getString("action.GitMachete.BaseSlideInBranchBelowAction.notification.message.multiple-remotes"),
              chosen._1().getName()));
    }
    return chosen._2();
  }
```

PawelLipski: ... or sth like that :)

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseSlideInBranchBelowAction.java
```diff
@@ -85,13 +76,13 @@ public void update(AnActionEvent anActionEvent) {
   @UIEffect
   public void actionPerformed(AnActionEvent anActionEvent) {
     var project = getProject(anActionEvent);
-    var selectedVcsRepository = getSelectedGitRepository(anActionEvent).getOrNull();
```
PawelLipski, line 88: While you're here... pls rename of couple other remaining `selectedVcsRepository` occurrences across the project

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/backgroundables/SlideInBackgroundable.java
```diff
@@ -0,0 +1,166 @@
...
+    } else {
+      entryToSlideIn = new BranchLayoutEntry(slideInOptions.getName(), /* customAnnotation */ null,
+          /* children */ List.empty());
+      targetBranchLayout = branchLayout;
+    }
+
+    var newBranchLayout = Try
+        .of(() -> targetBranchLayout.slideIn(parentName, entryToSlideIn))
+        .onFailure(e -> Match(e).of(
+            Case($(instanceOf(EntryDoesNotExistException.class)), exceptionWithMessageHandler(
```
PawelLipski, line 107: Maybe use `onFailure(Class<X> exceptionType, Consumer<? super X> action)` instead to simply things a bit

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/backgroundables/SlideInBackgroundable.java
```diff
@@ -0,0 +1,166 @@
...
+    IBranchLayout targetBranchLayout;
+    if (childEntryByName.isDefined()) {
+
+      if (slideInOptions.shouldReattach()) {
+        entryToSlideIn = childEntryByName.get();
+        targetBranchLayout = branchLayout;
+      } else {
+        entryToSlideIn = childEntryByName.map(e -> e.withChildren(List.empty())).getOrNull();
+        targetBranchLayout = Try.of(() -> branchLayout.slideOut(slideInOptions.getName()))
+            .onFailure(e -> Match(e).of(
```
PawelLipski, line 81: See below, maybe `onFailure(Class<X> exceptionType, Consumer<? super X> action)`

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseSlideInBranchBelowAction.java
PawelLipski: See my recent comments on `wait`

frontendResourceBundles/src/main/resources/GitMacheteBundle.properties
```diff
@@ -95,6 +95,9 @@ action.GitMachete.BaseSlideInBranchBelowAction.notification.title.slide-in-fail=
 action.GitMachete.BaseSlideInBranchBelowAction.notification.message.entry-does-not-exist=Branch <b>{0}</b> does not exist
 action.GitMachete.BaseSlideInBranchBelowAction.notification.message.entry-is-descendant-of=Branch <b>{0}</b> cannot be slid under branch <b>{1}</b> since <b>{0}</b> is already an ancestor of <b>{1}</b>
 action.GitMachete.BaseSlideInBranchBelowAction.notification.message.entry-is-root=Root branch <b>{0}</b> cannot be slid in
+action.GitMachete.BaseSlideInBranchBelowAction.notification.message.multiple-remotes=More than one remote with selected branch found. Checking out from the first remote ({0}).
```
PawelLipski, line 98:
```suggestion
action.GitMachete.BaseSlideInBranchBelowAction.notification.message.multiple-remotes=More than one remote with a branch named ''<branch-name>'' found. Checking out from the first remote ({0}).
```

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseSlideInBranchBelowAction.java
```diff
@@ -230,27 +157,62 @@ String createOrCheckoutNewBranch(Project project, GitRepository gitRepository, S
...
+      return null;
+    }
+
+    var remotesWithBranch = List.ofAll(remotes).flatMap(r -> {
+      var remoteBranchName = "${r.getName()}/${branchName}";
+      var remoteBranch = gitRepository.getBranches().findRemoteBranch(remoteBranchName);
+      return remoteBranch != null ? Option.some(Tuple.of(r, remoteBranch)) : Option.none();
+    }).sortBy(t -> !t._1().getName().equals("origin"));
+
+    assert remotesWithBranch.nonEmpty() : "remotes list is empty";
```
PawelLipski, line 209: Erm... where does this guarantee come from? :thinking:

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseSlideInBranchBelowAction.java
```diff
@@ -230,27 +157,62 @@ String createOrCheckoutNewBranch(Project project, GitRepository gitRepository, S
...
+    } else if (!options.shouldCheckout() && remoteBranch == null) {
+      preSlideInRunnable = () -> createNewBranch(project, repositories, startPoint, options);
     }
 
-    return options.getName();
+    return Tuple.of(options.getName(), preSlideInRunnable);
+  }
+
+  @Nullable
+  private static GitRemoteBranch getGitRemoteBranch(Project project, GitRepository gitRepository, String branchName) {
```
PawelLipski, line 197:
```suggestion
  private static @Nullable GitRemoteBranch getGitRemoteBranch(Project project, GitRepository gitRepository, String branchName) {
```

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseSlideInBranchBelowAction.java
```diff
@@ -230,27 +157,62 @@ String createOrCheckoutNewBranch(Project project, GitRepository gitRepository, S
...
+    var remotes = gitRepository.getRemotes();
+    if (remotes.isEmpty()) {
+      return null;
+    }
+
+    var remotesWithBranch = List.ofAll(remotes).flatMap(r -> {
+      var remoteBranchName = "${r.getName()}/${branchName}";
+      var remoteBranch = gitRepository.getBranches().findRemoteBranch(remoteBranchName);
+      return remoteBranch != null ? Option.some(Tuple.of(r, remoteBranch)) : Option.none();
+    }).sortBy(t -> !t._1().getName().equals("origin"));
```
PawelLipski, line 207: Hmm is `false < true` or `true < false`?...

PawelLipski: Pls drop a comment

PawelLipski: Ok, I see now it must be the former... anyway, that's a bit non-trivial at first glance

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseSlideInBranchBelowAction.java
PawelLipski: What if there are some remotes but none that would contain a branch of the given name?

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/backgroundables/SlideInBackgroundable.java
```diff
@@ -0,0 +1,158 @@
...
+    this.notifier = VcsNotifier.getInstance(project);
+  }
+
+  @Override
+  public void run(ProgressIndicator indicator) {
+    preSlideInRunnable.run();
+
+    // `preSlideInRunnable` may perform some sneakily-asynchronous operations (e.g. checkoutRemoteBranch).
+    // The high-level method used within the runnable do not allow us to schedule the tasks after them.
+    // (Stepping deeper is not an option since we loose some important logic or make us very dependent on git4idea.)
```
PawelLipski, line 64:
```suggestion
    // (Stepping deeper is not an option since we would lose some important logic or become very dependent on the internals of git4idea.)
```

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/backgroundables/SlideInBackgroundable.java
```diff
@@ -0,0 +1,158 @@
...
+    // Hence we await for the creation of the branch (with exponential backoff).
```
PawelLipski, line 65: `await + object` or `wait for + object`

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/backgroundables/SlideInBackgroundable.java
```diff
@@ -0,0 +1,158 @@
...
+      if (slideInOptions.shouldReattach()) {
+        entryToSlideIn = entryByName;
+        targetBranchLayout = branchLayout;
+      } else {
+        entryToSlideIn = entryByName.withChildren(List.empty());
+        targetBranchLayout = Try.of(() -> branchLayout.slideOut(slideInOptions.getName()))
+            .onFailure(EntryDoesNotExistException.class, exceptionWithMessageHandler(
+                format(getString("action.GitMachete.BaseSlideInBranchBelowAction.notification.message.entry-does-not-exist"),
+                    entryToSlideIn.getName()))::apply)
+            .onFailure(EntryIsRootException.class, exceptionWithMessageHandler(
```
PawelLipski, line 85: Maybe `exceptionWithMessageHandler` could return a ready `Consumer`? Then `::apply` won't be necessary

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/backgroundables/SlideInBackgroundable.java
```diff
@@ -0,0 +1,158 @@
...
+        .toOption();
+
+    newBranchLayout.map(nbl -> Try.run(() -> branchLayoutWriter.write(macheteFilePath, nbl, /* backupOldLayout */ true))
+        .onFailure(t -> notifier.notifyError(
+            /* title */ getString(
+                "action.GitMachete.BaseSlideInBranchBelowAction.notification.title.branch-layout-write-fail"),
+            getMessageOrEmpty(t))));
+  }
+
+  private void waitForLocalBranch() {
```
PawelLipski, line 120: For clarity:
```suggestion
  private void waitForCreationOfLocalBranch() {
```

PR #503, opened 2020-08-11
Submodule branch name checking fix

micpiotrowski:


frontendResourceBundles/src/main/resources/GitMacheteBundle.properties
```diff
@@ -182,23 +182,33 @@ action.GitMachete.SlideOutSelectedBranchAction.description=Slide out selected br
 
 action.GitMachete.DiscoverAction.text=Discover Branch Layout
 action.GitMachete.DiscoverAction.description=Discover Branch Layout Tree
-action.GitMachete.DiscoverAction.cant-discover-layout-error-title=Can't discover repository layout
-action.GitMachete.DiscoverAction.cant-get-current-repository-error-title=Can't get current git repository
+action.GitMachete.DiscoverAction.cant-discover-layout-error-title=Can''t discover repository layout
```
PawelLipski, line 185: @mkondratek @micpiotrowski will it be still correct if we just forbid single `'` in the strings completely? :thinking:

frontendResourceBundles/src/main/resources/GitMacheteBundle.properties
```diff
@@ -182,23 +182,33 @@ action.GitMachete.SlideOutSelectedBranchAction.description=Slide out selected br
...
 string.GitMachete.EnhancedGraphTable.omitted-branches-text=The following branches defined by machete file do not belong to the local repository:
 
 
+# Machete file editor
+
+string.GitMachete.MacheteAnnotator.cant-find-local-branch-in-repo=Can''t find local branch ''{0}'' in git repository
+string.GitMachete.MacheteAnnotator.cant-indent-first-entry=First entry cannot be indented
+string.GitMachete.MacheteAnnotator.cant-retrieve-local-branches=Local branches cannot be retrieved, so they will not be checked
+string.GitMachete.MacheteAnnotator.indent-char-not-match=Indentation character ({0}) does not match the indentation character in first indented line ({1})
+string.GitMachete.MacheteAnnotator.indent-width-not-match=Indentation width is not multiple of {0} as first indented line suggests
```
PawelLipski, line 208: Nit:
```suggestion
string.GitMachete.MacheteAnnotator.indent-char-not-match=Indentation character ({0}) does not match the indentation character in the first indented line ({1})
string.GitMachete.MacheteAnnotator.indent-width-not-match=Indentation width is not multiple of {0} as the first indented line suggests
```

frontendResourceBundles/src/main/resources/GitMacheteBundle.properties
micpiotrowski: With single `'` only this is evaluated to no `'` at all (IIUC your question).

frontendFile/src/main/java/com/virtuslab/gitmachete/frontend/file/highlighting/MacheteAnnotator.java
```diff
@@ -121,23 +126,24 @@ private void processIndentationElement(PsiElement element, AnnotationHolder hold
     var wrongIndentChar = thisIndentationText.chars().filter(c -> c != indentationParameters.indentationCharacter).findFirst();
```
PawelLipski, line 126: Nit: use explicit `OptionalInt` since the type of `wrongIndentChar` isn't obvious

frontendResourceBundles/src/main/resources/GitMacheteBundle.properties
PawelLipski: Ok but can we ban the regex `[^']'[^']` completely in this file?

It'd be nice to answer the following questions:

1. is there any context where `''` (double single quote) will NOT render as `'`?
2. is there any context where `'` (single single quote) will  NOT render as `'`?

I suspect the 1=>NO && 2=>YES... so forbidding single single quote would be a good idea then :thinking:

PawelLipski: Also the comment `WE use single quotes for indication` in line 2 is slightly misleading, maybe:

```
# Action descriptions & texts - no formatting allowed (WE use double single quotes, which render as single single quotes for indication)
```

frontendResourceBundles/src/main/resources/GitMacheteBundle.properties
```diff
@@ -205,23 +205,33 @@ string.GitMachete.BranchOrCommitCellRendererComponent.sync-to-remote-status-text
...
 action.GitMachete.DiscoverAction.discovered-branch-tree-dialog.save-button-text=Save Discovered Layout
 action.GitMachete.DiscoverAction.discovered-branch-tree-dialog.title=Discovered branch layout
 action.GitMachete.DiscoverAction.repository-discover-error-title=Repository discover error
 action.GitMachete.DiscoverAction.write-file-task-title=Writing branch layout file
-action.GitMachete.DiscoverAction.write-file-error-title=Can't write discovered branch layout
+action.GitMachete.DiscoverAction.write-file-error-title=Can''t write discovered branch layout
 
 string.GitMachete.EnhancedGraphTable.empty-machete-file.upper-text=Provided machete file ({0}) is empty.
-string.GitMachete.EnhancedGraphTable.empty-machete-file.lower-text=Discover the repository's branch layout
+string.GitMachete.EnhancedGraphTable.empty-machete-file.lower-text=Discover the repository''s branch layout
```
PawelLipski, line 218: Okay... so this renders correctly as `'`, right? If so (and if the answers to questions above are ` 1=>NO && 2=>YES`), then add a CI script against _single single_ quotes in this file

frontendResourceBundles/src/main/resources/GitMacheteBundle.properties
micpiotrowski: The followed rule of thumb provided by @PawelLipski via Slack conversation should be fine:
* strings that have any `{N}` must NOT have `'`, but `''` instead
* strings that do NOT have any `{N}` are NOT meant to be passed through MessageFormat and hence must NOT have `''` but plain `'` instead

micpiotrowski: These are detailed rules about `'` in `MessageFormat`:
```
Within a String, a pair of single quotes can be used to quote any arbitrary characters except single quotes. For example, pattern string "'{0}'" represents string "{0}", not a FormatElement. A single quote itself must be represented by doubled single quotes '' throughout a String. For example, pattern string "'{''}'" is interpreted as a sequence of '{ (start of quoting and a left curly brace), '' (a single quote), and }' (a right curly brace and end of quoting), not '{' and '}' (quoted left and right curly braces): representing string "{'}", not "{}".

A SubformatPattern is interpreted by its corresponding subformat, and subformat-dependent pattern rules apply. For example, pattern string "{1,number,$'#',##}" (SubformatPattern with underline) will produce a number format with the pound-sign quoted, with a result such as: "$#31,45". Refer to each Format subclass documentation for details.

Any unmatched quote is treated as closed at the end of the given pattern. For example, pattern string "'{0}" is treated as pattern "'{0}'".

Any curly braces within an unquoted pattern must be balanced. For example, "ab {0} de" and "ab '}' de" are valid patterns, but "ab {0'}' de", "ab } de" and "''{''" are not.
```
Source: https://docs.oracle.com/javase/7/docs/api/java/text/MessageFormat.html

frontendResourceBundles/src/main/resources/GitMacheteBundle.properties
micpiotrowski: No, It's not rendered correctly. My fault. Corrected on [revamp/user-facing-texts](https://github.com/VirtusLab/git-machete-intellij-plugin/pull/504)

frontendResourceBundles/src/main/resources/GitMacheteBundle.properties
PawelLipski: Ok, pls provide a script to enforce the two above rules

frontendResourceBundles/src/main/resources/GitMacheteBundle.properties
```diff
@@ -1,5 +1,5 @@
 # Formatting
-# Action descriptions & texts - no formatting allowed (WE use single quotes for indication)
+# Action descriptions & texts - no formatting allowed (for indication we use double single quotes, which are rendered as single single quotes)
```
mkondratek, line 2: whoops, I wrote it so badly :facepalm:

PawelLipski: +:care:

PR #515, opened 2020-08-13
Nonexistent repository log error fix

micpiotrowski:
In this case, logging itself was a bug :bug:
When Git Machete was opened (at the same beginning) `KEY_GIT_MACHETE_REPOSITORY_SNAPSHOT` was unavailable (and this is normal at this stage) but the new version of `onUpdate` for `*Current*` actions used to use this key by calling `getNameOfBranchUnderAction` method that used to call `getGitMacheteRepositorySnapshotWithLoggingOnEmpty` that caused log production.

Methods in `IExpectedKeyGitMacheteRepository` was renamed (and some was created) to clearly distinguish which do logging and which not.
Methods without logging were used in `*Current*` actions on `onUpdate` method.

In addition, for `BaseResetBranchToRemoteAction` class `getNameOfBranchUnderAction` was created (this method is without logging, previous `getNameOfBranchUnderAction` was renamed to `getNameOfBranchUnderActionWithLogging`). This method is used in `onUpdate` in this class.

### DIFF
--- a/frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseOverrideForkPointAction.java
+++ b/frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseOverrideForkPointAction.java
@@ -63,8 +63,8 @@ public abstract class BaseOverrideForkPointAction extends BaseGitMacheteReposito
   public void actionPerformed(AnActionEvent anActionEvent) {
     var project = getProject(anActionEvent);
     var gitRepository = getSelectedGitRepository(anActionEvent).getOrNull();
-    var branchUnderAction = getNameOfBranchUnderAction(anActionEvent);
-    var branch = branchUnderAction.flatMap(pn -> getGitMacheteBranchByName(anActionEvent, pn)).getOrNull();
+    var branchUnderAction = getNameOfBranchUnderActionWithLogging(anActionEvent);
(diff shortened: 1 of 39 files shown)


frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseOverrideForkPointAction.java
```diff
@@ -63,8 +63,8 @@ public void onUpdate(AnActionEvent anActionEvent) {
   public void actionPerformed(AnActionEvent anActionEvent) {
     var project = getProject(anActionEvent);
     var selectedVcsRepository = getSelectedGitRepository(anActionEvent).getOrNull();
-    var branchUnderAction = getNameOfBranchUnderAction(anActionEvent);
-    var branch = branchUnderAction.flatMap(pn -> getGitMacheteBranchByName(anActionEvent, pn)).getOrNull();
+    var branchUnderAction = getNameOfBranchUnderActionWithLogging(anActionEvent);
+    var branch = branchUnderAction.flatMap(pn -> getGitMacheteBranchByNameWithLoggingOnEmpty(anActionEvent, pn)).getOrNull();
```
mkondratek, line 67: inconsistency alert :exclamation: 

`<something>WithLogging`
VS
`<something>WithLoggingOnEmpty`

I think that we could you the first one since it is shorter

PR #522, opened 2020-08-14
Construct refspecs by a dedicated helper method rather than manually

PawelLipski:


frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/common/ActionUtils.java
```diff
@@ -5,6 +5,10 @@
 public final class ActionUtils {
   private ActionUtils() {}
 
+  public static String createRefspec(String sourceBranchFullName, String targetBranchFullName, boolean allowNonFastForward) {
+    return (allowNonFastForward ? "+" : "") + sourceBranchFullName + ":" + targetBranchFullName;
```
mkondratek, line 9: So I assume that the current version is so descriptive that we can consciously drop the comment
`// Note the '+' sign preceding the refspec. It permits non-fast-forward updates.`
right?

PawelLipski: Yes, that was exactly the idea

PR #525, opened 2020-08-17
Correctly skip non-existent root branches from the layout (and not only non-roots)

PawelLipski:


backendImpl/src/main/java/com/virtuslab/gitmachete/backend/impl/GitMacheteRepository.java
```diff
@@ -376,18 +381,18 @@ private RootCreatedBranchAndSkippedBranches createGitMacheteRootBranch(
       var createdRootBranch = new RootManagedBranchSnapshot(branchName, branchFullName,
           childBranches.getCreatedBranches(), pointedCommit, remoteTrackingBranch, syncToRemoteStatus, customAnnotation,
           statusHookOutput);
-      return RootCreatedBranchAndSkippedBranches.of(createdRootBranch, childBranches.getSkippedBranchNames());
+      return CreatedAndSkippedBranches.of(createdRootBranch, childBranches.getSkippedBranchNames());
```
mkondratek, line 384: My first thought was that it could be a branch that is BOTH created and skipped :thinking: I assume that it is not so easy to reflect the accurate sense here

mkondratek: My thinking may be the actual problem here :trollface:

PawelLipski: This could theoretically be `CreatedBranch{es,Snapshots}AndSkippedBranchNames`... but in turn this would be too verbose :/

PR #528, opened 2020-08-18
Fix too wide column in some cases

micpiotrowski:
* `BaseGraphTable#resetColumnWidth` was added. This method must be invoked after every graph model change and before `repaint` and `revalidate`.
* From now (as suggested on an issue) when a right click menu is visible, the column width is expanded to fill the window panel. This cause selection has full width (for better feeling).


frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/EnhancedGraphTable.java
```diff
@@ -339,7 +343,11 @@ public void popupMenuWillBecomeVisible(PopupMenuEvent popupMenuEvent) {
       new Timer().schedule(new TimerTask() {
         @Override
         public void run() {
-          GuiUtils.invokeLaterIfNeeded(() -> graphTable.setRowSelectionAllowed(true), NON_MODAL);
+          GuiUtils.invokeLaterIfNeeded(() -> {
+            // To selection have a full width (for better feeling)
```
PawelLipski, line 347: :grammar-nazi:
```suggestion
            // So that the selection spans over the full width of the row (for better feeling)
```

frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/EnhancedGraphTable.java
```diff
@@ -348,6 +356,8 @@ public void run() {
     @UIEffect
     public void popupMenuWillBecomeInvisible(PopupMenuEvent popupMenuEvent) {
       graphTable.setRowSelectionAllowed(false);
+      // To selection (and thus the area where tooltip is shown) have again the possible smallest size
```
PawelLipski, line 359:
```suggestion
      // So that the selection (and thus the area where tooltip is shown) has again the possible smallest size
```

PawelLipski: (`So that <subject> <verb 3rd person>` or `For <subject> to <verb infitive>...`)

frontendUiApi/src/main/java/com/virtuslab/gitmachete/frontend/ui/api/table/BaseGraphTable.java
```diff
@@ -38,6 +38,15 @@ public Component prepareRenderer(TableCellRenderer renderer, int row, int column
     return component;
   }
 
+  /**
+   * This method should be invoked after each graph table model change but BEFORE {@code revalidate} and {@code repaint}
+   * It cause the max column width is reset and can be properly recalculated in {@code prepareRenderer}
```
PawelLipski, line 43: Hmm but how is `prepareRenderer` invoked, in turn? pls add to comment

PawelLipski: Also:
```suggestion
   * It resets the max column width so that it can be properly recalculated in {@code prepareRenderer}
```

PawelLipski: Also, wrt. `prepareRenderer`... maybe instead just rework `prepareRenderer` without a need to invoke a separate method like `resetColumnWidth`? This would simplify the logic if possible

micpiotrowski: Ok, reworked.

frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/EnhancedGraphTable.java
```diff
@@ -348,6 +353,8 @@ public void run() {
     @UIEffect
     public void popupMenuWillBecomeInvisible(PopupMenuEvent popupMenuEvent) {
       graphTable.setRowSelectionAllowed(false);
+      // So that the selection (and thus the area where tooltip is shown) has again the possible smallest size
```
PawelLipski, line 356: Nit (didn't notice previously):
```suggestion
      // So that the selection (and thus the area where tooltip is shown) has again the smallest possible size
```

frontendUiApi/src/main/java/com/virtuslab/gitmachete/frontend/ui/api/table/BaseGraphTable.java
```diff
@@ -9,11 +9,14 @@
 import com.intellij.ui.SimpleTextAttributes;
 import com.intellij.ui.table.JBTable;
 import com.intellij.util.ui.JBUI;
+import io.vavr.collection.List;
 import org.checkerframework.checker.guieffect.qual.UI;
 import org.checkerframework.checker.guieffect.qual.UIEffect;
 import org.checkerframework.checker.nullness.qual.Nullable;
 
 public abstract class BaseGraphTable extends JBTable {
+  private static List<Integer> rowWidths = List.empty();
```
PawelLipski, line 18: Hmmm shouldn't it be an instance field? What if there is more than `BaseGraphTable` instance in the given JVM (when help/discover action is launched and/or multiple projects are opened)?

micpiotrowski: Whoops, right.

frontendUiApi/src/main/java/com/virtuslab/gitmachete/frontend/ui/api/table/BaseGraphTable.java
```diff
@@ -9,11 +9,14 @@
 import com.intellij.ui.SimpleTextAttributes;
 import com.intellij.ui.table.JBTable;
 import com.intellij.util.ui.JBUI;
+import io.vavr.collection.List;
 import org.checkerframework.checker.guieffect.qual.UI;
 import org.checkerframework.checker.guieffect.qual.UIEffect;
 import org.checkerframework.checker.nullness.qual.Nullable;
 
 public abstract class BaseGraphTable extends JBTable {
+  private List<Integer> rowWidths = List.empty();
```
PawelLipski, line 18: Nit: mark as `@UIEffect` to prevent access from outside UI thread

frontendUiApi/src/main/java/com/virtuslab/gitmachete/frontend/ui/api/table/BaseGraphTable.java
```diff
@@ -31,10 +34,16 @@ protected BaseGraphTable(TableModel model) {
   @Override
   @UIEffect
   public Component prepareRenderer(TableCellRenderer renderer, int row, int column) {
+    if (getRowCount() != rowWidths.size()) {
+      rowWidths = List.fill(getRowCount(), 0);
```
PawelLipski, line 38: Hmm by assigning a fresh, zero-filled list you're losing the data which was already in the list... possibly you want to just crop or extend the list without losing the existing data? or if we don't care about the pre-existing data when `getRowCount() != rowWidths.size()`, then document why so

micpiotrowski: No, I won't crop an existing list. This fresh list assignment is intentional. Documentation left.

PawelLipski (approved):

mkondratek (changes requested):
image

Since we now mark the whole line... is this column width limitation required? 
One can drag the graph from the left to the right and see the following in the result (image above). 
That looks crappy :hankey:

micpiotrowski:
Yes, it's still needed because without this tooltips will be displayed on the whole window panel width and not only (less or more) on branches.
Writing own Swing component for displaying this graph is probably the best solution (and it's probably not so hard)

PawelLipski:
> Writing own Swing component for displaying this graph is probably the best solution (and it seems to be not so hard)

Whoops... is it possible without writing our own Swing component to prevent the above problem when moving the table?

PawelLipski:
Also, given that the changes are quite fundamental & non-trivial... let's move this issue/PR to v0.7.1 (@micpiotrowski pls bump the version in version.gradle)

mkondratek:
I think that what we presently do with tooltip (limiting cell width) is **hackish**.
Personally... `tooltips will be displayed on the whole window panel width` is better than this

The problem, compared to the VCS log
image
is that:
- we have no right-side panel that would fill some width and shorten the width of our table
- we have no `author`, `date`, `hash` (no tags, that works too), in particular, nothing that belongs to the table and is aligned to the right

image

mkondratek:
just a stupid idea... can't we show tooltip on the text instead of the cell? that would solve the problem, right?

PawelLipski (approved):

PawelLipski (approved):

PR #530, opened 2020-08-18
Fix the logic of when ToggleListingCommitsAction gets disabled

PawelLipski:


frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/toolbar/ToggleListingCommitsAction.java
```diff
@@ -45,16 +44,6 @@ public void onUpdate(AnActionEvent anActionEvent) {
       return;
     }
 
-    boolean noChildBranchExists = branchLayout.get().getRootEntries()
-        .exists(rootBranch -> rootBranch.getChildren().isEmpty());
```
PawelLipski, line 49: There was a bug here... it should've been `.forAll`, not `.exists`

PawelLipski: But anyway, we don't need this check, `anyCommitExists == false` covers the case of no child branch as well

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/toolbar/ToggleListingCommitsAction.java
```diff
@@ -45,16 +44,6 @@ public void onUpdate(AnActionEvent anActionEvent) {
       return;
```
micpiotrowski, line 44: whoops... according to PR #515 on line 40 should be getBranchLayoutWith**out**Logging

micpiotrowski (changes requested):

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/toolbar/ToggleListingCommitsAction.java
PawelLipski: Fixed... but it's super-hard to enforce, lemme solve that `systemically` on another PR :thinking:

PawelLipski: Ok, done on #534

mkondratek (approved):

micpiotrowski (approved):

PR #533, opened 2020-08-19
Ensure each project-dependent action is disabled & hidden as long as PROJECT data key is missing

PawelLipski:
That's a potentially dangerous change, moving to next release (v0.7.1)


frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseGitMacheteRepositoryReadyAction.java
```diff
@@ -7,7 +7,7 @@
 import com.virtuslab.gitmachete.frontend.actions.expectedkeys.IExpectsKeyGitMacheteRepository;
 import com.virtuslab.gitmachete.frontend.resourcebundles.GitMacheteBundle;
 
-public abstract class BaseGitMacheteRepositoryReadyAction extends BaseProjectKeyAvailabilityAssuranceAction
+public abstract class BaseGitMacheteRepositoryReadyAction extends BaseProjectDependentAction
```
PawelLipski, line 10: Basically, there's *no access to the project key* (`getProject`) unless we're first sure that the project key is actually available - and this check only ever happens in `BaseProjectDependentAction#update`.

Most of the time the project is available... this is only covering corner cases around opening the project.

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseProjectDependentAction.java
```diff
@@ -0,0 +1,69 @@
...
+  public final void update(AnActionEvent anActionEvent) {
+    super.update(anActionEvent);
+
+    var maybeProject = anActionEvent.getProject();
+    var presentation = anActionEvent.getPresentation();
+    if (maybeProject == null) {
+      presentation.setEnabledAndVisible(false);
+    } else {
+      presentation.setEnabledAndVisible(true);
+      onUpdate(anActionEvent);
```
PawelLipski, line 29: Also! To prevent actions from becoming stateful (i.e. so that: what happens in `update` invocation #N should NOT depend on what happened in invocation #N-1), I've ensured that `update` resets the action state. It's enabled & visible by default if project key available, otherwise disabled & hidden; the prior state of the action doesn't matter.

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseProjectDependentAction.java
```diff
@@ -0,0 +1,69 @@
...
+
+  /**
+   * If overridden, {@code super.onUpdate(anActionEvent)} should always be called in the first line of overriding method.
+   * In addition, in this method we should use getters WITHOUT logging like
+   * {@code IExpectsKeyGitMacheteRepository#getGitMacheteRepositorySnapshotWithoutLogging}
+   *
+   * @param anActionEvent an action event
+   */
+  @UIEffect
+  protected abstract void onUpdate(AnActionEvent anActionEvent);
```
micpiotrowski, line 41: Because in many actions this method is just empty, so maybe get rid of `abstract` here and make it empty by default?

micpiotrowski (approved):

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseProjectDependentAction.java
PawelLipski: Good idea, done

mkondratek (approved):
lgtm

Issue #176, opened 2020-04-08
Propose automatically detected branch layout for user when machete file is not present or when it's empty


micpiotrowski:
Extracted from issue #44 

- [x] We want to propose automatically detected (by `discover` functionality) branch layout rather than only show message for user.

@PawelLipski :

- [ ] Note that the backend API for discover can be pretty easily configurable wrt. number of branches that we want to include in the discovered layout - most likely this could as well be configurable from the frontend.

- [ ] Also, the usage instructions (see #418) need to be updated

PawelLipski:
Note that the backend API for discover can be pretty easily configurable wrt. number of branches that we want to include in the discovered layout - most likely this could as well be configurable from the frontend.

PawelLipski:
- [ ] Also, the usage instructions (see #418) need to be updated

PR #550, opened 2020-08-26
Implement autodiscover in case of empty or absent machete file

micpiotrowski:
- [ ] once it's merged remember to update general usage instructions

### DIFF
--- a/frontendResourceBundles/src/main/resources/GitMacheteBundle.properties
+++ b/frontendResourceBundles/src/main/resources/GitMacheteBundle.properties
@@ -227,9 +227,9 @@ string.GitMachete.BranchOrCommitCellRendererComponent.sync-to-remote-status-text
 string.GitMachete.EnhancedGraphTable.automatic-discover.cant-discover-layout-error-title=Can't automatically discover repository layout
 string.GitMachete.EnhancedGraphTable.automatic-discover.task-title=Automatically discovering branch layout
 string.GitMachete.EnhancedGraphTable.automatic-discover.success-message=Branch layout has been automatically discovered
-string.GitMachete.EnhancedGraphTable.empty-machete-file.upper-text=Provided machete file ({0}) is empty and branch layout can''t be automatically detected
-string.GitMachete.EnhancedGraphTable.empty-text=Loading...
-string.GitMachete.EnhancedGraphTable.omitted-branches-text=The following branches defined by machete file do not belong to the local repository:
+string.GitMachete.EnhancedGraphTable.empty-table-text.cant-discover-layout=Provided machete file ({0}) is empty and branch layout can''t be automatically detected
+string.GitMachete.EnhancedGraphTable.empty-table-text.loading=Loading...
+string.GitMachete.EnhancedGraphTable.skipped-branches-text=The following branches defined by machete file do not belong to the local repository:
 
 
 # Unanchored
--- a/frontendResourceBundles/src/main/resources/GitMacheteBundle.properties
+++ b/frontendResourceBundles/src/main/resources/GitMacheteBundle.properties
@@ -30,6 +30,9 @@ action.GitMachete.GitCommandUpdatingCurrentBranchBackgroundable.notification.mes
 action.GitMachete.GitCommandUpdatingCurrentBranchBackgroundable.notification.title.update-fail={0} failed
 action.GitMachete.GitCommandUpdatingCurrentBranchBackgroundable.notification.title.all-files-are-up-to-date=All files are up-to-date
 
+# Quasi-title capitalization intended since we always write "Git Machete" with initial caps.
+action.GitMachete.GitMacheteRepositoryUpdateBackgroundable.task-title=Updating Git Machete status
+
 action.GitMachete.MergeCurrentBranchFastForwardOnlyBackgroundable.operation-name=Fast-forward merge
 
 action.GitMachete.PullCurrentBranchFastForwardOnlyBackgroundable.operation-name=Pull
@@ -46,6 +49,7 @@ action.GitMachete.BaseGitMacheteRepositoryReadyAction.description.disabled.undef
 
 
 action.GitMachete.BaseOverrideForkPointAction.description-action-name=Fork point override
+# {0} skipped intentionally (parent branch name is not needed)
 action.GitMachete.BaseOverrideForkPointAction.description=Override fork point of ''{1}''
 action.GitMachete.BaseOverrideForkPointAction.dialog.override-fork-point.label.branch-name=Select where the range of commits belonging to this branch starts
 action.GitMachete.BaseOverrideForkPointAction.dialog.override-fork-point.title=Override Fork Point
@@ -231,7 +235,7 @@ string.GitMachete.EnhancedGraphTable.omitted-branches-text=The following branche
 # Unanchored
 
 action.GitMachete.DiscoverAction.text=Discover Branch Layout
-action.GitMachete.DiscoverAction.description=Discover Branch Layout Tree
+action.GitMachete.DiscoverAction.description=Discover branch layout tree
 action.GitMachete.DiscoverAction.discovered-branch-tree-dialog.empty-table-text=No branch has been discovered
 action.GitMachete.DiscoverAction.discovered-branch-tree-dialog.save-button-text=Save
 action.GitMachete.DiscoverAction.discovered-branch-tree-dialog.title=Discovered branch layout
@@ -334,7 +338,7 @@ action.GitMachete.ToggleListingCommitsAction.description.disabled.no-branches=To
 # VCS menu
 
 action.GitMachete.OpenMacheteTabAction.text=Open Git Machete Tab
-action.GitMachete.OpenMacheteTabAction.description=Open Git Machete Tab
+action.GitMachete.OpenMacheteTabAction.description=Open Git Machete tab
 
 action.GitMachete.OpenMacheteTabAction.notification.title.could-not-open-tab=Could not open Git Machete tab
 action.GitMachete.OpenMacheteTabAction.notification.message.no-git=Ensure that Git VCS is enabled in the project
--- a/frontendResourceBundles/src/main/resources/GitMacheteBundle.properties
+++ b/frontendResourceBundles/src/main/resources/GitMacheteBundle.properties
@@ -220,6 +220,16 @@ string.GitMachete.BranchOrCommitCellRendererComponent.sync-to-remote-status-text
 string.GitMachete.BranchOrCommitCellRendererComponent.sync-to-remote-status-text.diverged-from-and-older-than-remote=(diverged from & older than {0})
 string.GitMachete.BranchOrCommitCellRendererComponent.sync-to-remote-status-text.untracked=(untracked)
 
+string.GitMachete.EnhancedGraphTable.automatic-discover.cant-discover-layout-error-title=Can't automatically discover repository layout
+string.GitMachete.EnhancedGraphTable.automatic-discover.task-title=Automatically discovering branch layout
+string.GitMachete.EnhancedGraphTable.automatic-discover.success-message=Branch layout has been automatically discovered
+string.GitMachete.EnhancedGraphTable.empty-machete-file.upper-text=Provided machete file ({0}) is empty and branch layout can''t be automatically detected
+string.GitMachete.EnhancedGraphTable.empty-text=Loading...
+string.GitMachete.EnhancedGraphTable.omitted-branches-text=The following branches defined by machete file do not belong to the local repository:
+
+
+# Unanchored
+
 action.GitMachete.DiscoverAction.text=Discover Branch Layout
 action.GitMachete.DiscoverAction.description=Discover Branch Layout Tree
 action.GitMachete.DiscoverAction.discovered-branch-tree-dialog.empty-table-text=No branch has been discovered
@@ -230,13 +240,6 @@ action.GitMachete.DiscoverAction.notification.title.cant-get-current-repository-
 action.GitMachete.DiscoverAction.notification.title.repository-discover-error=Repository discover error
 action.GitMachete.DiscoverAction.notification.title.write-file-error=Can't write discovered branch layout
 action.GitMachete.DiscoverAction.write-file-task-title=Writing branch layout file
-
-string.GitMachete.EnhancedGraphTable.empty-machete-file.upper-text=Provided machete file ({0}) is empty.
-string.GitMachete.EnhancedGraphTable.empty-machete-file.lower-text=Discover the repository's branch layout
-string.GitMachete.EnhancedGraphTable.empty-text=Loading...
-string.GitMachete.EnhancedGraphTable.nonexistent-machete-file.upper-text=There is no machete file ({0}) for this repository.
-string.GitMachete.EnhancedGraphTable.nonexistent-machete-file.lower-text=Discover the repository's branch layout
-string.GitMachete.EnhancedGraphTable.omitted-branches-text=The following branches defined by machete file do not belong to the local repository:
 
 
 # Machete file editor
--- a/frontendResourceBundles/src/main/resources/GitMacheteBundle.properties
+++ b/frontendResourceBundles/src/main/resources/GitMacheteBundle.properties
@@ -229,6 +229,7 @@ string.GitMachete.EnhancedGraphTable.automatic-discover.task-title=Automatically
 string.GitMachete.EnhancedGraphTable.automatic-discover.success-message=Branch layout has been automatically discovered
 string.GitMachete.EnhancedGraphTable.empty-table-text.cant-discover-layout=Provided machete file ({0}) is empty and branch layout can''t be automatically detected
 string.GitMachete.EnhancedGraphTable.empty-table-text.loading=Loading...
+string.GitMachete.EnhancedGraphTable.empty-table-text.only-skipped-in-machete-file=None of the branches provided by machete file ({0}) exists within the local repository
 string.GitMachete.EnhancedGraphTable.skipped-branches-text=The following branches defined by machete file do not belong to the local repository:
 
 
--- a/frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/EnhancedGraphTable.java
+++ b/frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/EnhancedGraphTable.java
@@ -57,6 +57,7 @@ import org.checkerframework.checker.guieffect.qual.UIEffect;
 import org.checkerframework.checker.nullness.qual.Nullable;
 
 import com.virtuslab.binding.RuntimeBinding;
+import com.virtuslab.branchlayout.api.BranchLayoutException;
 import com.virtuslab.branchlayout.api.readwrite.IBranchLayoutReader;
 import com.virtuslab.gitmachete.backend.api.IGitMacheteRepositoryCache;
 import com.virtuslab.gitmachete.backend.api.IGitMacheteRepositorySnapshot;
@@ -79,7 +80,6 @@ import com.virtuslab.gitmachete.frontend.vfsutils.GitVfsUtils;
  *  data like last clicked branch name, opened project or {@link IGitMacheteRepositorySnapshot} of current
  *  repository for actions
  */
-
 // TODO (#99): consider applying SpeedSearch for branches and commits
 @CustomLog
 public final class EnhancedGraphTable extends BaseEnhancedGraphTable
@@ -97,8 +97,8 @@ public final class EnhancedGraphTable extends BaseEnhancedGraphTable
   @UIEffect
   private boolean isListingCommits;
 
-  @UIEffect
   @Getter
+  @UIEffect
   private @Nullable IGitMacheteRepositorySnapshot gitMacheteRepositorySnapshot;
 
   @UIEffect
@@ -159,8 +159,6 @@ public final class EnhancedGraphTable extends BaseEnhancedGraphTable
 
   @UIEffect
   private void refreshModel(GitRepository gitRepository, List<String> skippedBranchNames) {
-    // TODO (#176): When machete file is absent or empty,
-    // propose using branch layout automatically detected by discover functionality
     if (!project.isInitialized() || ApplicationManager.getApplication().isUnitTestMode()) {
       LOG.debug("Project is not initialized or application is in unit test mode. Returning.");
       return;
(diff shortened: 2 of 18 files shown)


frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/speedsearch/DiscoverAction.java
```diff
@@ -1,27 +1,28 @@
-package com.virtuslab.gitmachete.frontend.actions.graphtable;
+package com.virtuslab.gitmachete.frontend.actions.speedsearch;
```
micpiotrowski, line 1: Changes in this file are unrelated

mkondratek: I would rather avoid piggybacking anything on bug fixing PRs (to lower the chances of providing another bug :rofl:)

---

Anyway... Why this change (rename)?
Where is the connection between this package and the IntelliJ speed search?

Afaik the speed search is an ability to simply write a text within some window/dialog and search (results marked with yellow on the light theme). If there is no connection to this feature please revert this change / rename to smth else to avoid confusion.

PawelLipski: Yeah, I don't like the `speedsearch` name either... just `unanchored` (?) would be better if this action doesn't belong to any particular `ActionPlace`

frontendResourceBundles/src/main/resources/GitMacheteBundle.properties
```diff
@@ -209,6 +209,16 @@ string.GitMachete.BranchOrCommitCellRendererComponent.sync-to-remote-status-text
 string.GitMachete.BranchOrCommitCellRendererComponent.sync-to-remote-status-text.diverged-from-and-older-than-remote=(diverged from & older than {0})
 string.GitMachete.BranchOrCommitCellRendererComponent.sync-to-remote-status-text.untracked=(untracked)
 
+string.GitMachete.EnhancedGraphTable.automatic-discover.cant-discover-layout-error-title=Can't automatically discover repository layout
+string.GitMachete.EnhancedGraphTable.automatic-discover.task-title=Automatic discovering branch layout
```
PawelLipski, line 213:
```suggestion
string.GitMachete.EnhancedGraphTable.automatic-discover.task-title=Automatically discovering branch layout
```

frontendResourceBundles/src/main/resources/GitMacheteBundle.properties
```diff
@@ -209,6 +209,16 @@ string.GitMachete.BranchOrCommitCellRendererComponent.sync-to-remote-status-text
...
+string.GitMachete.EnhancedGraphTable.automatic-discover.success-message=Branch layout has been automatically detected
```
PawelLipski, line 214: I'd rather use `discover` consistently instead of `detect`

frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/EnhancedGraphTable.java
```diff
@@ -208,13 +209,74 @@ private void refreshModel(GitRepository gitRepository, List<String> skippedBranc
...
+  private void doAutomaticDiscover(Path macheteFilePath) {
+    var selectedRepository = getGitRepositorySelectionProvider().getSelectedGitRepository().getOrNull();
+    if (selectedRepository == null) {
+      LOG.error("Can't do automatic discover because of undefined selected repository");
+      return;
+    }
+    var mainDirPath = GitVfsUtils.getMainDirectoryPath(selectedRepository).toAbsolutePath();
+    var gitDirPath = GitVfsUtils.getGitDirectoryPath(selectedRepository).toAbsolutePath();
+
+    @SuppressWarnings({"nullness", "guieffect"}) // To allow runResult be uninitialized and Tuple accept @Nullable IBranchLayout
```
PawelLipski, line 221: Hmm what's the specific error caused by `runResult` being uninitialized? :thinking: Is it about dereference of `runResult` in line 245?

PawelLipski: If so, then consider checking against `runResult == null` in `onSuccess`

frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/EnhancedGraphTable.java
```diff
@@ -208,13 +209,74 @@ private void refreshModel(GitRepository gitRepository, List<String> skippedBranc
...
+              return Tuple(branchLayoutWriter, branchLayout, repositorySnapshot);
+            })
+            .filterTry(tuple -> tuple._2 != null, __ -> new EmptyBranchLayoutException())
+            .andThenTry(tuple -> tuple._1.write(macheteFilePath, tuple._2, /* backupBranchLayout */ true));
+      }
+      // This method is invoked on UI dispatch thread
+      @UIEffect
+      @Override
+      public void onSuccess() {
+        // Match below returns empty string just because must return something
```
PawelLipski, line 244: Hmm TBH `Match` feels a bit artificial here (also due to the reason stated in this comment)... anyway, maybe it's not a good idea to handle the failures in `onSuccess` callback? Can we use `onFailure` or a similar callback for that purpose?

frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/EnhancedGraphTable.java
```diff
@@ -208,13 +209,74 @@ private void refreshModel(GitRepository gitRepository, List<String> skippedBranc
...
+      @Override
+      public void run(ProgressIndicator indicator) {
+        runResult = Try.of(() -> RuntimeBinding.instantiateSoleImplementingClass(IGitMacheteRepositoryCache.class)
+            .getInstance(mainDirPath, gitDirPath).discoverLayoutAndCreateSnapshot())
+            .filterTry(repositorySnapshot -> repositorySnapshot.getRootBranches().size() > 0,
+                __ -> new EmptyRepositoryException())
+            .map(repositorySnapshot -> {
+              var branchLayoutWriter = project.getService(BranchLayoutWriterProvider.class).getBranchLayoutWriter();
+              var branchLayout = repositorySnapshot.getBranchLayout().getOrNull();
+              return Tuple(branchLayoutWriter, branchLayout, repositorySnapshot);
```
PawelLipski, line 235: I'd rather create a dedicated private static class for the result... tuples with 3+ fields of different types are confusing

frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/EnhancedGraphTable.java
```diff
@@ -208,13 +209,74 @@ private void refreshModel(GitRepository gitRepository, List<String> skippedBranc
...
+    var gitDirPath = GitVfsUtils.getGitDirectoryPath(selectedRepository).toAbsolutePath();
+
+    @SuppressWarnings({"nullness", "guieffect"}) // To allow runResult be uninitialized and Tuple accept @Nullable IBranchLayout
+    var task = new Task.Backgroundable(project,
+        getString("string.GitMachete.EnhancedGraphTable.automatic-discover.task-title")) {
+      private Try<Tuple3<IBranchLayoutWriter, @Nullable IBranchLayout, IGitMacheteRepositorySnapshot>> runResult;
+
+      @Override
+      public void run(ProgressIndicator indicator) {
+        runResult = Try.of(() -> RuntimeBinding.instantiateSoleImplementingClass(IGitMacheteRepositoryCache.class)
```
PawelLipski, line 228: TBH I think this code might read easier if un-FP-ified ;D Possibly even the failure/success can be dispatched onto UI thread directly from this method (with no need of `runResult`)... just musing over ;)

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/speedsearch/DiscoverAction.java
```diff
@@ -38,51 +39,69 @@ public IEnhancedLambdaLogger log() {
   @UIEffect
   public void actionPerformed(AnActionEvent anActionEvent) {
```
PawelLipski, line 40: This action is going to be rarely tested in everyday use... consider adding a test case for discover to `UITestSuite` (as a first thing the test would need to wipe out or delete machete file)

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/speedsearch/DiscoverAction.java
micpiotrowski: TBH, speed search is probably only place where this action is available, so this is the reason of this change.
But ok, `unanchored` is also ok.

frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/EnhancedGraphTable.java
micpiotrowski: Unfortunately with `Task.Backgroundable` not. I try use a `Task.WithResult` but handling exception when receive the result is awkward (this is how it works in that case).

frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/EnhancedGraphTable.java
micpiotrowski: No, this is about this field is uninitialized by constructor: `error: [[fbc, nullness, allcheckers]:initialization.fields.uninitialized] the constructor does not initialize fields: runResult`

PawelLipski: Ahh right... then maybe mark this field as `@MonotonicNonNull`?

frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/EnhancedGraphTable.java
```diff
@@ -276,6 +272,12 @@ public void onSuccess() {
 
   private static final class EmptyRepositoryException extends Exception {}
   private static final class EmptyBranchLayoutException extends Exception {}
+  @RequiredArgsConstructor(staticName = "of")
```
PawelLipski, line 275: Nit: add empty line before this one

frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/EnhancedGraphTable.java
```diff
@@ -218,51 +216,49 @@ private void doAutomaticDiscover(Path macheteFilePath) {
...
         getString("string.GitMachete.EnhancedGraphTable.automatic-discover.task-title")) {
-      private Try<Tuple3<IBranchLayoutWriter, @Nullable IBranchLayout, IGitMacheteRepositorySnapshot>> runResult;
+      private Try<DiscoverResult> discoverResult;
 
       @Override
       public void run(ProgressIndicator indicator) {
-        runResult = Try.of(() -> RuntimeBinding.instantiateSoleImplementingClass(IGitMacheteRepositoryCache.class)
+        discoverResult = Try.of(() -> RuntimeBinding.instantiateSoleImplementingClass(IGitMacheteRepositoryCache.class)
             .getInstance(mainDirPath, gitDirPath).discoverLayoutAndCreateSnapshot())
             .filterTry(repositorySnapshot -> repositorySnapshot.getRootBranches().size() > 0,
```
PawelLipski, line 229: Hmm how about the following: in case the condition is not satisfied, just `GuiUtils.invokeLater` the `setTextForEmptyTable(...)` piece and return from the method? Similarly for the case where `result.branchLayout == null` below. This way, the intermediate exceptions won't be needed and the entire flow will be likely easier to analyze

frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/EnhancedGraphTable.java
```diff
@@ -208,11 +207,76 @@ private void refreshModel(GitRepository gitRepository, List<String> skippedBranc
...
+            .filterTry(result -> result.branchLayout != null, __ -> new EmptyBranchLayoutException())
+            .andThenTry(
+                result -> result.branchLayoutWriter.write(macheteFilePath, result.branchLayout, /* backupBranchLayout */ true));
+      }
+      // This method is invoked on UI dispatch thread
+      @UIEffect
+      @Override
+      public void onSuccess() {
+        discoverResult.onFailure(exception -> {
+          Class<? extends Throwable> exceptionClass = exception.getClass();
```
PawelLipski, line 245: Hmm no idea why this failing on `FenumChecker` :thinking: anyway, possibly as in the[comment](https://github.com/VirtusLab/git-machete-intellij-plugin/pull/550/files#r478652551) above, the temporary exception classes can be eliminated

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/speedsearch/DiscoverAction.java
mkondratek: CMIIW but I dare timidly suppose that you misunderstand the meaning of `speed search`. 

This is `Find Action...` dialog (?):
image

This is speed search functionality:
image

No action is "available from `speed search`".

PawelLipski: One can argue that speed search is also wired up to Ctrl+Shift+A... as in, IIUC it's "speed search" in the sense that it yields the results as you type (as opposed to how e.g. `grep` operates - only yielding the results once the search pattern is considered complete)

frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/EnhancedGraphTable.java
micpiotrowski: Get rid of this field completely

frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/EnhancedGraphTable.java
micpiotrowski: Out of date. Whole job is done in `run` method.

frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/EnhancedGraphTable.java
micpiotrowski: revamped

frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/EnhancedGraphTable.java
micpiotrowski: revamped

frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/EnhancedGraphTable.java
micpiotrowski: get rid of this

frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/EnhancedGraphTable.java
```diff
@@ -216,67 +213,57 @@ private void doAutomaticDiscover(Path macheteFilePath) {
...
-                exception.getMessage() != null ? exception.getMessage() : "");
-          }
-        })
-            .onSuccess(result -> {
-              VcsNotifier.getInstance(project)
-                  .notifyInfo(getString("string.GitMachete.EnhancedGraphTable.automatic-discover.success-message"));
-
-              gitMacheteRepositorySnapshot = result.repositorySnapshot;
+                exception.getMessage() != null ? exception.getMessage() : ""), NON_MODAL))
+            .onSuccess(v -> {
```
PawelLipski, line 253: Nit:
```suggestion
            .onSuccess(__ -> {
```

frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/EnhancedGraphTable.java
```diff
@@ -216,67 +213,57 @@ private void doAutomaticDiscover(Path macheteFilePath) {
...
+
+        var branchLayoutWriter = project.getService(BranchLayoutWriterProvider.class).getBranchLayoutWriter();
+        var branchLayout = repositorySnapshot.getBranchLayout().getOrNull();
+
+        if (branchLayout == null) {
+          LOG.error("Can't get branch layout from repository snapshot");
+          return;
+        }
+
+        @SuppressWarnings("guieffect") // To allow lambda invocation in `onSuccess` method
```
PawelLipski, line 248: Hmm I'm afraid that the GuiEffect error here is justified :thinking:
See the comment below to `setModel`

frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/EnhancedGraphTable.java
```diff
@@ -216,67 +213,57 @@ private void doAutomaticDiscover(Path macheteFilePath) {
...
+            .onSuccess(v -> {
+              GuiUtils.invokeLaterIfNeeded(
+                  () -> VcsNotifier.getInstance(project)
+                      .notifyInfo(getString("string.GitMachete.EnhancedGraphTable.automatic-discover.success-message")),
+                  NON_MODAL);
+
+              gitMacheteRepositorySnapshot = repositorySnapshot;
               var repositoryGraph = repositoryGraphCache.getRepositoryGraph(gitMacheteRepositorySnapshot, isListingCommits);
               setModel(new GraphTableModel(repositoryGraph));
               repaint();
```
PawelLipski: Hmm shouldn't those lines be executed on UI thread? :thinking:

frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/EnhancedGraphTable.java
micpiotrowski: Probably should be even `___` (triple `_`) coz double `_` is already used in this scope

src/main/resources/META-INF/plugin.xml
```diff
@@ -199,7 +199,7 @@
             from `Search Everywhere` and by link displayed in graph table when `machete` file
             is empty or absent -->
         <action id="GitMachete.DiscoverAction"
-                class="com.virtuslab.gitmachete.frontend.actions.graphtable.DiscoverAction"
+                class="com.virtuslab.gitmachete.frontend.actions.unanchored.DiscoverAction"
```
PawelLipski, line 202: One file to update is also docs/plugins.jetbrains.com/general_usage_instructions.html - there's still a TODO for #176 there :/

frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/EnhancedGraphTable.java
```diff
@@ -176,26 +180,18 @@ private void refreshModel(GitRepository gitRepository, List<String> skippedBranc
     } else {
       repositoryGraph = repositoryGraphCache.getRepositoryGraph(gitMacheteRepositorySnapshot, isListingCommits);
       if (gitMacheteRepositorySnapshot.getRootBranches().isEmpty()) {
-        setTextForEmptyTable(
```
PawelLipski, line 179: Line 162 still has a no longer relevant TODO for #176

PawelLipski:
Taking over from @micpiotrowski...

frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/EnhancedGraphTable.java
PawelLipski: Done, reworked

frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/EnhancedGraphTable.java
PawelLipski: Reworked, no longer relevant

frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/EnhancedGraphTable.java
PawelLipski: Done

mkondratek:
I am not sure if this is an issue but consider the following scenario:

1. Machete file has a valid content (branches exists, properly indented)
2. Remove the whole content
3. Insert `secret` (assuming there is no branch called `secret`) in the first line
4. Refresh table (reload machete changes)

Actual:
`secret` "lost", replaced by the discovered layout

Expected:
A notification saying that `secret` does not belong to the local repo???

mkondratek:
Another case is when there is no local branch withing the repo. Shouldn't refresh show some notification (`"No local branches within the repo"` or smth)? Presently nothing happens

PawelLipski:
> Another case is when there is no local branch withing the repo. Shouldn't refresh show some notification (`"No local branches within the repo"` or smth)? Presently nothing happens

#562 opened

mkondratek (approved):
gj

PawelLipski (approved):

Issue #270, opened 2020-05-09
Add plugin settings under IDE settings


mkondratek:
Lombok plugin has a tab:
![image](https://user-images.githubusercontent.com/19799111/81486159-39ed1100-9253-11ea-9335-99ed1fd90f5e.png)
 We want one too! :running_man: 

Jet Brains Docs: https://www.jetbrains.org/intellij/sdk/docs/basics/settings.html

It would be nice to have custom settings tab. 
It could allow to:
- manage persistent state options
- modify spacing (#87 #206 :trollface:)
- modify colors (?)
- set up git-workflow-related options
- agree to beta features :b: 
- tbd...

PawelLipski:
If ever done, re-enable `buildSearchableOptions` Gradle task (see #1470)

PR #575, opened 2020-09-08
In discover, pick `develop` as a fixed root if `master` isn't present

PawelLipski:
For tests to pass, CLI v2.15.5 (with the same changes applied) must be released

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -34,6 +34,12 @@ jobs:
           name: Compile tests
           command: ./gradlew compileTestJava
       - run:
+          name: '[TEMP] Upgrade git-machete CLI'
+          command: |
+            apt-get update
+            apt-get install --no-install-recommends -y python3-pip
+            pip3 install --upgrade git-machete==2.15.5
+      - run:
           name: Run backend tests
           command: ./gradlew test
       # Unfortunately, wildcards for test result paths aren't supported by CircleCI yet.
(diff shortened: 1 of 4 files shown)


mkondratek:
Hmmm... :thinking: it looks like something that should be mentioned somehow to the user (by event log?), does it not?
Ideally, we could provide a way to specify **main** branch(es) (custom settings tab #270)

PawelLipski:
Hmm @mkondratek I don't think presenting that detail is so important :thinking: anyway, how would the proposed message go?

mkondratek:
> Hmm @mkondratek I don't think presenting that detail is so important  anyway, how would the proposed message go?

Okay then. It could be smth like `Taking (master|develop) as a presumed root` but let's not get to orthodox about notifying "everything".

PawelLipski:
@mkondratek Hmm yeah the problem is that discover might as well deem some other branches as roots. 
`master` (or `develop`, if `master` missing) are _fixed roots_ aka _roots no matter what_... but other branches may also be considered roots if there's no clear parent to attach them under.
So... I don't what to expose too many details to the user, the discover algorithm is already convoluted enough (and yet still far from perfect).

.circleci/config.yml
```diff
@@ -33,6 +33,12 @@ jobs:
       - run:
           name: Compile tests
           command: ./gradlew compileTestJava
+      - run:
+          name: '[TEMP] Upgrade git-machete CLI'
```
mkondratek, line 37: What is to be done to remove that? We need to rebuild and push the docker image, right?

PawelLipski: image

micpiotrowski (approved):

.circleci/config.yml
micpiotrowski: Image regenerated as `gitmachete/intellij-plugin-ci:2.0.0`

PR #587, opened 2020-09-18
Force repository and Vfs update after current branch reset

micpiotrowski:
... as original IntelliJ dialog implementation do


PawelLipski:
Whooops... WTF is this binary incompatibility, lemme see...

PawelLipski:
(WTF since it doesn't fail on `fix/opened-machete-file-update` branch)

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseResetBranchToRemoteAction.java
```diff
@@ -237,6 +240,10 @@ public void run(ProgressIndicator indicator) {
             log().error(result.getErrorOutputAsJoinedString());
             VcsNotifier.getInstance(project).notifyError(VCS_NOTIFIER_TITLE, result.getErrorOutputAsHtmlString());
           }
+
+          var repositoryRoot = getMainDirectory(gitRepository);
+          GitRepositoryManager.getInstance(project).updateRepository(repositoryRoot);
+          GitUtil.refreshVfsInRoot(repositoryRoot);
```
PawelLipski, line 246: Looks that this method isn't there in 201.x yet... and it looks it can be easily replaced with `VfsUtil.markDirtyAndRefresh`:

```
public static void refreshVfsInRoot(@NotNull VirtualFile root) {
    RefreshVFsSynchronously.trace("refresh root " + root);
    VfsUtil.markDirtyAndRefresh(false, true, false, root);
  }
```

PawelLipski (approved):

PR #588, opened 2020-09-19
Consider `main` branch as a fixed root in discovery alongside `master` and `develop`

PawelLipski:


frontendUiRootImpl/build.gradle
```diff
@@ -13,3 +13,4 @@ dependencies {
 }
 
 addIntellijToCompileClasspath(withGit4Idea: true)
+applyI18nFormatterChecker()
```
PawelLipski, line 16: Since there's a dep on `project(':frontendResourceBundles')`

backendImpl/src/test/resources/reference-cli-version.properties
```diff
@@ -1 +1 @@
-referenceCliVersions=2.15.5
+referenceCliVersions=2.15.5,2.15.6
```
PawelLipski, line 1: Not yet released, but will be soon...

PR #605, opened 2020-10-02
Fix buggy file parsing when the last line of machete file contain only whitespace(s)

micpiotrowski:


PawelLipski (approved):
Tested locally, works fine :)

mkondratek (approved):

PR #610, opened 2020-10-04
Cache Gradle dependencies via CI instead of Docker image

PawelLipski:


PawelLipski:
Ok, saving the full `~/.gradle/` directory takes ~90 seconds...

image

PawelLipski:
...and restoring (1.4GB) takes ~25 seconds, actually mostly spent not on downloading, but extracting:

image

mkondratek (approved):

PR #616, opened 2020-10-10
Reword the override forkpoint dialog main sentence

micpiotrowski:


frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/dialogs/OverrideForkPointDialog.kt
```diff
@@ -40,10 +40,18 @@ class OverrideForkPointDialog
 
   override fun createCenterPanel() =
       panel {
-        row(
-            getString(
-                "action.GitMachete.BaseOverrideForkPointAction.dialog.override-fork-point.label.branch-name")) {
+        row {
```
micpiotrowski, line 43: I must done it in that way, coz when I try to use format with `<html></html>` tags this dialog cuts the last word  :cry:

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/dialogs/OverrideForkPointDialog.kt
```diff
@@ -40,10 +40,18 @@ class OverrideForkPointDialog
...
+          label(
+              format(
+                  getString(
+                      "action.GitMachete.BaseOverrideForkPointAction.dialog.override-fork-point.label.branch-name.1"),
+                  branch.name),
+              bold = false)
           label(branch.name, bold = true)
```
mkondratek, line 50: Can you show some ss?

PawelLipski: yeah, screenshot would be useful

micpiotrowski: This is how it looks like now:
Screenshot from 2020-10-12 12-28-11
Unfortunately there are spaces between each `label()`

When I use the string like `<html>Select after which commit the range of commits belonging to <b>{0}</b> branch starts</html>` and code like `row(format(getString("action.GitMachete.BaseOverrideForkPointAction.dialog.override-fork-point.label.branch-name"), branch.name)) {}`, then it looks like that:
Screenshot from 2020-10-12 12-31-15
The last word is cut.

When we sightly stretch the window, then this word appears:
Screenshot from 2020-10-12 12-31-24

First I think that this is an error with computation of rendered string width when we use a bold formatting, but even without `<b></b>` the last word is cut:
Screenshot from 2020-10-12 12-32-57

Finally this seems to be connected with `<html></html>` tags. When I use string like `Select after which commit the range of commits belonging to {0} branch starts` it seems to be rendered correctly:
Screenshot from 2020-10-12 12-40-47
But we lose a branch name emphasize that IMHO is a good (and important) thing here.

PawelLipski: Wooow that's weird indeed :hushed:
Let's ship as is then!

PawelLipski (approved):

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/dialogs/OverrideForkPointDialog.kt
mkondratek: Alright then, you could take a look and try to find some other way to display text like this (some substitute of label)

For reference:
https://jetbrains.org/intellij/sdk/docs/user_interface_components/kotlin_ui_dsl.html

micpiotrowski: Great! Wrapping labels into single cell was enough.

micpiotrowski:
Now it looks like this:
Screenshot from 2020-10-13 12-49-56

PawelLipski (approved):

mkondratek (approved):

Issue #620, opened 2020-10-12
Every async'ly executing method must return a Future-like object


PawelLipski:
This could prevent errors like https://github.com/VirtusLab/git-machete-intellij-plugin/pull/619/files (and possibly more serious ones) from ever happening (since the programmer would instantly see which method executes some stuff in the background).

Callback hell as e.g. around `com.virtuslab.gitmachete.frontend.ui.impl.table.EnhancedGraphTable#refreshModel` is becoming hard to maintain.

Instead of directly invoking `com.intellij.openapi.progress.Task#queue` we should use some abstraction that would return a handle to the async computation (most likely, some kind of `Future`/`Promise` - TBD what's currently the best choice for us, possibly just Vavr Future since we use Vavr already + it has a familiar Scala-like API).

PawelLipski:
Let's first try rewriting a single Task.Backgroundable to Futures as a PoC to make sure it really makes sense first!

I still take into account the possible outcome of _sorry, we need to leave stuff as is_ ☹️

azdrojowa123:
I tried to rewrite single Task.Backgroundable and I ran into a few problems that are hard for me to jump to currently e.g. somehow override onSucess/onThrowable from single Task.Backgroundable (means that default onSuccess should be in single Task.Backgroundable and from time to time override it in specific situation), custom annotations on callbacks e.g. @UIEffect, wrap queue() method to something different because queue() return void (only wrapping is not a big deal but propagating exceptions from run() method into this wrapped one will be probably a problem).

PawelLipski:
Closing, won't fix

PR #621, opened 2020-10-15
Rediscover suggestion 

mkondratek:
Present look of the dialog:
image

TODO:
- [x] Redicover schedule
- [x] Do not ask again until the project (/IDE) reopening
  - the dialog will show up again after project reopening if declined 
- [x] Dialog does not when the machete tab is opened on the project opening
  - ~WONT FIX: there is no simple way to subscribe for tool window opening :disappointed: anyway IMHO depending on the git machete tab opening is enough~ wired up within graph table (instead of tab open) 
- [x] Hmmm.... would it be easily possible to show the branch layout before while the rediscover dialog is shown? As for now, when the rediscover dialog is showing, there's nothing in the graph table thinking

TESTING:
https://askubuntu.com/questions/62492/how-can-i-change-the-date-modified-created-of-a-file
`touch -d "20 days ago" .git/machete`


frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/GitMacheteRepositoryDiscoverer.java
```diff
@@ -0,0 +1,86 @@
...
+@CustomLog
+public class GitMacheteRepositoryDiscoverer {
+
+  private final Project project;
+  private final IGitRepositorySelectionProvider gitRepositorySelectionProvider;
+  private final Consumer<Path> onFailurePathConsumer;
+  private final Consumer<IGitMacheteRepositorySnapshot> onSuccessRepositoryConsumer;
+
+  @UIEffect
+  public void doDiscover(Path macheteFilePath) {
```
PawelLipski, line 37: Hmm I would call a method that has an async component (`queue()`) using the word `(en)queue...` (or #620 for the state-of-the-art-ish solution)

src/main/resources/META-INF/plugin.xml
```diff
@@ -12,6 +12,7 @@
         <changesViewContent tabName="Git Machete"
                             className="com.virtuslab.gitmachete.frontend.ui.impl.root.GitMacheteContentProvider"
                             predicateClassName="com.virtuslab.gitmachete.frontend.ui.impl.root.GitMacheteVisibilityPredicate"/>
+        <postStartupActivity implementation="com.virtuslab.gitmachete.frontend.ui.impl.root.GitMacheteStartupActivity"/>
```
PawelLipski, line 15: Nice! So this will run after:
1. the plugin classes are loaded
or later when
2. the tab is opened?

frontendUiRootImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/root/GitMacheteStartupActivity.java
```diff
@@ -0,0 +1,38 @@
...
+
+    if (toolWindow == null) {
+      LOG.debug("VCS tool window does not exist");
+      VcsNotifier.getInstance(project).notifyWarning(
+          getString("action.GitMachete.OpenMacheteTabAction.notification.title.could-not-open-tab"), // t0d0: proper msg here
+          getString("action.GitMachete.OpenMacheteTabAction.notification.message.no-git")); // t0d0: proper msg here
+      return;
+    }
+
+    var listener = new RediscoverSuggester(project);
```
PawelLipski, line 35: Hmm I run IDE against the project where .git/machete has been modified in July:
```
$ stat .git/machete
  File: .git/machete
  Size: 142       	Blocks: 8          IO Block: 4096   regular file
Device: fd03h/64771d	Inode: 16259565    Links: 1
Access: (0664/-rw-rw-r--)  Uid: ( 1000/pawellipski)   Gid: ( 1000/pawellipski)
Context: unconfined_u:object_r:user_home_t:s0
Access: 2020-10-15 21:49:28.396911004 +0200
Modify: 2020-07-21 17:03:20.211654006 +0200
Change: 2020-07-21 17:03:20.211654006 +0200
 Birth: 2020-07-21 17:03:20.210654002 +0200
```
but nothing really showed up... or is the core functionality not ready yet? :thinking:

src/main/resources/META-INF/plugin.xml
mkondratek: this one will run on classes loading and add a listener for tab open

frontendUiRootImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/root/GitMacheteStartupActivity.java
mkondratek: whoops, strange :thinking: did you try closing and reopening git machete tab :xd:

PawelLipski: Hmm this isn't even a heisenbug, it just doesn't seem to work (I tested on a few other old projects)... would be nice if you put some logs in the related codebase that I could send you :thinking:

frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/GitMacheteRepositoryDiscoverer.java
PawelLipski: So `do...` is discouraged coz it sounds like sth synchronous :/

frontendUiRootImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/root/GitMacheteStartupActivity.java
mkondratek: I do perform some logging here and there :deciduous_tree: can you retry?

PawelLipski: Okay! working now

PawelLipski: Hmmm.... would it be easily possible to show the branch layout before while the rediscover dialog is shown? As for now, when the rediscover dialog is showing, there's nothing in the graph table :thinking:

frontendResourceBundles/src/main/resources/GitMacheteBundle.properties
```diff
@@ -239,6 +239,9 @@ string.GitMachete.EnhancedGraphTable.empty-table-text.loading=Loading...
 string.GitMachete.EnhancedGraphTable.empty-table-text.only-skipped-in-machete-file=None of the branches provided by machete file ({0}) exists within the local repository
 string.GitMachete.EnhancedGraphTable.skipped-branches-text=The following branches defined by machete file do not belong to the local repository:
 
+string.GitMachete.RediscoverSuggester.dialog.title=Git Machete: Rediscover Suggester
```
PawelLipski, line 242: `Suggester` sounds rather odd... maybe just `Suggestion`

frontendResourceBundles/src/main/resources/GitMacheteBundle.properties
```diff
@@ -239,6 +239,9 @@ string.GitMachete.EnhancedGraphTable.empty-table-text.loading=Loading...
...
+string.GitMachete.RediscoverSuggester.dialog.question=It looks like you have not modified your git machete file for a while. Would you like to rediscover the branch layout?
```
PawelLipski, line 243: I'd say `It looks that...` rather than `like`

frontendUiRootImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/root/RediscoverSuggester.java
```diff
@@ -0,0 +1,93 @@
...
+
+    var lastModifiedTimeMillis = getFileModificationDate(macheteFilePath).getOrNull();
+    if (lastModifiedTimeMillis == null) {
+      LOG.warn("Cannot proceed rediscover suggestion workflow - could not get file modification date");
+      return;
+    }
+
+    var daysDiff = daysDiffTillNow(lastModifiedTimeMillis);
+    // TODO (#270): a candidate for custom settings tab
+    if (daysDiff > 14) { // magic val
```
PawelLipski, line 60: Maybe extract to a constant in this file...

frontendUiRootImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/root/RediscoverSuggester.java
```diff
@@ -0,0 +1,93 @@
...
+      var yesNo = MessageDialogBuilder.YesNo.yesNo(
+          getString("string.GitMachete.RediscoverSuggester.dialog.title"),
+          getString("string.GitMachete.RediscoverSuggester.dialog.question"));
+
+      switch (yesNo.show()) {
+        case Messages.YES :
+          LOG.info("Branch layout has not been modified within given time - enqueueing rediscover");
```
PawelLipski, line 67: ...and use in the logs `as `within ${...} days`

frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/EnhancedGraphTable.java
```diff
@@ -216,71 +210,39 @@ private void refreshModel(
...
-              getString("string.GitMachete.EnhancedGraphTable.automatic-discover.cant-discover-layout-error-title"),
-              exception.getMessage() != null ? exception.getMessage() : ""), NON_MODAL);
-        }
-      }
-    }.queue();
+  @UIEffect
+  private Consumer<Path> getUnsuccessfulDiscoverMacheteFilePathConsumer() {
+    return (Path macheteFilePath) -> GuiUtils.invokeLaterIfNeeded(
+        () -> setTextForEmptyTable(
+            format(getString("string.GitMachete.EnhancedGraphTable.empty-table-text.cant-discover-layout"),
```
PawelLipski, line 243: Nit: I can see there's one message string that uses `cannot-open` rather than `cant`... pls make consistent

frontendUiRootImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/root/GitMacheteStartupActivity.java
PawelLipski: (maybe the dialog can be non-modal?)

PawelLipski: @mkondratek what's the resolution on the above? (https://github.com/VirtusLab/git-machete-intellij-plugin/pull/621#discussion_r506924768)

mkondratek: done

PawelLipski:
@mkondratek hmmm still getting the following:

image

so the graph table shows `Loading...` instead of the current layout when the dialog is shown... is it the same for you?

Maybe the dialog could be instead in the same non-modal way as the below `Install Dependencies` prompt?

image

PawelLipski:
Hmmm the dialog doesn't seem to be opening for me at all, even on some ancient projects I haven't opened in months... I can't spend any extra time on this today, lemme revisit this PR tmrw...

PawelLipski:
Actually... given that this functionality is rarely used (and hence won't be naturally dogfooded over time like e.g. push or rebase), maybe extending the existing `com.virtuslab.gitmachete.uitest.UITestSuite#discoverBranchLayout` with a call to `java.io.File#setLastModified` + then clicking the button on the dialog (as in `Project.discoverBranchLayout` in uiTests/src/test/resources/project.rhino.js) would be a good idea :thinking:

frontendVfsUtils/src/main/java/com/virtuslab/gitmachete/frontend/vfsutils/GitVfsUtils.java
```diff
@@ -50,4 +53,13 @@ public static Path getMainDirectoryPath(GitRepository gitRepository) {
   public static Path getMacheteFilePath(GitRepository gitRepository) {
     return getGitDirectoryPath(gitRepository).resolve(MACHETE_FILE_NAME);
   }
+
+  /**
+   * @param filePath {@link Path} to file
+   * @return an option of {@link Long} stating for time of last modification in milliseconds
```
PawelLipski, line 59: For clarity: `milliseconds since the Unix epoch start`? if that's the intended semantics ofc

frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/RediscoverSuggester.java
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  public void performIfNotDeclined() {
+    if (wasRediscoverSuggestionDeclined) {
+      return;
+    }
+
+    var macheteFilePath = Option.of(gitRepository).map(GitVfsUtils::getMacheteFilePath).getOrNull();
+    if (macheteFilePath == null) {
+      LOG.warn("Cannot proceed rediscover suggestion workflow - selected machete file is null");
```
PawelLipski, line 46: `proceed with`

frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/RediscoverSuggester.java
```diff
@@ -0,0 +1,98 @@
...
+      return;
+    }
+
+    var lastModifiedTimeMillis = getFileModificationDate(macheteFilePath).getOrNull();
+    if (lastModifiedTimeMillis == null) {
+      LOG.warn("Cannot proceed rediscover suggestion workflow - could not get file modification date");
```
PawelLipski, line 52: nit, ditto

uiTests/src/test/resources/project.rhino.js
```diff
@@ -24,11 +24,15 @@ function Project(underlyingProject) {
   };
 
   // Tab & model management
-  this.openTab = function () {
+  this.openGitMacheteTab = function () {
+    this.openTab(ToolWindowId.VCS, 'Git Machete');
+  }
+
+  this.openTab = function (toolWindowId, tabName) {
```
PawelLipski, line 31: Nit: I guess some of these method could be "private" (`const foo = function...`)

uiTests/src/test/scala/com/virtuslab/gitmachete/uitest/UITestSuite.scala
```diff
@@ -166,4 +173,9 @@ class UITestSuite extends BaseGitRepositoryBackedIntegrationTestSuite(SETUP_WITH
   private def overwriteMacheteFile(content: String): Unit = {
     repositoryGitDir.resolve("machete").write(content + "\n")
   }
+
+  private def setLastModifiedDateOfMacheteFileToZero(): Unit = {
```
PawelLipski, line 177: Nit: maybe `....ToEpochStart` instead of `...ToZero` for clarity (?)

uiTests/src/test/scala/com/virtuslab/gitmachete/uitest/UITestSuite.scala
```diff
@@ -1,11 +1,11 @@
...
+import java.nio.file.Files
+import java.nio.file.attribute.FileTime
+
 import com.virtuslab.gitmachete.testcommon.BaseGitRepositoryBackedIntegrationTestSuite
 import com.virtuslab.gitmachete.testcommon.BaseGitRepositoryBackedIntegrationTestSuite.SETUP_WITH_SINGLE_REMOTE
-import org.junit.After
-import org.junit.Assert
-import org.junit.Before
-import org.junit.Test
+import org.junit.{After, Assert, Before, Test}
```
PawelLipski, line 8: While you're here... pls add a mention to development-guide.md that installing Scala plugin might be useful for editing UI tests

uiTests/src/test/scala/com/virtuslab/gitmachete/uitest/UITestSuite.scala
PawelLipski: ...or whatever this zero means here

frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/RediscoverSuggester.java
```diff
@@ -0,0 +1,98 @@
...
+  private final GitRepository gitRepository;
+
+  private final Runnable discoverOperation;
+
+  // TODO (#270): a candidate for custom settings tab
+  private final int DAYS_AFTER_WHICH_TO_SUGGEST_DISCOVER = 14;
+
+  private static boolean wasRediscoverSuggestionDeclined = false;
+
+  public RediscoverSuggester(Project project, GitRepository gitRepository, Runnable discoverOperation) {
```
PawelLipski, line 33: Nit: `@RequiredArgsConstructor`?

frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/GitMacheteRepositoryDiscoverer.java
```diff
@@ -0,0 +1,84 @@
...
+          LOG.error("Can't get branch layout from repository snapshot");
+          return;
+        }
+
+        try {
+          branchLayoutWriter.write(macheteFilePath, branchLayout, /* backupOldLayout */ true);
+          onSuccessRepositoryConsumer.accept(repositorySnapshot);
+        } catch (BranchLayoutException exception) {
+          GuiUtils.invokeLaterIfNeeded(() -> VcsNotifier.getInstance(project).notifyError(
+              getString("string.GitMachete.EnhancedGraphTable.automatic-discover.cannot-discover-layout-error-title"),
```
PawelLipski, line 78: Nit: typically `title` was separated by a dot `.` in our property keys

PawelLipski (changes requested):
Tested locally, works perfectly :rocket: 
Minor comments, almost ready to approve

micpiotrowski:
I'm afraid for some users this dialog can be annoying (e.g. when they don't need to use our plugin for some period of time in selected project). So maybe checkbox `Don't show again` can be a good idea?

.circleci/config.yml
```diff
@@ -116,7 +116,7 @@ jobs:
             equal: [ master, << pipeline.git.branch >> ]
           steps:
             - deploy:
-                # Publishing should be the first step since if it fails, we should NOT proceed opening backport PR/GitHub release etc.
+                # Publishing should be the first step since if it fails, we should NOT proceed with opening backport PR/GitHub release, etc.
                 name: Publish plugin to Jetbrains Marketplace
                 command: ./gradlew --max-workers=2 publishPlugin
```
PawelLipski, line 121: While you're here... pls also remove `--max-workers=2`, this should no longer be necessary (I hope!)

frontendResourceBundles/src/main/resources/GitMacheteBundle.properties
```diff
@@ -230,14 +230,18 @@ string.GitMachete.BranchOrCommitCellRendererComponent.sync-to-remote-status-text
...
+string.GitMachete.EnhancedGraphTable.automatic-discover.notification.title.cannot-discover-layout-error=Can't automatically discover repository layout
 string.GitMachete.EnhancedGraphTable.automatic-discover.task-title=Automatically discovering branch layout
 string.GitMachete.EnhancedGraphTable.automatic-discover.success-message=Branch layout has been automatically discovered
-string.GitMachete.EnhancedGraphTable.empty-table-text.cant-discover-layout=Provided machete file ({0}) is empty and branch layout can''t be automatically detected
+string.GitMachete.EnhancedGraphTable.empty-table-text.cannot-discover-layout=Provided machete file ({0}) is empty and branch layout can''t be automatically detected
 string.GitMachete.EnhancedGraphTable.empty-table-text.loading=Loading...
 string.GitMachete.EnhancedGraphTable.empty-table-text.only-skipped-in-machete-file=None of the branches provided by machete file ({0}) exists within the local repository
 string.GitMachete.EnhancedGraphTable.skipped-branches-text=The following branches defined by machete file do not belong to the local repository:
 
+string.GitMachete.RediscoverSuggester.task-title="Suggesting rediscover..."
```
PawelLipski, line 241: Hmm it looks like the sole non-commented occurrence of double quotes in the entire file... is this needed? If not, then also consider patching scripts/enforce-property-single-quote-in-bundle-message to forbid them

PawelLipski: (or add a separate script)

uiTests/src/test/resources/project.rhino.js
```diff
@@ -109,12 +113,19 @@ function Project(underlyingProject) {
...
+    findAndClickButton('Save');
+  }
+
+  this.acceptSuggestedBranchLayout = function () {
+    findAndClickButton('Yes');
+  };
 
-    const getSaveButton = function() {
+  const findAndClickButton = function (name) {
+    const getSaveButton = function () {
```
PawelLipski, line 124: The `Save` part of the name is no longer relevant

frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/RediscoverSuggester.java
```diff
@@ -0,0 +1,91 @@
...
+  public void performIfNotDeclined() {
+    var macheteFilePath = Option.of(gitRepository).map(GitVfsUtils::getMacheteFilePath).getOrNull();
+    if (macheteFilePath == null) {
+      LOG.warn("Cannot proceed with rediscover suggestion workflow - selected machete file is null");
+      return;
+    }
+
+    var lastModifiedTimeMillis = getFileModificationDate(macheteFilePath).getOrNull();
+    if (lastModifiedTimeMillis == null) {
+      LOG.warn("Cannot proceed with rediscover suggestion workflow - could not get file modification date");
```
PawelLipski, line 45: Seems that `string.GitMachete.RediscoverSuggester.failed-to-get-file-attrs` is unused... not sure about other strings marked as unused by IntelliJ (maybe false negatives? as for the `*.GitMacheteToolbar.*` ones):

image

PawelLipski:
Hmm @mkondratek now (as of `0.8.1-42-SNAPSHOT+git.21ea602`) it seems that sometimes the dialog opens even if the tab isn't opened... not sure if intended, can be surprising for the user, esp. if they can't see what's the current branch layout:

image

If you can't reproduce locally, I can send some logs from my machine

frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/RediscoverSuggester.java
```diff
@@ -0,0 +1,91 @@
...
+  private final Project project;
+
+  private final GitRepository gitRepository;
+
+  private final Runnable discoverOperation;
+
+  // TODO (#270): a candidate for custom settings tab
+  private final int DAYS_AFTER_WHICH_TO_SUGGEST_DISCOVER = 14;
+
+  public void performIfNotDeclined() {
```
PawelLipski, line 36: Is the `IfNotDeclined` part still relevant? :thinking:

PawelLipski:
And the other problem... after clicking `No`, sometimes the dialog re-opens...

Generally... maybe the better approach instead of a dialog on the center of the screen would be to open a more subtle dialog somewhere in the right-bottom corner of the IDE? As it's with IIRC whether to add file to Git index :thinking:

scripts/prohibit-property-double-quote-wrapping-in-bundle-message
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+
+set -o pipefail -u
+
+self_dir=$(cd "$(dirname "$0")" &>/dev/null; pwd -P)
+source "$self_dir"/utils.sh
+
+
+# Find the strings that have the property value wrapped in double quotes (")
+values_wrapped_in_double_quotes=$(git grep -Ph -e '=".*"$' -- '*.properties')
```
PawelLipski, line 10: Nit: no need for Perl syntax `-P` since `.`, `*`, `$` are basic regex syntax

uiTests/src/test/scala/com/virtuslab/gitmachete/uitest/UITestSuite.scala
```diff
@@ -64,6 +64,7 @@ class UITestSuite extends BaseGitRepositoryBackedIntegrationTestSuite(SETUP_WITH
   }
 
   @Test def skipNonExistentBranches_toggleListingCommits_slideOutRoot(): Unit = {
+    intelliJ.machete.runJs(s"project.openGitMacheteTab()")
```
PawelLipski, line 67: Maybe drop a comment on why this was needed... if there's a clear reason, at least

PawelLipski: Also, nit: `scala.StringContext#s` not needed since no interpolation happens inside

frontendUiRootImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/RediscoverSuggester.java
```diff
@@ -56,31 +51,25 @@ public void performIfNotDeclined() {
     }
   }
 
+  @UIEffect
   private void queueSuggestion(Path macheteFilePath) {
```
PawelLipski, line 55: I don't think anything is "queued" here anymore, it's a `@UIEffect` method, hence it's meant to execute synchronously

frontendUiRootImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/root/GitMachetePanel.java
```diff
@@ -41,6 +47,19 @@ public GitMachetePanel(Project project) {
     setToolbar(createGitMacheteVerticalToolbar(graphTable).getComponent());
     add(createShrinkingWrapper(selectionComponent), BorderLayout.NORTH);
     setContent(ScrollPaneFactory.createScrollPane(graphTable));
+
+    addAncestorListener(new AncestorListenerAdapter() {
+      @Override
+      public void ancestorAdded(AncestorEvent event) {
```
PawelLipski, line 53: Maybe drop a comment about what's the specific event that we aim to capture here

frontendUiRootImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/RediscoverSuggester.java
PawelLipski: Oh ok... see my comment below

frontendUiRootImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/RediscoverSuggester.java
```diff
@@ -56,31 +51,25 @@ public void performIfNotDeclined() {
...
-              LOG.info("Unknown response message");
-              break;
-          }
-        }, ModalityState.NON_MODAL);
-      }
-    }.queue();
+    switch (yesNo.show()) {
+      case Messages.YES :
+        LOG.info("Enqueueing rediscover");
+        discoverOperation.run();
```
PawelLipski, line 63: Hmmm I'd rename it to `queueDiscoverOperation` to signalize that it's going to have an asynchronous part; then the entire method can remain `queueSuggestion`

PawelLipski (changes requested):
Tested locally, works fine :spinning-parrot:
Just minor comments left

frontendUiRootImpl/build.gradle
```diff
@@ -6,6 +6,7 @@ dependencies {
   implementation project(':binding')
```
PawelLipski, line 6: Btw (unrel) if you could bump 2020.3 EAP version in build.gradle to the recent Beta as per https://www.jetbrains.com/intellij-repository/snapshots/ ... ofc if it fails for any reason then let's investigate separately

uiTests/src/test/scala/com/virtuslab/gitmachete/uitest/UITestSuite.scala
mkondratek: > Maybe drop a comment on why this was needed... if there's a clear reason, at least

I do not think that any comment in code is necessary.
Furthermore, I see no right place for such a comment. 
However, I will give an explanation here.

We are talking about this fragment of code:
```
@Test def discoverBranchLayout(): Unit = {
    // When model is refreshed and machete file is has not been modified for a long time, then discover suggestion should occur
    setLastModifiedDateOfMacheteFileToEpochStart()
    intelliJ.machete.runJs("project.openGitMacheteTab()")
    //...
  }
```

The point is that to test the discover suggestion functionality we must prepare the test (set the modified date of the file) before the check (of the time last) occurs, obviously.

The "problem" that appeared is that the check happens on the component creation (at the moment when intellij content manager "adds" (displays) `GitMachetePanel` actually). Hence we must prepare for the test before the first tab opening. To do this I had to extract the tab opening from before each method.

frontendUiRootImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/root/GitMachetePanel.java
mkondratek: done, here is an example event

image

PawelLipski (approved):
:tada:

PR #627, opened 2020-10-21
Fix code for opening backport PRs

PawelLipski:


.circleci/config.yml
```diff
@@ -140,11 +140,10 @@ jobs:
                   git checkout -b "$branch"
                   git push origin "$branch"
 
-                  # In case of a PR build, CI_PULL_REQUEST should be a link of the form https://github.com/VirtusLab/git-machete-intellij-plugin/pull/123
```
PawelLipski, line 143: This code is only executed for `master` builds... hence `CI_PULL_REQUEST` in undefined

mkondratek (approved):

frontendResourceBundles/src/main/resources/GitMacheteBundle.properties
```diff
@@ -51,7 +51,7 @@ action.GitMachete.BaseOverrideForkPointAction.description-action-name=Fork point
 # {0} skipped intentionally (parent branch name is not needed)
 action.GitMachete.BaseOverrideForkPointAction.description=Override fork point of ''{1}''
 action.GitMachete.BaseOverrideForkPointAction.dialog.override-fork-point.label.branch-name.1=Select after which commit the range of commits belonging to
-action.GitMachete.BaseOverrideForkPointAction.dialog.override-fork-point.label.branch-name.2=branch starts
+action.GitMachete.BaseOverrideForkPointAction.dialog.override-fork-point.label.branch-name.2=branch starts:
```
mkondratek, line 54: while we are here...
the suffixes `branch-name.1` and `branch-name.2` are relics. 
In the first place, they were actually containing branch names IIRC but now they do not make much sense

mkondratek: image

PawelLipski: Ok, renamed

PR #628, opened 2020-10-21
Fix graph scaling (closes #626)

### LINKED ISSUE #626: Graph does not scale along with font size

mkondratek:


frontendGraphImpl/src/main/java/com/virtuslab/gitmachete/frontend/graph/impl/paint/GraphCellPainter.java
```diff
@@ -29,7 +29,12 @@
 
   @UIEffect
   protected int getRowHeight() {
-    return table.getRowHeight();
+    var font = table.getFont();
```
PawelLipski, line 32: Pls add a comment in the code:
1. when (semantically) the font is null and when not (if we know that)
2. that the purpose is to make graph scale alongside the default font from IntelliJ Settings

PawelLipski:
Tested locally, works fine for me:

image

PawelLipski (approved):

PawelLipski (approved):

PawelLipski:
@mkondratek pls bump version.gradle to 0.8.1

PR #637, opened 2020-10-29
Switch to runPluginVerifier Gradle task native to gradle-intellij-plugin v0.6+

PawelLipski:


mkondratek (approved):
+1 
image

.circleci/config.yml
```diff
@@ -97,17 +97,17 @@ jobs:
           command: ./scripts/verify-artifact-contents
```
mkondratek, line 97: while we are here, line 121:
Publishing should be the first step since if it fails, we should NOT proceed **with** opening backport PR/GitHub **release, etc**.

mkondratek (approved):

.circleci/config.yml
PawelLipski: Whoops... didn't notice, could you fix this on your open PR?

mkondratek: fine

PR #655, opened 2020-12-04
[frontend][graph-table] Prettify a graph table due to highlighting in IntelliJ 2020.3

micpiotrowski:
Add right padding for graph table row width and disable row full space filling when right click (for consistency)

`2020.3`:

`2020.2`:

### DIFF
--- a/frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/EnhancedGraphTable.java
+++ b/frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/EnhancedGraphTable.java
@@ -40,7 +40,6 @@ import com.intellij.openapi.vfs.VfsUtil;
 import com.intellij.ui.GuiUtils;
 import com.intellij.ui.PopupMenuListenerAdapter;
 import com.intellij.ui.ScrollingUtil;
-import com.intellij.ui.table.JBTable;
 import com.intellij.util.messages.Topic;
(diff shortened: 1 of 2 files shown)


mkondratek:
pics or it didn't happen

frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/EnhancedGraphTable.java
```diff
@@ -370,11 +369,7 @@ public void popupMenuWillBecomeVisible(PopupMenuEvent popupMenuEvent) {
       new Timer().schedule(new TimerTask() {
         @Override
         public void run() {
-          GuiUtils.invokeLaterIfNeeded(() -> {
```
PawelLipski, line 373: Not that we should care much about 2020.2 at this point... but how does it affect the old IntelliJ versions?

PawelLipski:
image

frontendUiTableImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/EnhancedGraphTable.java
micpiotrowski: The selection will has a smaller width (not the size of entire pane) but this still should look OK (just like the highlighting in 2020.3)

PawelLipski (approved):
:ok_hand:

mkondratek (approved):

PR #663, opened 2020-12-11
Add a subtitle parameter for FetchBackgroundable

micpiotrowski:
Screenshot from 2020-12-11 17-27-46
Screenshot from 2020-12-11 17-28-01

### DIFF
--- a/frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/backgroundables/FetchBackgroundable.java
+++ b/frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/backgroundables/FetchBackgroundable.java
@@ -10,6 +10,7 @@ import git4idea.repo.GitRemote;
 import git4idea.repo.GitRepository;
 import lombok.CustomLog;
 import org.checkerframework.checker.guieffect.qual.UIEffect;
+import org.checkerframework.checker.nullness.qual.Nullable;
 
 @CustomLog
 public class FetchBackgroundable extends Task.Backgroundable {
@@ -20,6 +21,7 @@ public class FetchBackgroundable extends Task.Backgroundable {
   private final String refspec;
   private final String failureNotificationText;
   private final String successNotificationText;
+  private final @Nullable String taskSubtitle;
 
   /** Use as {@code remoteName} when referring to the local repository. */
   public static final String LOCAL_REPOSITORY_NAME = ".";
@@ -30,7 +32,8 @@ public class FetchBackgroundable extends Task.Backgroundable {
       String refspec,
       String taskTitle,
       String failureNotificationText,
-      String successNotificationText) {
+      String successNotificationText,
+      @Nullable String taskSubtitle) {
     super(project, taskTitle, /* canBeCancelled */ true);
     this.project = project;
     this.gitRepository = gitRepository;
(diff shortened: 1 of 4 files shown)


frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/backgroundables/FetchBackgroundable.java
```diff
@@ -30,18 +32,35 @@ public FetchBackgroundable(Project project,
       String refspec,
       String taskTitle,
       String failureNotificationText,
-      String successNotificationText) {
+      String successNotificationText,
+      @Nullable String taskSubtitle) {
```
PawelLipski, line 36: Hmm this is `null` only for the Pull task currently? or anything else?

PawelLipski: If only for pull, then maybe you could give `Pulling...` there instead of the default, and the other c'tor isn't even needed

micpiotrowski: No, this is also for Slide-in (in case of remote branch)

PawelLipski: Hmm ok so in case of Slide-in it's still `Fetching...` that shows up right? :thinking:

frontendResourceBundles/src/main/resources/GitMacheteBundle.properties
```diff
@@ -42,6 +42,7 @@ action.GitMachete.BaseFastForwardMergeBranchToParentAction.description=Fast-forw
 action.GitMachete.BaseFastForwardMergeBranchToParentAction.notification.title.ff-fail=Fast-forward merge branch ''{1}'' to ''{0}'' failed
 action.GitMachete.BaseFastForwardMergeBranchToParentAction.notification.title.ff-success=Branch <b>{1}</b> has been fast-forward merged to <b>{0}</b>
 action.GitMachete.BaseFastForwardMergeBranchToParentAction.task-title=Fast-forward merging
+action.GitMachete.BaseFastForwardMergeBranchToParentAction.task-subtitle=Fast-forwarding...
```
mkondratek, line 45: or rather `fast-forward merging...` to be consistent with the recent clarification?

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/backgroundables/FetchBackgroundable.java
micpiotrowski: Yes, it's true. IMHO its appropriate message in that case - fetching is what actually this action do.

frontendResourceBundles/src/main/resources/GitMacheteBundle.properties
micpiotrowski: Actually, I'd leave as is, coz the main action title is `Fast-forward merging` and the same subtitle bellow it IMHO can look strange, WDYT?

PawelLipski: Ok, let's leave as is

PawelLipski (approved):

PR #664, opened 2020-12-19
Fix inconsistencies to CLI in discovery algorithm + treat `develop` as root even if `master` present

PawelLipski:
Wrt. "treat both `master` and `develop` as roots by default" part: applied the same changes in the CLI as 2.15.9

### DIFF
--- a/backendImpl/src/main/java/com/virtuslab/gitmachete/backend/impl/GitMacheteRepository.java
+++ b/backendImpl/src/main/java/com/virtuslab/gitmachete/backend/impl/GitMacheteRepository.java
@@ -20,7 +20,6 @@ import io.vavr.collection.Map;
 import io.vavr.collection.Queue;
 import io.vavr.collection.Seq;
 import io.vavr.collection.Set;
-import io.vavr.collection.SortedMap;
 import io.vavr.control.Option;
 import io.vavr.control.Try;
 import lombok.AccessLevel;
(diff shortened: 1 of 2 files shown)


backendImpl/src/main/java/com/virtuslab/gitmachete/backend/impl/GitMacheteRepository.java
```diff
@@ -818,20 +819,18 @@ IGitMacheteRepositorySnapshot discoverLayoutAndCreateSnapshot(int mostRecentlyCh
...
-        nonFixedRootBranchNames = localBranchNames.reject(Predicate.isEqual(DEVELOP));
-      } else {
-        fixedRootBranchNames = List.empty();
-        nonFixedRootBranchNames = localBranchNames;
+        fixedRootBranchNames = fixedRootBranchNames.append(MAIN);
+        nonFixedRootBranchNames = nonFixedRootBranchNames.remove(MAIN);
+      }
+      if (localBranchNames.contains(DEVELOP)) {
+        fixedRootBranchNames = fixedRootBranchNames.append(DEVELOP);
+        nonFixedRootBranchNames = nonFixedRootBranchNames.remove(DEVELOP);
```
PawelLipski, line 833: That's the part when logic is changed: `develop` is treated as root even if `master` (or `main`) is present
This is also reflected in CLI 2.15.8 (but not yet 2.15.7; note that's just an implementation change, hence no minor bump).

backendImpl/src/main/java/com/virtuslab/gitmachete/backend/impl/GitMacheteRepository.java
```diff
@@ -845,18 +844,18 @@ IGitMacheteRepositorySnapshot discoverLayoutAndCreateSnapshot(int mostRecentlyCh
...
-          .toSortedMap(name -> Tuple.of(name, new MyBranchLayoutEntry(name)));
-      SortedMap<String, MyBranchLayoutEntry> entryByFreshNonFixedRootBranch = freshNonFixedRootBranchNames
-          .toSortedMap(name -> Tuple.of(name, new MyBranchLayoutEntry(name)));
-      SortedMap<String, MyBranchLayoutEntry> entryByIncludedBranchName = entryByFixedRootBranchNames
+      // Let's use linked maps to ensure a deterministic result.
+      Map<String, MyBranchLayoutEntry> entryByFixedRootBranchNames = fixedRootBranchNames
+          .toLinkedMap(name -> Tuple.of(name, new MyBranchLayoutEntry(name)));
+      Map<String, MyBranchLayoutEntry> entryByFreshNonFixedRootBranch = freshNonFixedRootBranchNames
+          .toLinkedMap(name -> Tuple.of(name, new MyBranchLayoutEntry(name)));
+      Map<String, MyBranchLayoutEntry> entryByIncludedBranchName = entryByFixedRootBranchNames
```
PawelLipski, line 858: Aaand here's the part where an existing inconsistency with CLI has been fixed... it's pretty subtle so I haven't `discovered` (no pun intended) until I changed the conditions for `develop` being a root

backendImpl/src/test/resources/reference-cli-version.properties
```diff
@@ -1 +1 @@
-referenceCliVersions=2.15.5,2.15.6,2.15.7
+referenceCliVersions=2.15.8
```
PawelLipski, line 1: To be released soon...

backendImpl/src/main/java/com/virtuslab/gitmachete/backend/impl/GitMacheteRepository.java
```diff
@@ -818,20 +817,18 @@ IGitMacheteRepositorySnapshot discoverLayoutAndCreateSnapshot(int mostRecentlyCh
         throws GitCoreException, GitMacheteException {
```
mkondratek, line 817: While we are here... are we okay with throwing these two exceptions?
Of course, this method is not being exposed outside the package but still, it is somewhat messy to throw exceptions of different layers.

mkondratek (approved):

backendImpl/src/main/java/com/virtuslab/gitmachete/backend/impl/GitMacheteRepository.java
PawelLipski: Yeah, I'm definitely not okay with that... but the alternative IIRC was that the exceptions were wrapped in method code, which would leave the code even more messy (a lot of noise added by all those extra `Try`es or `try...catch`es)

PR #665, opened 2020-12-19
Extract Markdown validation to separate, locally runnable scripts

PawelLipski:


.circleci/config.yml
```diff
@@ -25,37 +25,12 @@ jobs:
           command: ./gradlew
 
       - run:
-          name: Validate Markdown ToCs
+          name: Validate Markdown files
           command: |
-            git grep --files-with-matches 'START doctoc' | xargs doctoc
```
mkondratek, line 30: ups... it has become quite a mess :no_good:

mkondratek (approved):

.circleci/config.yml
PawelLipski: yeah, beyond certain threshold keeping scripts directly in CI config rather obfuscates than clarifies the view

PR #673, opened 2020-12-27
Fix NPE in GitCoreRepository#deriveMergeBase

PawelLipski:
Happened due to https://github.com/typetools/checker-framework/issues/4079. Not very harmful since it only happens when `DEBUG` log is turned on AND merge base can't be figured out for some two commits in the repo.

### DIFF
--- a/frontendResourceBundles/src/main/resources/GitMacheteBundle.properties
+++ b/frontendResourceBundles/src/main/resources/GitMacheteBundle.properties
@@ -29,6 +29,7 @@ action.GitMachete.GitCommandUpdatingCurrentBranchBackgroundable.notification.tit
 
 # Quasi-title capitalization intended since we always write "Git Machete" with initial caps.
 action.GitMachete.GitMacheteRepositoryUpdateBackgroundable.task-title=Updating Git Machete status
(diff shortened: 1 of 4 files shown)


micpiotrowski (approved):

frontendResourceBundles/src/main/resources/GitMacheteBundle.properties
```diff
@@ -29,6 +29,7 @@ action.GitMachete.GitCommandUpdatingCurrentBranchBackgroundable.notification.tit
 
 # Quasi-title capitalization intended since we always write "Git Machete" with initial caps.
 action.GitMachete.GitMacheteRepositoryUpdateBackgroundable.task-title=Updating Git Machete status
+action.GitMachete.GitMacheteRepositoryUpdateBackgroundable.notification.title.failed=Cannot display Git Machete branch layout
```
micpiotrowski, line 32: Maye `Cannot create Git Machete branch layout` would be more appropriate?

PawelLipski: Done

PR #682, opened 2021-01-07
[discover] Add discover dialog mnemonics

mkondratek:


frontendResourceBundles/src/main/resources/GitMacheteBundle.properties
```diff
@@ -250,6 +250,8 @@ action.GitMachete.DiscoverAction.text=Discover Branch Layout
 action.GitMachete.DiscoverAction.description=Discover branch layout tree
 action.GitMachete.DiscoverAction.discovered-branch-tree-dialog.empty-table-text=No branch has been discovered
 action.GitMachete.DiscoverAction.discovered-branch-tree-dialog.save-button-text=Save
```
PawelLipski, line 252: Hmm now some mnemonics are here, and some in GraphTableDialog.kt ... maybe this can be unified in favor of property files?

mkondratek: I tried to move them all into the bundle. Some of the texts belong to swing components directly, some to higher-level IntelliJ components. The mnemonic handling works differently in each case :c

PawelLipski (approved):

micpiotrowski (approved):

PR #689, opened 2021-01-26
[discover] Do not queue machete repository update in project resolver…

### LINKED ISSUE #667: Autodiscover happens prematurely
Open a project with no .git/machete file. Notification shows up before Git Machete tab is even opened:

image

This might be confusing for newcomers.

I remember we've already resolved this issue... possibly it only manifests itself on 2020.3? Or a regression happened :c

mkondratek:
… - fixes premature autodiscover (closes #667)

### DIFF
--- a/frontendExternalSystem/src/main/java/com/virtuslab/gitmachete/frontend/externalsystem/project/MacheteProjectResolver.java
+++ b/frontendExternalSystem/src/main/java/com/virtuslab/gitmachete/frontend/externalsystem/project/MacheteProjectResolver.java
@@ -9,15 +9,11 @@ import com.intellij.openapi.externalSystem.model.project.ProjectData;
 import com.intellij.openapi.externalSystem.model.task.ExternalSystemTaskId;
 import com.intellij.openapi.externalSystem.model.task.ExternalSystemTaskNotificationListener;
 import com.intellij.openapi.externalSystem.service.project.ExternalSystemProjectResolver;
-import io.vavr.control.Option;
-import lombok.CustomLog;
 import org.checkerframework.checker.nullness.qual.Nullable;
 
 import com.virtuslab.gitmachete.frontend.externalsystem.MacheteProjectService;
 import com.virtuslab.gitmachete.frontend.externalsystem.settings.MacheteExecutionSettings;
-import com.virtuslab.gitmachete.frontend.ui.providerservice.GraphTableProvider;
 
-@CustomLog
 public class MacheteProjectResolver implements ExternalSystemProjectResolver<MacheteExecutionSettings> {
 
   @Override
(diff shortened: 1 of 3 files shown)


micpiotrowski (approved):

frontendExternalSystem/src/main/java/com/virtuslab/gitmachete/frontend/externalsystem/project/MacheteProjectResolver.java
```diff
@@ -28,16 +24,6 @@
       @Nullable MacheteExecutionSettings settings,
       ExternalSystemTaskNotificationListener listener)
       throws ExternalSystemException, IllegalArgumentException, IllegalStateException {
-
-    var graphTableProvider = Option.of(settings)
```
PawelLipski, line 32: Wow okay... any idea what was the original purpose of this piece? As in, why it used to be needed back then but redundant/erroneous now?...

PawelLipski: The point being, whether by removing this stuff we don't break sth else

PawelLipski:
Hmm I'm getting the following error on a repo where there was NO .git/machete file:

image

But other than that, all happened as expected (autodiscover only when the tab opened, with a notification popping up)

frontendExternalSystem/src/main/java/com/virtuslab/gitmachete/frontend/externalsystem/project/MacheteProjectResolver.java
mkondratek: Now we have got this part of the code

```
addAncestorListener(new AncestorListenerAdapter() {
      @Override
      public void ancestorAdded(AncestorEvent event) {
        var gitRepository = selectedGitRepositoryProvider.getSelectedGitRepository().getOrNull();
        if (gitRepository != null) {
          var macheteFilePath = getMacheteFilePath(gitRepository);
          Runnable queueDiscoverOperation = () -> graphTable.queueDiscover(macheteFilePath, () -> {});
          var rediscoverSuggester = new RediscoverSuggester(gitRepository, queueDiscoverOperation);
          graphTable.queueRepositoryUpdateAndModelRefresh(() -> rediscoverSuggester.perform());
        }
      }
    });
```

https://github.com/VirtusLab/git-machete-intellij-plugin/blob/59f1e2000aa390475f2a50005f2acf064bb0f745/frontendUiRootImpl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/root/GitMachetePanel.java#L60

The reason is that we want to queue repository (first) update at the right moment. Resolver called it at the very beginning of the IDE run. In the meantime came the above part which calls the update later, on Git Machete tab opening.

PawelLipski (approved):

PR #693, opened 2021-01-27
Extend supported version range to 2021.1 + check compatibility with 2021.1-EAP

PawelLipski:
As for now there are suspiciously ZERO additional binary incompatibilities comparing to 2020.3... let's wait till a few further EAP snapshots with actually releasing this version range extension.


.circleci/config.yml
```diff
@@ -86,14 +86,14 @@ jobs:
           command: ./gradlew runPluginVerifier
 
       - run:
-          name: Run UI tests against earlier supported major IntelliJ versions
-          command: ./scripts/run-ui-tests latestMinorsOfOldSupportedMajors
+          name: Run UI tests against an EAP of the latest supported major IntelliJ version (if applicable)
```
PawelLipski, line 89: Reverted the order for the sake of quicker feedback in case anything goes wrong

build.gradle
```diff
@@ -22,13 +22,20 @@ ext {
     // Most recent minor versions of all major releases
     // between earliest supported (incl.) and latest stable (excl.), used for binary compatibility checks
     latestMinorsOfOldSupportedMajors: ['2020.1.4', '2020.2.4'],
-    latestSupportedMajor: '2020.3',
```
PawelLipski, line 25: One less value to bump manually ;p
This value (corresponding to `until-build`) can be derived automatically, see below

version.gradle
```diff
@@ -1,3 +1,3 @@
 ext {
-  PROSPECTIVE_RELEASE_VERSION = '0.8.3'
+  PROSPECTIVE_RELEASE_VERSION = '0.9.0'
```
mkondratek, line 2: :open_mouth:

mkondratek (approved):

version.gradle
PawelLipski: As per convention... minor is bumped when the IDE compatibility range is extended

PR #694, opened 2021-01-27
Do not show double-listed branches in graph table (closes #341)

### LINKED ISSUE #341: Disallow listing the same branch more than once in .git/machete
Unless we find it useful for whatever reason (but I don't think that's the case)

This should be addressed both on editor/language level and the branch layout parser level

mkondratek:
image

---

- add `duplicatedBranchNames` (to `IGitMacheteRepositorySnapshot`)
- add tests for the derivation of skipped and duplicated 
- fix floating "refresh" button

### DIFF
--- a/backendApi/src/main/java/com/virtuslab/gitmachete/backend/api/IGitMacheteRepositorySnapshot.java
+++ b/backendApi/src/main/java/com/virtuslab/gitmachete/backend/api/IGitMacheteRepositorySnapshot.java
@@ -1,6 +1,7 @@
 package com.virtuslab.gitmachete.backend.api;
 
 import io.vavr.collection.List;
+import io.vavr.collection.Set;
 import io.vavr.control.Option;
 
 import com.virtuslab.branchlayout.api.IBranchLayout;
@@ -21,9 +22,9 @@ public interface IGitMacheteRepositorySnapshot {
 
   Option<IManagedBranchSnapshot> getManagedBranchByName(String branchName);
 
-  List<String> getDuplicatedBranchNames();
+  Set<String> getDuplicatedBranchNames();
 
-  List<String> getSkippedBranchNames();
+  Set<String> getSkippedBranchNames();
 
   Option<IExecutionResult> executeMachetePreRebaseHookIfPresent(IGitRebaseParameters gitRebaseParameters)
       throws GitMacheteException;
--- a/backendApi/src/main/java/com/virtuslab/gitmachete/backend/api/IGitMacheteRepositorySnapshot.java
+++ b/backendApi/src/main/java/com/virtuslab/gitmachete/backend/api/IGitMacheteRepositorySnapshot.java
@@ -21,6 +21,8 @@ public interface IGitMacheteRepositorySnapshot {
 
   Option<IManagedBranchSnapshot> getManagedBranchByName(String branchName);
 
+  List<String> getDuplicatedBranchNames();
+
   List<String> getSkippedBranchNames();
 
   Option<IExecutionResult> executeMachetePreRebaseHookIfPresent(IGitRebaseParameters gitRebaseParameters)
--- a/backendImpl/src/main/java/com/virtuslab/gitmachete/backend/impl/GitMacheteRepository.java
+++ b/backendImpl/src/main/java/com/virtuslab/gitmachete/backend/impl/GitMacheteRepository.java
@@ -20,6 +20,7 @@ import io.vavr.collection.Map;
 import io.vavr.collection.Queue;
 import io.vavr.collection.Seq;
 import io.vavr.collection.Set;
+import io.vavr.collection.TreeSet;
 import io.vavr.control.Option;
 import io.vavr.control.Try;
 import lombok.AccessLevel;
@@ -312,9 +313,10 @@ public class GitMacheteRepository implements IGitMacheteRepository {
       var rootBranchTries = branchLayout.getRootEntries().map(entry -> Try.of(() -> createGitMacheteRootBranch(entry)));
       var rootBranchCreationResults = Try.sequence(rootBranchTries).getOrElseThrow(GitMacheteException::getOrWrap).toList();
       var rootBranches = rootBranchCreationResults.flatMap(creationResult -> creationResult.getCreatedBranches());
-      var skippedBranchNames = rootBranchCreationResults.flatMap(creationResult -> creationResult.getSkippedBranchNames());
+      var skippedBranchNames = rootBranchCreationResults.flatMap(creationResult -> creationResult.getSkippedBranchNames())
+          .toSet();
       var duplicatedBranchNames = rootBranchCreationResults
-          .flatMap(creationResult -> creationResult.getDuplicatedBranchNames()).toSet().toList();
+          .flatMap(creationResult -> creationResult.getDuplicatedBranchNames()).toSet();
 
       var managedBranchByName = createManagedBranchByNameMap(rootBranches);
 
@@ -357,7 +359,7 @@ public class GitMacheteRepository implements IGitMacheteRepository {
       return branchByName;
     }
 
-    private CreatedAndSkippedAndDuplicatedBranches<RootManagedBranchSnapshot> createGitMacheteRootBranch(
+    private CreatedAndDuplicatedAndSkippedBranches<RootManagedBranchSnapshot> createGitMacheteRootBranch(
         IBranchLayoutEntry entry) throws GitCoreException {
 
       var branchName = entry.getName();
@@ -367,7 +369,7 @@ public class GitMacheteRepository implements IGitMacheteRepository {
             .map(childEntry -> Try.of(() -> createGitMacheteRootBranch(childEntry)));
         var newRoots = Try.sequence(childBranchTries)
             .getOrElseThrow(GitCoreException::getOrWrap)
-            .fold(CreatedAndSkippedAndDuplicatedBranches.empty(), CreatedAndSkippedAndDuplicatedBranches::merge);
+            .fold(CreatedAndDuplicatedAndSkippedBranches.empty(), CreatedAndDuplicatedAndSkippedBranches::merge);
         return newRoots.withExtraSkippedBranch(branchName);
       }
 
@@ -376,7 +378,7 @@ public class GitMacheteRepository implements IGitMacheteRepository {
             .map(childEntry -> Try.of(() -> createGitMacheteRootBranch(childEntry)));
         var newRoots = Try.sequence(childBranchTries)
             .getOrElseThrow(GitCoreException::getOrWrap)
-            .fold(CreatedAndSkippedAndDuplicatedBranches.empty(), CreatedAndSkippedAndDuplicatedBranches::merge);
+            .fold(CreatedAndDuplicatedAndSkippedBranches.empty(), CreatedAndDuplicatedAndSkippedBranches::merge);
         return newRoots.withExtraDuplicatedBranch(branchName);
       }
 
@@ -394,25 +396,25 @@ public class GitMacheteRepository implements IGitMacheteRepository {
       var createdRootBranch = new RootManagedBranchSnapshot(branchName, branchFullName,
           childBranches.getCreatedBranches(), pointedCommit, remoteTrackingBranch, syncToRemoteStatus, customAnnotation,
           statusHookOutput);
-      return CreatedAndSkippedAndDuplicatedBranches.of(List.of(createdRootBranch),
+      return CreatedAndDuplicatedAndSkippedBranches.of(List.of(createdRootBranch),
           childBranches.getDuplicatedBranchNames(), childBranches.getSkippedBranchNames());
     }
 
-    private CreatedAndSkippedAndDuplicatedBranches<NonRootManagedBranchSnapshot> createGitMacheteNonRootBranch(
+    private CreatedAndDuplicatedAndSkippedBranches<NonRootManagedBranchSnapshot> createGitMacheteNonRootBranch(
         IGitCoreLocalBranchSnapshot parentCoreLocalBranch,
         IBranchLayoutEntry entry) throws GitCoreException {
 
       var branchName = entry.getName();
       IGitCoreLocalBranchSnapshot coreLocalBranch = localBranchByName.get(branchName).getOrNull();
       if (coreLocalBranch == null) {
-        CreatedAndSkippedAndDuplicatedBranches<NonRootManagedBranchSnapshot> childResult = deriveChildBranches(
+        CreatedAndDuplicatedAndSkippedBranches<NonRootManagedBranchSnapshot> childResult = deriveChildBranches(
             parentCoreLocalBranch,
             entry.getChildren());
         return childResult.withExtraSkippedBranch(branchName);
       }
 
       if (!createdBranches.add(branchName)) {
-        CreatedAndSkippedAndDuplicatedBranches<NonRootManagedBranchSnapshot> childResult = deriveChildBranches(
+        CreatedAndDuplicatedAndSkippedBranches<NonRootManagedBranchSnapshot> childResult = deriveChildBranches(
             parentCoreLocalBranch,
             entry.getChildren());
         return childResult.withExtraDuplicatedBranch(branchName);
@@ -449,7 +451,7 @@ public class GitMacheteRepository implements IGitMacheteRepository {
       var result = new NonRootManagedBranchSnapshot(branchName, branchFullName, childBranches.getCreatedBranches(),
           pointedCommit, remoteTrackingBranch, syncToRemoteStatus, customAnnotation, statusHookOutput, forkPoint,
           commits.map(CommitOfManagedBranch::new), syncToParentStatus);
-      return CreatedAndSkippedAndDuplicatedBranches.of(List.of(result),
+      return CreatedAndDuplicatedAndSkippedBranches.of(List.of(result),
           childBranches.getDuplicatedBranchNames(), childBranches.getSkippedBranchNames());
     }
 
@@ -606,7 +608,7 @@ public class GitMacheteRepository implements IGitMacheteRepository {
       }
     }
 
-    private CreatedAndSkippedAndDuplicatedBranches<NonRootManagedBranchSnapshot> deriveChildBranches(
+    private CreatedAndDuplicatedAndSkippedBranches<NonRootManagedBranchSnapshot> deriveChildBranches(
         IGitCoreLocalBranchSnapshot parentCoreLocalBranch,
         List<IBranchLayoutEntry> entries) throws GitCoreException {
 
@@ -614,7 +616,7 @@ public class GitMacheteRepository implements IGitMacheteRepository {
           () -> createGitMacheteNonRootBranch(parentCoreLocalBranch, entry)));
       return Try.sequence(childBranchTries)
           .getOrElseThrow(GitCoreException::getOrWrap)
-          .fold(CreatedAndSkippedAndDuplicatedBranches.empty(), CreatedAndSkippedAndDuplicatedBranches::merge);
+          .fold(CreatedAndDuplicatedAndSkippedBranches.empty(), CreatedAndDuplicatedAndSkippedBranches::merge);
     }
 
     private SyncToRemoteStatus deriveSyncToRemoteStatus(IGitCoreLocalBranchSnapshot coreLocalBranch) throws GitCoreException {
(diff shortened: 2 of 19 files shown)


PawelLipski:
Hmm looking at the screenshot... why is `build-chain` not mentioned in notification or graph table? Is it just un-refreshed yet?

mkondratek:
> Hmm looking at the screenshot... why is `build-chain` not mentioned in notification or graph table? Is it just un-refreshed yet?

wooops, that actually is a really interesting case 

The second occurrence of `call-ws` has been ignored **with all of its children**. 
I think that a more accurate warning message is neccecary.
Merging children from reoccurring branches would be tough since our structure here is immutable. :C

PawelLipski:
> The second occurrence of call-ws has been ignored with all of its children.
Hmm I think if this second occurrence of `call-ws` was a **non-existent** branch instead, children would be just re-attached under `call-ws`'s parent... is the logic supposed to be different for duplicates? 🤔

mkondratek:
image

mkondratek:
Additionally... I lied here https://github.com/VirtusLab/git-machete-intellij-plugin/pull/689#discussion_r564807372

The project resolver must perform the update and refresh because this code is called after clicking the floating button from the machete file. On the other hand, it was responsible for the premature discover. Hence, let's update and refresh from it only if the machete tab is opened.

backendImpl/src/main/java/com/virtuslab/gitmachete/backend/impl/GitMacheteRepository.java
```diff
@@ -929,27 +951,36 @@ IGitMacheteRepositorySnapshot discoverLayoutAndCreateSnapshot(int mostRecentlyCh
 
   @Getter
   @RequiredArgsConstructor(access = AccessLevel.PRIVATE)
-  private static final class CreatedAndSkippedBranches<T extends BaseManagedBranchSnapshot> {
+  private static final class CreatedAndSkippedAndDuplicatedBranches<T extends BaseManagedBranchSnapshot> {
```
PawelLipski, line 954: Nit/paranoia: since both duplicated and skipped branches have the same type, they can be confused in contexts where both are expected (like `GitMacheteRepositorySnapshot`'s c'tor). I'd suggest keeping one fixed order (alphabetical FTW) everywhere in names/fields/methods to reduce the risk of such mistake... so this would be `CreatedAndDuplicatedAndSkippedBranches`

frontendExternalSystem/src/main/java/com/virtuslab/gitmachete/frontend/externalsystem/project/MacheteProjectResolver.java
```diff
@@ -33,4 +54,16 @@
   public boolean cancelTask(ExternalSystemTaskId taskId, ExternalSystemTaskNotificationListener listener) {
     return false;
   }
+
+  @SuppressWarnings({"guieffect:call.invalid.ui", "interning:not.interned"})
```
PawelLipski, line 58: `not.intern`... just like you ‼️

frontendResourceBundles/src/main/resources/GitMacheteBundle.properties
```diff
@@ -250,6 +250,7 @@ string.GitMachete.EnhancedGraphTable.automatic-discover.success-message=Branch l
 string.GitMachete.EnhancedGraphTable.empty-table-text.cannot-discover-layout=Provided machete file ({0}) is empty and branch layout can''t be automatically detected
 string.GitMachete.EnhancedGraphTable.empty-table-text.loading=Loading...
 string.GitMachete.EnhancedGraphTable.empty-table-text.only-skipped-in-machete-file=None of the branches provided by machete file ({0}) exists within the local repository
+string.GitMachete.EnhancedGraphTable.duplicated-branches-text=The following branches defined by machete file appear more than once:
```
PawelLipski, line 253: Or just
```suggestion
string.GitMachete.EnhancedGraphTable.duplicated-branches-text=The following branches appear more than once in machete file:
```

backendImpl/src/main/java/com/virtuslab/gitmachete/backend/impl/GitMacheteRepository.java
mkondratek: Whooops, good point. Firstly I added the duplicated field as the last one, then I moved it b4 skipped to keep the order but I forgot about the class name Xd

frontendExternalSystem/src/main/java/com/virtuslab/gitmachete/frontend/externalsystem/project/MacheteProjectResolver.java
```diff
@@ -33,4 +54,16 @@
...
+  private Boolean hasSelectedGitMacheteTab(Project project) {
+    var toolWindowManager = ToolWindowManager.getInstance(project);
+    var toolWindow = toolWindowManager.getToolWindow(ToolWindowId.VCS);
+    if (toolWindow != null) {
+      var contentManager = toolWindow.getContentManager();
+      var gitMacheteContent = contentManager.findContent("Git Machete");
```
PawelLipski, line 64: Nit: do we keep tab name somewhere in properties?

frontendExternalSystem/src/main/java/com/virtuslab/gitmachete/frontend/externalsystem/project/MacheteProjectResolver.java
```diff
@@ -24,6 +31,20 @@
       @Nullable MacheteExecutionSettings settings,
       ExternalSystemTaskNotificationListener listener)
       throws ExternalSystemException, IllegalArgumentException, IllegalStateException {
+
+    var graphTableProvider = Option.of(settings)
+        .map(s -> s.getProject().getService(GraphTableProvider.class));
+    var hasSelectedGitMacheteTab = Option.of(settings)
+        .map(s -> s.getProject())
+        .map(this::hasSelectedGitMacheteTab)
+        .getOrElse(Boolean.FALSE);
```
PawelLipski, line 40: Wow how come I've never seen that 😯

PawelLipski: (Most likely I was just always relying on autoboxing for booleans... also `Boolean` is relatively rarely a type param in any generic 🤔 )

frontendResourceBundles/src/main/resources/GitMacheteBundle.properties
```diff
@@ -250,6 +250,7 @@ string.GitMachete.EnhancedGraphTable.automatic-discover.success-message=Branch l
...
 string.GitMachete.EnhancedGraphTable.skipped-branches-text=The following branches defined by machete file do not belong to the local repository:
```
PawelLipski, line 254: Btw, sounds more natural IMO:
```suggestion
string.GitMachete.EnhancedGraphTable.skipped-branches-text=The following branches listed in machete file do not belong to the local repository:
```

PawelLipski: As in, technically it's git that _defines_ the branches, not us ;p we just reference them

frontendExternalSystem/src/main/java/com/virtuslab/gitmachete/frontend/externalsystem/project/MacheteProjectResolver.java
PawelLipski, line 40: Or just:
```suggestion
    var hasSelectedGitMacheteTab = settings != null && hasSelectedGitMacheteTab(settings.getProject());
```

frontendExternalSystem/src/main/java/com/virtuslab/gitmachete/frontend/externalsystem/project/MacheteProjectResolver.java
mkondratek: I can bundle it but note that the bundle cannot be used by plugin.xml
```
<extensions defaultExtensionNs="com.intellij">
        <changesViewContent tabName="Git Machete"
                            className="com.virtuslab.gitmachete.frontend.ui.impl.root.GitMacheteContentProvider"
                            predicateClassName="com.virtuslab.gitmachete.frontend.ui.impl.root.GitMacheteVisibilityPredicate"/>
...
```

PawelLipski: Aaa ok, leave as is then

backendImpl/src/main/java/com/virtuslab/gitmachete/backend/impl/GitMacheteRepository.java
```diff
@@ -312,6 +313,8 @@ IGitMacheteRepositorySnapshot createSnapshot(IBranchLayout branchLayout) throws
       var rootBranchCreationResults = Try.sequence(rootBranchTries).getOrElseThrow(GitMacheteException::getOrWrap).toList();
       var rootBranches = rootBranchCreationResults.flatMap(creationResult -> creationResult.getCreatedBranches());
       var skippedBranchNames = rootBranchCreationResults.flatMap(creationResult -> creationResult.getSkippedBranchNames());
+      var duplicatedBranchNames = rootBranchCreationResults
+          .flatMap(creationResult -> creationResult.getDuplicatedBranchNames()).toSet().toList();
```
PawelLipski, line 317: Hmm why does have it be both `toSet` and `toList`?  `toSet` for removing duplicates from the list of duplicates itself? ;)

frontendExternalSystem/src/main/java/com/virtuslab/gitmachete/frontend/externalsystem/project/MacheteProjectResolver.java
mkondratek: Yeah, I would rather leave it (we use `"Git Machete"` in two places only)

backendImpl/src/main/java/com/virtuslab/gitmachete/backend/impl/GitMacheteRepository.java
mkondratek: :yes:

PawelLipski: This actually suggests that `duplicatedBranchNames` (and possibly `skippedBranchNames`) could instead be represented as `Set` everywhere (since we DO care about things being unique in these collections, but DON'T care about ordering &mdash; at least not until we show them to the user, when we might want to sort them first)

backendApi/src/main/java/com/virtuslab/gitmachete/backend/api/IGitMacheteRepositorySnapshot.java
```diff
@@ -21,6 +21,8 @@
 
   Option<IManagedBranchSnapshot> getManagedBranchByName(String branchName);
 
+  List<String> getDuplicatedBranchNames();
```
PawelLipski, line 24: Btw: crazy/unlikely corner case, what if a branch is both duplicated AND skipped? ;) that's perfectly possible ofc

PawelLipski: Also, pls add some 3-ish test cases to `backendImpl/src/tests`

backendImpl/src/main/java/com/virtuslab/gitmachete/backend/impl/GitMacheteRepository.java
```diff
@@ -929,27 +951,36 @@ IGitMacheteRepositorySnapshot discoverLayoutAndCreateSnapshot(int mostRecentlyCh
...
+    CreatedAndSkippedAndDuplicatedBranches<T> withExtraSkippedBranch(String skippedBranchName) {
+      return new CreatedAndSkippedAndDuplicatedBranches<T>(getCreatedBranches(), getDuplicatedBranchNames(),
+          getSkippedBranchNames().append(skippedBranchName));
     }
 
-    static <T extends BaseManagedBranchSnapshot> CreatedAndSkippedBranches<T> empty() {
-      return new CreatedAndSkippedBranches<T>(List.empty(), List.empty());
+    static <T extends BaseManagedBranchSnapshot> CreatedAndSkippedAndDuplicatedBranches<T> of(List<T> createdBranches,
+        List<String> duplicatedBranchName, List<String> skippedBranchNames) {
+      return new CreatedAndSkippedAndDuplicatedBranches<T>(createdBranches, duplicatedBranchName, skippedBranchNames);
```
PawelLipski, line 972: Whoops okay here args are swapped 😅

PawelLipski: Oh ok they aren't coz fields are in abc order... nvm

frontendExternalSystem/src/main/java/com/virtuslab/gitmachete/frontend/externalsystem/project/MacheteProjectResolver.java
```diff
@@ -33,4 +54,16 @@
   public boolean cancelTask(ExternalSystemTaskId taskId, ExternalSystemTaskNotificationListener listener) {
     return false;
   }
+
+  @SuppressWarnings({"guieffect:call.invalid.ui", "interning:not.interned"})
+  private Boolean hasSelectedGitMacheteTab(Project project) {
+    var toolWindowManager = ToolWindowManager.getInstance(project);
+    var toolWindow = toolWindowManager.getToolWindow(ToolWindowId.VCS);
+    if (toolWindow != null) {
+      var contentManager = toolWindow.getContentManager();
```
PawelLipski, line 63: Actually... why suppressing `guieffect:call.invalid.ui` was needed here? Possibly the offending method should have the Checker signature changed in stub files instead... since we likely **treat it as UI-thread-only** while in fact **it's any-thread-safe** (?)

frontendExternalSystem/src/main/java/com/virtuslab/gitmachete/frontend/externalsystem/project/MacheteProjectResolver.java
```diff
@@ -33,4 +54,16 @@
...
+      var gitMacheteContent = contentManager.findContent("Git Machete");
+      return contentManager.getSelectedContent() == gitMacheteContent;
+    }
+    return Boolean.FALSE;
```
PawelLipski, line 67: Hmm just `boolean`/`false` ;)

frontendExternalSystem/src/main/java/com/virtuslab/gitmachete/frontend/externalsystem/project/MacheteProjectResolver.java
PawelLipski: Rule of the thumb: `.getOrElse(false/true)` can almost always be rewritten to sth more readable/less "wrapped"

backendApi/src/main/java/com/virtuslab/gitmachete/backend/api/IGitMacheteRepositorySnapshot.java
mkondratek: Presently it will be considered as a skipped one only. does it make sense?

PawelLipski: Okay, makes sense, actually maybe even more than the inverse (i.e. considered duplicate only)... but still, that's more of a puzzle than actual problem ;)

mkondratek:
btw opened #695; a temporary solution implemented

frontendExternalSystem/src/main/java/com/virtuslab/gitmachete/frontend/externalsystem/project/MacheteProjectResolver.java
```diff
@@ -55,15 +52,17 @@ public boolean cancelTask(ExternalSystemTaskId taskId, ExternalSystemTaskNotific
     return false;
   }
 
-  @SuppressWarnings({"guieffect:call.invalid.ui", "interning:not.interned"})
+  @SuppressWarnings("interning:not.interned")
   private Boolean hasSelectedGitMacheteTab(Project project) {
     var toolWindowManager = ToolWindowManager.getInstance(project);
     var toolWindow = toolWindowManager.getToolWindow(ToolWindowId.VCS);
     if (toolWindow != null) {
-      var contentManager = toolWindow.getContentManager();
```
PawelLipski: Whoops okay so this was potentially unsafe?

frontendExternalSystem/src/main/java/com/virtuslab/gitmachete/frontend/externalsystem/project/MacheteProjectResolver.java
```diff
@@ -55,15 +52,17 @@ public boolean cancelTask(ExternalSystemTaskId taskId, ExternalSystemTaskNotific
...
-      var gitMacheteContent = contentManager.findContent("Git Machete");
-      return contentManager.getSelectedContent() == gitMacheteContent;
+      var contentManager = toolWindow.getContentManagerIfCreated();
```
PawelLipski, line 60: Btw you can mark these methods as `@Nullable` in stubs for safety in future (Checker isn't very sound if it comes to assumptions about methods coming from compiler jars, see https://github.com/typetools/checker-framework/issues/4079)

branchLayoutApi/src/main/java/com/virtuslab/branchlayout/api/BranchLayout.java
```diff
@@ -37,7 +37,14 @@ public IBranchLayout slideOut(String branchName) throws EntryDoesNotExistExcepti
     if (entryToSlideOut == null) {
       throw new EntryDoesNotExistException("Branch entry '${branchName}' does not exist");
     }
-    return new BranchLayout(rootEntries.flatMap(rootEntry -> slideOut(rootEntry, entryToSlideOut)));
+
+    var newBranchLayout = new BranchLayout(rootEntries.flatMap(rootEntry -> slideOut(rootEntry, entryToSlideOut)));
```
PawelLipski, line 41: Hmm I don't like this double slide-out here.... how about just sth like:

`return new BranchLayout(rootEntries.flatMap(rootEntry -> slideOut(rootEntry, branchName)));`

as in, branches just slid out based on their matching name, with no need to perform `findEntryByName` in the first place?

Also, if you could a test that would've captured the bug in the previous implementation (i.e. the one where only one of duplicated entries is being slid out)

frontendExternalSystem/src/main/java/com/virtuslab/gitmachete/frontend/externalsystem/project/MacheteProjectResolver.java
mkondratek: :yes:

branchLayoutApi/src/main/java/com/virtuslab/branchlayout/api/BranchLayout.java
```diff
@@ -32,29 +32,19 @@ public BranchLayout(List<IBranchLayoutEntry> rootEntries) {
...
-    }
+  public IBranchLayout slideOut(String branchName) {
+    // TODO (#695): disable slide out for duplicated branches
+    return new BranchLayout(rootEntries.flatMap(rootEntry -> slideOut(rootEntry, branchName)));
   }
 
   private List<IBranchLayoutEntry> slideOut(
       @FindDistinct IBranchLayoutEntry entry,
-      @FindDistinct IBranchLayoutEntry entryToSlideOut) {
+      @FindDistinct String entryNameToSlideOut) {
```
PawelLipski, line 42: `@FindDistinct` likely no longer needed on either param (https://checkerframework.org/api/org/checkerframework/checker/interning/qual/FindDistinct.html)

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/backgroundables/SlideInBackgroundable.java
```diff
@@ -76,15 +76,7 @@ public void run(ProgressIndicator indicator) {
         targetBranchLayout = branchLayout;
       } else {
         entryToSlideIn = entryByName.withChildren(List.empty());
-        try {
-          targetBranchLayout = branchLayout.slideOut(slideInOptions.getName());
-        } catch (EntryDoesNotExistException e) {
```
PawelLipski, line 81: Thing to make sure now: whether sliding in a managed branch under a new parent still works :goncern:

mkondratek: > whether it fails when the provided branch name is invalid
what do you mean?

PawelLipski: Ah ok, nvm 🤦🏻

PawelLipski (approved):

PR #699, opened 2021-01-29
Prevent calling UI-thread-unsafe methods on UI thread (complement of GuiEffectChecker)

PawelLipski:
GuiEffectChecker ensures that:
1. UI-touching (aka `@UIEffect`) methods are **only ever** called on UI thread
2. UI-touching methods are never called from non-UI-touching (aka `@SafeEffect`) methods

This PR kinda-ensures that:
1. UI-unsafe methods (mostly the heavyweight ones, due to disk/network access) are **never** called on UI thread
2. UI-unsafe methods are never called from non-UI-unsafe methods

Note that non-UI-unsafe does NOT imply UI-touching,
and non-UI-touching does NOT imply UI-unsafe.
In fact, most methods are both non-UI-touching and non-UI-unsafe.

There are, however, obviously no methods that would be both UI-touching AND UI-unsafe.

### DIFF
--- a/frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseSlideOutBranchAction.java
+++ b/frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseSlideOutBranchAction.java
@@ -13,7 +13,6 @@ import com.intellij.openapi.application.ModalityState;
 import com.intellij.openapi.progress.ProgressIndicator;
 import com.intellij.openapi.progress.Task;
 import com.intellij.openapi.project.Project;
-import com.intellij.openapi.util.ThrowableComputable;
 import com.intellij.openapi.vcs.VcsException;
 import com.intellij.openapi.vcs.VcsNotifier;
 import com.intellij.openapi.vfs.VirtualFile;
@@ -24,14 +23,13 @@ import git4idea.repo.GitRepository;
 import io.vavr.control.Option;
 import lombok.CustomLog;
 import org.checkerframework.checker.guieffect.qual.UIEffect;
-import org.checkerframework.checker.nullness.qual.Nullable;
 
 import com.virtuslab.branchlayout.api.BranchLayoutException;
 import com.virtuslab.gitmachete.backend.api.IManagedBranchSnapshot;
 import com.virtuslab.gitmachete.frontend.actions.dialogs.DeleteBranchOnSlideOutSuggestionDialog;
 import com.virtuslab.gitmachete.frontend.actions.expectedkeys.IExpectsKeyGitMacheteRepository;
 import com.virtuslab.logger.IEnhancedLambdaLogger;
-import com.virtuslab.qual.guieffect.NotUIThreadSafe;
+import com.virtuslab.qual.guieffect.UIThreadUnsafe;
 
 @CustomLog
 public abstract class BaseSlideOutBranchAction extends BaseGitMacheteRepositoryReadyAction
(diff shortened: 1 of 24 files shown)


frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseSlideOutBranchAction.java
```diff
@@ -140,9 +139,15 @@ private void suggestBranchDeletion(AnActionEvent anActionEvent, String branchNam
 
     new Task.Backgroundable(project, "Deleting branch if required...") {
       @Override
+      @UIThreadUnsafe
       public void run(ProgressIndicator indicator) {
         if (slideOutOptions != null) {
           handleBranchDeletionDecision(project, branchName, gitRepository, anActionEvent, slideOutOptions.shouldDelete());
+
+          if (slideOutOptions.shouldRemember()) {
```
PawelLipski, line 147: @mkondratek since you're the original author... if you could eyeball this area if there won't be any unintended consequences of this change (race conditions?)

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseSlideOutBranchAction.java
```diff
@@ -210,12 +211,10 @@ private void handleBranchDeletionDecision(Project project, String branchName, Gi
     getGraphTable(anActionEvent).queueRepositoryUpdateAndModelRefresh();
   }
 
-  @NotUIThreadSafe
+  @UIThreadUnsafe
   private Option<Boolean> getDeleteLocalBranchOnSlideOutGitConfigValue(Project project, VirtualFile root) {
     try {
-      ThrowableComputable<@Nullable String, VcsException> computable = () -> GitConfigUtil.getValue(project, root,
```
PawelLipski, line 216: This was redundant, looked like some leftover from past implementation (?)

qual/src/main/java/com/virtuslab/qual/guieffect/UIThreadUnsafe.java
```diff
@@ -6,6 +6,8 @@
...
 
 // TODO (typetools/checker-framework#3252): replace with proper @Heavyweight
-@Retention(RetentionPolicy.CLASS)
-@Target({ElementType.METHOD})
-public @interface NotUIThreadSafe {}
+
+// Must have runtime retention to be visible to ArchUnit tests.
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.METHOD, ElementType.CONSTRUCTOR})
+public @interface UIThreadUnsafe {}
```
PawelLipski, line 13: I don't like this name TBH, mostly due to the negation in _unsafe_ which makes things confusing... any ideas how to rename? 🤔

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseSlideOutBranchAction.java
mkondratek: lgtm

mkondratek: Keeping this piece of code where it was would cause a test failure, right?

mkondratek (approved):
fine

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseSlideOutBranchAction.java
PawelLipski: Yes, exactly: ArchUnit would complain about calling `@UIThreadUnsafe` method from `@UIEffect` (or actually, just from non-`@UIThreadUnsafe`) method

PR #701, opened 2021-02-15
Resolves #289 - Added support for git config core.hooksPath property

### LINKED ISSUE #289: Derive git hooks directory location according to the spec
`git config core.hooksPath` also needs to be taken into account, see `git help hooks`. Currently we're just always looking into `$GITDIR/hooks` for `machete-status-branch` and `machete-pre-rebase`.

DcortezMeleth:


backendImpl/src/main/java/com/virtuslab/gitmachete/backend/impl/hooks/PreRebaseHookExecutor.java
```diff
@@ -20,10 +23,12 @@
   private final File mainDirectory;
   private final File hookFile;
 
-  public PreRebaseHookExecutor(Path mainDirectoryPath, Path gitDirectoryPath) {
+  public PreRebaseHookExecutor(IGitCoreRepository gitCoreRepository, Path mainDirectoryPath, Path gitDirectoryPath) {
+    val hooksDir = gitCoreRepository.deriveConfigValue("core", "hooksPath");
+    val hooksDirPath = hooksDir.map(Paths::get).getOrElse(gitDirectoryPath.resolve("hooks"));
```
PawelLipski, line 28: One thing to consider &mdash; in case of built-in git hooks (like `pre-commit`), what happens if `core.hooksPath` is set, but:
1. given `hooksPath` doesn't exist/is not a directory
2. given `hooksPath` exists but the given hook doesn't exist/isn't executable?

Is the git's behavior to just ignore the hook's execution and NOT do any fallback to `.git/hooks`?

PawelLipski: (which btw reminded me of one thing... https://github.com/VirtusLab/git-machete-intellij-plugin/pull/702)

DcortezMeleth: I can check how Intellij handles those cases. I assume we would like to take the same approach?

PawelLipski: Actually, IntelliJ calls command-line `git` underneath via git4idea, so you can as well check `git` CLI. Yeah, let's be consistent &mdash; I guess the answer is **no fallback** but let's verify

DcortezMeleth: Imho quite interesting how Intellij handle those. In case on missing folder you won't even know that such issue exist. Intellij will work just fine and just skip hook call. No warning by notification, in Event Log and in Git console are present.
In case of hook which is not executable it somewhat better but still you need to search for warning as it's only displayed in Git console. Below there is an example of how Git console looks like after both tests were performed.
image

DcortezMeleth: Now we should ask ourself if we want to be consistent with Intellij or verbose about encountered errors?
Do you have an idea how git cli handle those cases or should I test them too?

PawelLipski: I'd say the most important part was to make sure that the fallback to .git/hooks doesn't occur, error messages (or lack thereof) are secondary

PawelLipski: Let's then stick to the current impl... no further action required 👍🏻

PawelLipski (approved):

PR #708, opened 2021-02-21
Resolves #651 - Added tooltip for root branch as well as non-root bra…

DcortezMeleth:
…nches without action suggested

PR #713, opened 2021-02-23
Ignore spurious, irrelevant errors happening in UI tests

PawelLipski:
Ok @DcortezMeleth, this should fix the errors in the build of #708 until https://github.com/VirtusLab/ide-probe/issues/95 is fixed 🔧 

You can fast-forward your branch to include this change, and close this PR.

### DIFF
--- a/uiTests/src/test/scala/com/virtuslab/gitmachete/uitest/UITestSuite.scala
+++ b/uiTests/src/test/scala/com/virtuslab/gitmachete/uitest/UITestSuite.scala
@@ -2,10 +2,9 @@ package com.virtuslab.gitmachete.uitest
 
 import java.nio.file.Files
 import java.nio.file.attribute.FileTime
-
 import com.virtuslab.gitmachete.testcommon.BaseGitRepositoryBackedIntegrationTestSuite
 import com.virtuslab.gitmachete.testcommon.BaseGitRepositoryBackedIntegrationTestSuite.SETUP_WITH_SINGLE_REMOTE
-import org.junit.{After, Assert, Before, Test}
+import org.junit.{ After, AfterClass, Assert, Before, BeforeClass, Test }
 import org.junit.runner.RunWith
 import org.junit.runners.JUnit4
 import org.virtuslab.ideprobe.Extensions._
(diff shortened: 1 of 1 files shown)


uiTests/src/test/scala/com/virtuslab/gitmachete/uitest/UITestSuite.scala
```diff
@@ -2,18 +2,40 @@ package com.virtuslab.gitmachete.uitest
...
-import org.virtuslab.ideprobe.junit4.RunningIntelliJPerSuite
+
+trait RunningIntelliJPerSuite extends RunningIntelliJPerSuiteBase {
+  @BeforeClass override final def setup(): Unit = super.setup()
+
+  @AfterClass override final def teardown(): Unit = {
+    try {
+      super.teardown()
+    } catch {
+      case e: Exception =>
```
DcortezMeleth, line 23: As I'm not familiar with Scala I will ask why do we have to write singe case here? 
Does it allow us for some type interference or is it not to catch Throwable?
Is it some kind of Scala convention to prefer single case over if?

uiTests/src/test/scala/com/virtuslab/gitmachete/uitest/UITestSuite.scala
```diff
@@ -2,18 +2,40 @@ package com.virtuslab.gitmachete.uitest
...
+        val filteredSuppressed = e.getSuppressed.filterNot { s =>
+          // A duct tape to ignore https://github.com/VirtusLab/ide-probe/issues/95
+          s.getMessage.contains("scala.MatchError: IDE_UPDATE (of class com.intellij.notification.NotificationType)") ||
```
DcortezMeleth, line 26: It doesn't look great but just as you pointed out, this is duct tape we are planning to remove when appropriate issue is closed?

uiTests/src/test/scala/com/virtuslab/gitmachete/uitest/UITestSuite.scala
PawelLipski: It's just that I checked the impl of `super.teardown`, and they simply always throw a `new Exception` in case of any IDE errors (yeah lol xD)

PawelLipski: > Is it some kind of Scala convention to prefer single case over if?

I'd say that yes, Scala likes pattern matching in general (even in cases where `if-else` could be equivalently used), and even requires pattern matching in certain cases &mdash; most notably for `catch` clause, even when just catching a single exception type

PawelLipski: That's just the equivalent of Java's:
```java
try {
...
} catch (Exception e) {
...
}
```

uiTests/src/test/scala/com/virtuslab/gitmachete/uitest/UITestSuite.scala
PawelLipski: Yes, I hope so ;)

DcortezMeleth (approved):

PR #731, opened 2021-04-01
Introduce fetch all timeout (closes #721)

### LINKED ISSUE #721: Avoid doing `git pull` for some time after a recent `git fetch`
Even after running _Fetch all remotes_, when the user runs _Pull branch fast-forward_, there's still a regular `git pull` happening.
I sometimes catch myself doing _Reset to remote_ rather then _Pull_ on the branches that are behind remote to avoid this `git pull`.

In CLI, `git machete traverse --fetch` does a fetch at the very beginning, and later only `git merge --ff-only` when a visited branch is behind remote:

image

In the above listing, `Updating e797251a..f5988961` was a `git merge --ff-only`, not a `git pull`.

The same behavior could be reflected in the plugin. There are no "sessions" similar to `git machete traverse`, but we could implement some timeout (1 minute?).

mkondratek:


frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/toolbar/FetchAllRemotesAction.java
```diff
@@ -19,6 +19,14 @@
 @CustomLog
 public class FetchAllRemotesAction extends BaseProjectDependentAction {
 
+  private static final long FETCH_ALL_UP_TO_DATE_TIMEOUT = 60 * 1000;
+
+  private static long lastFetchTimeMillis = 0;
+
+  public static boolean isUpToDate() {
+    return System.currentTimeMillis() > lastFetchTimeMillis + FETCH_ALL_UP_TO_DATE_TIMEOUT;
```
PawelLipski, line 27: Hmmm isn't the method name INcompatible with the condition? Since if `lastFetchTimeMillis` is, say, `0`, than the condition is satisfied but well, the action is NOT up to date :/

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseFastForwardMergeBranchToParentAction.java
```diff
@@ -84,9 +84,7 @@ public void actionPerformed(AnActionEvent anActionEvent) {
   private void doFastForwardCurrentBranch(Project project,
       GitRepository gitRepository,
       INonRootManagedBranchSnapshot targetBranch) {
-
```
PawelLipski, line 87: Minor typos I've spotted, while you're here:
`build.gradle:14: s/amd/and`
`.circleci/config.yml:21: s/got/go`

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BasePullBranchFastForwardOnlyAction.java
```diff
@@ -86,7 +88,11 @@ public void actionPerformed(AnActionEvent anActionEvent) {
 
       val currentBranchName = Option.of(gitRepository.getCurrentBranch()).map(b -> b.getName()).getOrNull();
       if (localBranchName.equals(currentBranchName)) {
-        doPullCurrentBranchFastForwardOnly(project, gitRepository, remoteBranch);
+        if (FetchAllRemotesAction.isUpToDate()) {
+          doPullCurrentBranchFastForwardOnly(project, gitRepository, remoteBranch);
```
PawelLipski, line 92: Hmm the factoring here is a bit confusing... since we're checking `isUpToDate` here on a different level in this case than in the other case 🤔

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BasePullBranchFastForwardOnlyAction.java
```diff
@@ -137,18 +159,15 @@ private void doPullNonCurrentBranchFastForwardOnly(Project project,
...
-            format(getString("action.GitMachete.BasePullBranchFastForwardOnlyAction.notification.title.pull-success"),
-                localBranch.getName()))
-                    .queue();
+        ourRemoteToOurLocalBranchFetchBackgroundable.queue();
       }
-    }.queue();
+    };
+
+    if (FetchAllRemotesAction.isUpToDate()) {
+      remoteRepoToOurRemoteBranchFetchBackgroundable.queue();
```
PawelLipski, line 167: Related: no need to create both bgables since only one is ever executed

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BasePullBranchFastForwardOnlyAction.java
```diff
@@ -18,8 +18,10 @@
 import com.virtuslab.gitmachete.backend.api.IRemoteTrackingBranchReference;
```
PawelLipski, line 18: Also, pls bump version to 0.9.2

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BasePullBranchFastForwardOnlyAction.java
mkondratek: that ain't true. note that one of the backgroundables executes the other one

PawelLipski: Whoops right, leave as is then

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/common/FastForwardMergeProps.kt
```diff
@@ -0,0 +1,8 @@
+package com.virtuslab.gitmachete.frontend.actions.common
+
+data class FastForwardMergeProps(
```
PawelLipski, line 3: Hmm
1. any specific reason for that one being in kotlin?
2. in general, should Kotlin source live under `src/main/java/` or rather `src/main/kotlin/` (does it even make a difference?)

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/toolbar/FetchAllRemotesAction.java
```diff
@@ -19,6 +20,16 @@
 @CustomLog
 public class FetchAllRemotesAction extends BaseProjectDependentAction {
 
+  private static final long FETCH_ALL_UP_TO_DATE_TIMEOUT = 60 * 1000;
+
+  private static final java.util.Map<String, Long> LAST_FETCH_TIME_MILLIS_BY_REPOSITORY_NAME = new java.util.TreeMap<>();
```
PawelLipski, line 25: Hmm I'd say it should be `lowerCamelCase` as it's not a "constant" really

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/toolbar/FetchAllRemotesAction.java
```diff
@@ -19,6 +20,16 @@
 @CustomLog
 public class FetchAllRemotesAction extends BaseProjectDependentAction {
 
+  private static final long FETCH_ALL_UP_TO_DATE_TIMEOUT = 60 * 1000;
```
PawelLipski, line 23: Pls also include `_MILLIS` for consistency/clarity (there's https://checkerframework.org/manual/#units-checker but that's maybe an overkill for the time being 😅)

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/toolbar/FetchAllRemotesAction.java
PawelLipski: Also, I'd use som concurrent impl since it's accessed not only from UI thread (see `run` below, multiple non-UI threads can execute that method concurrently) :/

frontendActions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BasePullBranchFastForwardOnlyAction.java
```diff
@@ -83,22 +85,35 @@ public void actionPerformed(AnActionEvent anActionEvent) {
         log().warn("Branch '${localBranchName}' does not have a remote tracking branch");
         return;
       }
-
       val currentBranchName = Option.of(gitRepository.getCurrentBranch()).map(b -> b.getName()).getOrNull();
-      if (localBranchName.equals(currentBranchName)) {
-        doPullCurrentBranchFastForwardOnly(project, gitRepository, remoteBranch);
+
+      if (FetchAllRemotesAction.isUpToDate(gitRepository)) {
```
PawelLipski, line 90: Nice, much easier to follow now

PawelLipski:
Hmmm I can't do a fast-forward pull 🤔 

image

image

Both git machete and `git status` tell that local branch is behind remote... yet an attempt to do a `Pull current branch` right after doing a `Fetch` tells me that the merge would be non-fast-forward 😯 

Can you see any obvious reason for that? I can provide more details on the repo state if needed

.circleci/config.yml
```diff
@@ -18,7 +18,7 @@ jobs:
             # If the above key is not found, try finding a cache entry with the same set of libraries
```
PawelLipski, line 18: Another unrel, pls bump image version to 6.0.0

PR #803, opened 2022-04-04
Implement squash-merge detection as in the CLI

PawelLipski:


gitCoreApi/src/main/java/com/virtuslab/gitcore/api/IGitCoreCommit.java
```diff
@@ -12,14 +12,12 @@
 public interface IGitCoreCommit {
   String getShortMessage();
 
-  IGitCorePersonIdentity getAuthor();
```
PawelLipski, line 15: Unused

PR #827, opened 2022-05-05
Fix shortcut and docs for the git machete tab open on mac

mkondratek:
Open Git Machete Tab:
Linux/Windows - `Ctrl + Alt + Shift + M`
Mac OS - `Ctrl + Option + Shift + M`

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -55,8 +55,16 @@ ln -s ../../scripts/git-hooks/post-commit .git/hooks/post-commit
 ln -s ../../scripts/run-pre-build-checks .git/hooks/pre-commit
 ```
 
+#### Windows
 **The hooks do not work on Windows** (however their execution seems to be possible theoretically).
 This is because one may not be emulating bash environment in any way or doing it in some specific way.
+
+#### Mac OS
+Some hooks use `grep`. The Mac OS version of `grep` (FreeBSD) differs from GNU `grep`.
+In order to make `grep` and eventually the hooks working one must:
+1. Install `grep` via `brew` (it will not override system's `grep` - it can be executed as `ggrep`)
+2. Run `brew ls -v grep`; among the other a path like should be found `/opt/homebrew/Cellar/grep/3.7/libexec/gnubin/grep`
+3. Add the found path without `/grep` suffix to `PATH` (`/opt/homebrew/Cellar/grep/3.7/libexec/gnubin` in that case)
 
 ### (optional) Windows
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -64,7 +64,7 @@ Some hooks use `grep`. The Mac OS version of `grep` (FreeBSD) differs from GNU `
 In order to make `grep` and eventually the hooks working one must:
 1. Install `grep` via `brew` (it will not override system's `grep` - it can be executed as `ggrep`)
 2. Run `brew ls -v grep`; among the other a path like should be found `/opt/homebrew/Cellar/grep/3.7/libexec/gnubin/grep`
-3. Add the found path without `/grep` suffix to `PATH` (`/opt/homebrew/Cellar/grep/3.7/libexec/gnubin` in that case)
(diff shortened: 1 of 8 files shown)


azdrojowa123 (approved):

src/main/resources/META-INF/plugin.xml
```diff
@@ -192,6 +193,8 @@
                  We can consider `VcsGroups` that seems to be more stable and place this action in VCS menu -->
             <add-to-group group-id="GitRepositoryActions" anchor="last"/>
             <keyboard-shortcut keymap="$default" first-keystroke="control alt shift M" />
+            <keyboard-shortcut first-keystroke="control alt shift M" keymap="Mac OS X" replace-all="true"/>
```
PawelLipski, line 196: Huh so the previous setting had no effect for Mac at all? 🤔 or was `control` interpreted as `cmd`?

docs/plugins.jetbrains.com/general_usage_instructions.html
```diff
@@ -6,7 +6,7 @@
     Open Git Machete tab:
     <ul>
       <li>Navigate <b>Git</b> (tool window) -> <b>Git Machete</b> (tab)</li>
-      <li>Or use key shortcut: Ctrl + Alt + Shift + M</li>
+      <li>Or use key shortcut: Ctrl + Alt + Shift + M (Ctrl + Option + Shift + M on Mac OS)</li>
```
PawelLipski, line 9: Pls update on Marketplace site... do you have access there?

src/main/resources/META-INF/plugin.xml
```diff
@@ -192,6 +193,8 @@
...
+            <keyboard-shortcut first-keystroke="control alt shift M" keymap="Mac OS X 10.5+" replace-all="true"/>
```
PawelLipski, line 197: Also! this is a user-facing change (fix?) so maybe update CHANGE_NOTES

CONTRIBUTING.md
```diff
@@ -55,6 +55,17 @@ ln -s ../../scripts/git-hooks/post-commit .git/hooks/post-commit
...
+#### Windows
+**The hooks do not work on Windows** (however their execution seems to be possible theoretically).
+This is because one may not be emulating bash environment in any way or doing it in some specific way.
+
+#### Mac OS
+Some hooks use `grep`. The Mac OS version of `grep` (FreeBSD) differs from GNU `grep`.
+In order to make `grep` and eventually the hooks working one must:
+1. Install `grep` via `brew` (it will not override system's `grep` - it can be executed as `ggrep`)
+2. Run `brew ls -v grep`; among the other a path like should be found `/opt/homebrew/Cellar/grep/3.7/libexec/gnubin/grep`
+3. Add the found path without `/grep` suffix to `PATH` (`/opt/homebrew/Cellar/grep/3.7/libexec/gnubin` in that case)
```
PawelLipski, line 67: I'd even specify:
```suggestion
3. Prepend the found path without `/grep` suffix to `PATH` (`/opt/homebrew/Cellar/grep/3.7/libexec/gnubin` in that case)
```
Coz appending will not have the effect, as BSD `grep` will still keep precedence

src/main/resources/META-INF/plugin.xml
```diff
@@ -8,6 +8,7 @@
     <depends>Git4Idea</depends>
 
     <extensions defaultExtensionNs="com.intellij">
+<!--        todo (#825): drop support for 2021.1 -->
```
PawelLipski, line 11: Pls add a comment next to the shortcut that `control` in `$default` keymap means `cmd` on Mac so that nobody scratches their head in future

docs/plugins.jetbrains.com/general_usage_instructions.html
PawelLipski: Updated already
image

src/main/resources/META-INF/plugin.xml
PawelLipski: (went for that option indeed)

docs/plugins.jetbrains.com/general_usage_instructions.html
mkondratek: image

PawelLipski (approved):

PR #831, opened 2022-05-10
Improvement/increase max workers

azdrojowa123:


.circleci/config.yml
```diff
@@ -36,8 +37,8 @@ jobs:
           command: ./gradlew spotlessCheck
       - run:
           name: Compile production code
-          # Given the RAM limits on CI (4GB), max-workers=2 is necessary to prevent OOMs.
-          command: ./gradlew --max-workers=2 compileJava compileKotlin
+          # Given the RAM limits on CI (8GB), max-workers=4 is necessary to prevent OOMs.
```
PawelLipski, line 40: Actually! IIRC the higher number of workers did NOT cause OOMs on Docker-large executor right?

azdrojowa123: yes, 7 workers cause OOM

PawelLipski (approved):

Issue #796, opened 2022-03-26
Merge some of the simple, single-package `frontend*` Gradle modules into one


PawelLipski:
Like `frontendVfsUtils`, `frontendDefs` etc. &mdash; TBD which ones exactly and how to call the resulting module.

PawelLipski:
Rel: #344

Issue #344, opened 2020-06-18
Group Gradle subprojects under e.g. gitCore/, backend/, frontend/ etc. top-level folders


PawelLipski:
There's getting too many top-level folders in this project :/

See https://github.com/Camelcade/Perl5-IDEA/blob/master/settings.gradle for a sample

PR #833, opened 2022-05-11
Refactor/merge folders

azdrojowa123:


frontend/ui/impl/build.gradle
```diff
@@ -0,0 +1,23 @@
...
+vavr()
+
+dependencies {
+  implementation project(':binding')
+  implementation project(':branchLayout:api')
+  implementation project(':backend:api')
+  implementation project(':frontend:compatibilityBridge')
+  implementation project(':frontend:defs')
+  implementation project(':frontend:dataKeys')
+  implementation project(':frontendGraph:api')
```
PawelLipski, line 12: `frontend:graph:api`?

frontend/ui/impl/build.gradle
```diff
@@ -0,0 +1,23 @@
...
+  implementation project(':frontend:vfsUtils')
+  implementation project(':frontendFile')
+  implementation project(':frontend:providerServices')
+  implementation project(':frontend:resourceBundles')
+
+  api project(':frontend:ui:api')
```
PawelLipski, line 18: Huh I see it was in the original `frontendUiImpl/build.gradle`... but I'm not sure why that was needed in the first place 🤔 

There's in principle no need for `...:impl` modules to expose of its dependencies as `api`, since `...:impl` are NOT being depended upon themselves... pls change this one to:

```suggestion
  implementation project(':frontend:ui:api')
```

and my hypothesis is that the project will still build.

Extra reading: https://docs.gradle.org/current/userguide/java_library_plugin.html#sec:java_library_separation, without the parts abour publishing... coz we don't publish anything to Maven/Ivy, only to the (somewhat simplistic in comparison) Jetbrains Marketplace.

settings.gradle
```diff
@@ -3,22 +3,22 @@ rootProject.name = 'git-machete-intellij-plugin'
...
+include 'backend:api'
+include 'backend:impl'
+include 'frontend:resourceBundles'
+include 'frontend:compatibilityBridge'
+include 'frontend:defs'
+include 'frontend:dataKeys'
+include 'frontend:icons'
+include 'frontend:vfsUtils'
 include 'frontendActions'
 include 'frontendFile'
```
PawelLipski, line 19: Also! is #796 to be done on this PR as well?

PawelLipski: I think it can go into a subsequent PR in fact

scripts/enforce-settings-gradle-match-directories
```diff
@@ -6,7 +6,7 @@ self_dir=$(cd "$(dirname "$0")" &>/dev/null; pwd -P)
 source "$self_dir"/utils.sh
 
 diff -u \
-  <(git ls-files */ | cut -d/ -f1 | uniq | grep -Exv '\..*|build|config|docs|gradle|.*\.hprof\..*|src|scripts') \
+  <(git ls-files --recurse-submodules */ | uniq | grep -E 'build.gradle'  | awk '$1=$1' FS="/" OFS=":" | awk '{gsub("build.gradle", "");print}' | sed 's/.$//' |  sed  's/gitCore:jGit/&\nqual/g') \
```
PawelLipski, line 9: `--recurse-submodules` recurses into submodules in the sense of [git submodule](https://git-scm.com/book/en/v2/Git-Tools-Submodules), which we don't keep/need in this project (in fact, I've only ever worked on two repos in one project that used submodules 😅). Not a bug per se, but redundant here

PawelLipski: Check

```shell
diff -u <(git ls-files --recurse-submodules */) <(git ls-files */)
```

to see there's no difference (see https://virtuslab.github.io/tips/#shell/output-comparing for `<(...)` syntax)

PawelLipski: Going further... the entire part 
```
git ls-files */ | uniq | grep -E 'build.gradle'
```

needed to list build.gradle files located **outside** the main directory can be simplified to:

```
git ls-files '**/build.gradle'
```

(again, process substitution for comparing the outputs FTW)

PawelLipski: Then another step:

```
... | awk '$1=$1' FS="/" OFS=":" | awk '{gsub("build.gradle", "");print}' | sed 's/.$//'
```

this can be replaced with less esoteric `tr` (not that `awk` is completely absent from this repo 😅 but for operations **within a single line** `sed`, `cut`, `tr` are usually better alternatives:

```
... | tr / : | sed 's/:build.gradle//'
```

PawelLipski: And finally, if it comes to missing build.gradle in `qual`... I think an empty build.gradle can be added there regardless of this script, just to make it clear that it's a regular Gradle module (even though it doesn't require any dependencies/special config)

settings.gradle
azdrojowa123: yes

PawelLipski: Unlinking then...

frontend/ui/impl/build.gradle
azdrojowa123: so maybe put every frontend module inside frontend directory ? include frontend actions ?

PawelLipski: Yeah, exactly... to be further revamped within #796 (but even then, I think all modules will stay under `frontend/`)

scripts/enforce-settings-gradle-match-directories
```diff
@@ -6,7 +6,7 @@ self_dir=$(cd "$(dirname "$0")" &>/dev/null; pwd -P)
 source "$self_dir"/utils.sh
 
 diff -u \
-  <(git ls-files --recurse-submodules */ | uniq | grep -E 'build.gradle'  | awk '$1=$1' FS="/" OFS=":" | awk '{gsub("build.gradle", "");print}' | sed 's/.$//' |  sed  's/gitCore:jGit/&\nqual/g') \
+  <(git ls-files '**/build.gradle' | uniq | tr / : | sed 's/:build.gradle//') \
```
PawelLipski, line 9: Nit: `uniq` seems redundant, full file **paths** within the repo can't possibly be duplicated (file **names** could be, but not full paths)

PawelLipski:
As a general rule... pls `Resolve` the resolved comments so that it's clear what's left to be addressed 🕵🏻 

As for this PR... it seems that everything of #344 is covered, does still it need to be a draft?

PawelLipski (approved):

PR #835, opened 2022-05-12
Inform user about not supported java version

azdrojowa123:


build.gradle
```diff
@@ -146,6 +146,17 @@ allprojects {
     mavenCentral()
   }
 
+  def checkJavaVersion = { task ->
+    if(JavaVersion.current() != JavaVersion.VERSION_11){
+       throw new GradleException("Project must be build with version 11")
```
PawelLipski, line 151:
```suggestion
       throw new GradleException("Project must be built with Java version 11")
```

build.gradle
```diff
@@ -146,6 +146,17 @@ allprojects {
...
+    }
+  }
+
+  afterEvaluate {
+    for(def task in it.tasks.build)
```
PawelLipski, line 156: What tasks are included under `it.tasks.build`... does it correspond to all dependencies of `build` task? 🤔

mkondratek:
on the other hand... ✋ can't we simply make the project compatible with a newer java?

build.gradle
azdrojowa123: it.tasks.build correspond to each module build task. So it doesn't matter if we trigger the build task from whole plugin or from another Gradle module inside them -> checkJavaVersion will be call

PawelLipski (approved):

PR #836, opened 2022-05-12
refactors frontend directories

azdrojowa123:


frontend/ui/impl/build.gradle
```diff
@@ -6,15 +6,9 @@ dependencies {
   implementation project(':binding')
   implementation project(':branchLayout:api')
   implementation project(':backend:api')
-  implementation project(':frontend:compatibilityBridge')
-  implementation project(':frontend:defs')
-  implementation project(':frontend:dataKeys')
   implementation project(':frontend:graph:api')
-  implementation project(':frontend:vfsUtils')
   implementation project(':frontend:file')
-  implementation project(':frontend:providerServices')
```
PawelLipski, line 15: If it comes to `frontend:providerServices`... I'd rather see this module's content moved into `frontend:ui:api` (as these services logicaly constitute part of the API for UI)... would it be possible without creating a cyclic dependency?

azdrojowa123: that is, in practice to remove providerServices module (and content from them put into frontend:ui:api module )?

PawelLipski:
Looks that you've included commits from the merged PR #835 into this PR 🤔 probably from wrong range passed to `git rebase`... see https://virtuslab.github.io/tips/#git/git-rebase (btw one of the main features of git-machete is guessing the right start of the range of commits to rebase (i.e. _fork point_))

PawelLipski (approved):

PR #841, opened 2022-05-16
Add gradle task for checking new eap

azdrojowa123:
and also migrate intellijVersions to properties file

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -24,6 +24,10 @@ plugins {
   id 'org.jetbrains.kotlin.jvm' version '1.6.20' apply false
 }
 
+if(JavaVersion.current() != JavaVersion.VERSION_11) {
+  throw new GradleException("Project must be built with version 11")
+}
+
 ext {
   javaMajorVersion = JavaVersion.VERSION_11
   scalaMajorVersion = '2.13'
@@ -97,6 +101,10 @@ static String buildNumberToIntellijVersion(String buildNumber) {
   return '20' + buildNumber.substring(0, 2) + '.' + buildNumber.substring(2, 3)
 }
 
+static String majorPartOfBuildNumber(String buildNumber){
+  return buildNumber.substring(0,3)
+}
+
 static String intellijVersionToBuildNumber(String version) {
   return version.substring(2, 6).replace('.', '')
 }
@@ -143,7 +151,7 @@ String checkForNewerEap(String intellijMajorVersion) {
     matcher.find()
     String foundVersion = matcher.group(1)
     if (!intellijMajorVersion){
-      if (Integer.parseInt(foundVersion.substring(0,3)) > Integer.parseInt(intellijVersionToBuildNumber(project.intellijVersions.latestStable))) {
+      if (Integer.parseInt(majorPartOfBuildNumber(foundVersion)) > Integer.parseInt(intellijVersionToBuildNumber(project.intellijVersions.latestStable))) {
         return foundVersion
       }
     } else {
@@ -153,11 +161,11 @@ String checkForNewerEap(String intellijMajorVersion) {
   return null
 }
 
-task updateEapOfLatestSupportedMajor {
+task updateEapVersion {
   Properties properties = intellijVersionsProperties()
   File propFile = intellijVersionsFile()
   String latestEap = project.intellijVersions.eapOfLatestSupportedMajor
-  String newerVersion = latestEap ? project.checkForNewerEap(latestEap.split("\\.")[0]) :
+  String newerVersion = latestEap ? project.checkForNewerEap(majorPartOfBuildNumber(latestEap)) :
             project.checkForNewerEap(/*intellijMajorVersion*/ null)
   if (newerVersion && latestEap != newerVersion) {
     properties.setProperty("eapOfLatestSupportedMajor", newerVersion)
@@ -185,18 +193,6 @@ allprojects {
   repositories {
     mavenCentral()
   }
-
-  def checkJavaVersion = {
-    if(JavaVersion.current() != JavaVersion.VERSION_11) {
-       throw new GradleException("Project must be built with version 11")
-    }
-  }
-
-  afterEvaluate {
-    for(def task in it.tasks.build)
-      task.dependsOn  checkJavaVersion
-  }
-
 
   apply plugin: 'org.gradle.java-library'
 
--- a/build.gradle
+++ b/build.gradle
@@ -142,23 +142,14 @@ import org.jsoup.nodes.Document
 
 String findNewestEapBuildNumber() {
   Document htmlContent = Jsoup.connect(project.intellijSnapshotsUrl).get()
-  Element link = htmlContent.select('a[href^='+project.intellijSnapshotsUrl+'com/jetbrains/intellij/idea/ideaIC/][href$="-EAP-SNAPSHOT.pom"]')[0]
+  Element link = htmlContent.select('a[href~=('+ project.intellijSnapshotsUrl + 'com/jetbrains/intellij/idea/ideaIC/)\\d+\\.\\d+\\.\\d+][href$="-EAP-SNAPSHOT.pom"]')[0]
   if (link) {
     String attr = link.attr("href")
-    Pattern pattern = Pattern.compile("ideaIC-(.*?)-EAP-SNAPSHOT.pom")
+    Pattern pattern = Pattern.compile("(?<=ideaIC-)\\d+\\.\\d+\\.\\d+(?=-EAP-SNAPSHOT.pom)")
     Matcher matcher = pattern.matcher(attr)
     matcher.find()
-    return matcher.group(1)
-  }
-  return null
-}
-
-String checkForEapOfFutureRelease(String intellijMajorPartOfBuildNumber) {
-  String foundBuildNumber = project.findNewestEapBuildNumber()
-  if (foundBuildNumber) {
-    if (Integer.parseInt(majorPartOfBuildNumber(foundBuildNumber)) > Integer.parseInt(intellijMajorPartOfBuildNumber)) {
-      return foundBuildNumber
-    }
+    println matcher.group()
+    return matcher.group()
   }
   return null
 }
@@ -178,8 +169,8 @@ task updateEapBuildNumber {
   File propFile = project.intellijVersionsFile()
   String latestEapBuildNumber = project.intellijVersions.eapOfLatestSupportedMajor
   String newerEapBuildNumber = latestEapBuildNumber ? project.checkForEapWithBuildNumberHigherThan(latestEapBuildNumber.split('-')[0]) :
-            project.checkForEapOfFutureRelease(intellijVersionToBuildNumber(project.intellijVersions.latestStable))
-  if (newerEapBuildNumber && latestEapBuildNumber != newerEapBuildNumber) {
+          project.checkForEapWithBuildNumberHigherThan(intellijVersionToBuildNumber(project.intellijVersions.latestStable)+'.999999.999999')
+  if (newerEapBuildNumber) {
     properties.setProperty("eapOfLatestSupportedMajor", newerEapBuildNumber+"-EAP-SNAPSHOT")
     properties.store(propFile.newWriter(), /*comments*/ null)
   }
--- a/build.gradle
+++ b/build.gradle
@@ -135,22 +135,22 @@ import org.jsoup.nodes.Document
 
 String checkForNewerEap(String intellijMajorVersion) {
   Document htmlContent = Jsoup.connect(project.intellijSnapshotsUrl).get()
-  Element link = htmlContent.select('a[href^='+project.intellijSnapshotsUrl+'com/jetbrains/intellij/idea/ideaIC/'+intellijMajorVersion+'.][href$="-EAP-SNAPSHOT.pom"]')[0]
-  if (link) {
+  Element link = htmlContent.select('a[href^='+project.intellijSnapshotsUrl+'com/jetbrains/intellij/idea/ideaIC/][href$="-EAP-SNAPSHOT.pom"]')[0]
+  if (link){
     String attr = link.attr("href")
     Pattern pattern = Pattern.compile("ideaIC-(.*?).pom")
     Matcher matcher = pattern.matcher(attr)
     matcher.find()
-    return matcher.group(1)
-  } else {
-    return null
+    String foundVersion = matcher.group(1)
+    if (!intellijMajorVersion){
+      if (Integer.parseInt(foundVersion.substring(0,3)) > Integer.parseInt(intellijVersionToBuildNumber(project.intellijVersions.latestStable))) {
+        return foundVersion
+      }
+    } else {
+      return foundVersion
+    }
   }
-}
-
-String newEapVersion() {
-  char[] latestStable = project.intellijVersions.latestStable.toCharArray()
-  return Character.getNumericValue(latestStable[5]) == 3 ? Character.toString(latestStable[2]) + Character.toString(latestStable[3] + 1) + "1":
-                Character.toString(latestStable[2]) + Character.toString(latestStable[3]) + Character.toString(latestStable[5]+1)
+  return null
 }
 
 task updateEapOfLatestSupportedMajor {
@@ -158,7 +158,7 @@ task updateEapOfLatestSupportedMajor {
   File propFile = intellijVersionsFile()
   String latestEap = project.intellijVersions.eapOfLatestSupportedMajor
   String newerVersion = latestEap ? project.checkForNewerEap(latestEap.split("\\.")[0]) :
-            project.checkForNewerEap(project.newEapVersion())
+            project.checkForNewerEap(/*intellijMajorVersion*/ null)
   if (newerVersion && latestEap != newerVersion) {
     properties.setProperty("eapOfLatestSupportedMajor", newerVersion)
     properties.store(propFile.newWriter(), /*comments*/ null)
--- a/build.gradle
+++ b/build.gradle
@@ -101,6 +101,10 @@ static String buildNumberToIntellijVersion(String buildNumber) {
   return '20' + buildNumber.substring(0, 2) + '.' + buildNumber.substring(2, 3)
 }
 
+static Integer fullIntellijVersionToNumber(String fullVersion) {
+  return Integer.parseInt(fullVersion.replace('.',''))
+}
+
 static String majorPartOfBuildNumber(String buildNumber){
   return buildNumber.substring(0,3)
 }
@@ -141,21 +145,23 @@ import org.jsoup.Jsoup
 import org.jsoup.nodes.Element
 import org.jsoup.nodes.Document
 
-String checkForNewerEap(String intellijMajorVersion) {
+String checkForEapWithVersionHigherThan(String intellijMajorVersion) {
   Document htmlContent = Jsoup.connect(project.intellijSnapshotsUrl).get()
   Element link = htmlContent.select('a[href^='+project.intellijSnapshotsUrl+'com/jetbrains/intellij/idea/ideaIC/][href$="-EAP-SNAPSHOT.pom"]')[0]
-  if (link){
+  if (link) {
     String attr = link.attr("href")
-    Pattern pattern = Pattern.compile("ideaIC-(.*?).pom")
+    Pattern pattern = Pattern.compile("ideaIC-(.*?)-EAP-SNAPSHOT.pom")
     Matcher matcher = pattern.matcher(attr)
     matcher.find()
     String foundVersion = matcher.group(1)
-    if (!intellijMajorVersion){
-      if (Integer.parseInt(majorPartOfBuildNumber(foundVersion)) > Integer.parseInt(intellijVersionToBuildNumber(project.intellijVersions.latestStable))) {
+    if (!project.intellijVersions.eapOfLatestSupportedMajor) {
+      if (Integer.parseInt(project.majorPartOfBuildNumber(foundVersion)) > Integer.parseInt(intellijMajorVersion)) {
         return foundVersion
       }
     } else {
-      return foundVersion
+      if (project.fullIntellijVersionToNumber(foundVersion) > project.fullIntellijVersionToNumber(intellijMajorVersion)) {
+        return foundVersion
+      }
     }
   }
   return null
@@ -164,11 +170,11 @@ String checkForNewerEap(String intellijMajorVersion) {
 task updateEapVersion {
   Properties properties = intellijVersionsProperties()
   File propFile = intellijVersionsFile()
-  String latestEap = project.intellijVersions.eapOfLatestSupportedMajor
-  String newerVersion = latestEap ? project.checkForNewerEap(majorPartOfBuildNumber(latestEap)) :
-            project.checkForNewerEap(/*intellijMajorVersion*/ null)
-  if (newerVersion && latestEap != newerVersion) {
-    properties.setProperty("eapOfLatestSupportedMajor", newerVersion)
+  String latestEapVersion = project.intellijVersions.eapOfLatestSupportedMajor
+  String newerVersion = latestEapVersion ? project.checkForEapWithVersionHigherThan(latestEapVersion.split('-')[0]) :
+            project.checkForEapWithVersionHigherThan(project.intellijVersionToBuildNumber(project.intellijVersions.latestStable))
+  if (newerVersion && latestEapVersion != newerVersion) {
+    properties.setProperty("eapOfLatestSupportedMajor", newerVersion+"-EAP-SNAPSHOT")
     properties.store(propFile.newWriter(), /*comments*/ null)
   }
 }
--- a/build.gradle
+++ b/build.gradle
@@ -24,8 +24,8 @@ plugins {
   id 'org.jetbrains.kotlin.jvm' version '1.6.20' apply false
 }
 
-if(JavaVersion.current() != JavaVersion.VERSION_11) {
-  throw new GradleException("Project must be built with version 11")
+if (JavaVersion.current() != JavaVersion.VERSION_11) {
+  throw new GradleException("Project must be built with Java version 11")
 }
 
 ext {
@@ -50,7 +50,6 @@ ext {
   vavrVersion = '0.10.4'
 
   intellijSnapshotsUrl = "https://www.jetbrains.com/intellij-repository/snapshots/"
-
   Properties intellijVersionsProp = intellijVersionsProperties()
 
   // See https://www.jetbrains.com/intellij-repository/releases/ -> Ctrl+F .idea
@@ -87,22 +86,12 @@ ext {
   shouldUseDelombok = project.hasProperty('useDelombok')
 }
 
-Properties intellijVersionsProperties() {
-  Properties properties = new Properties()
-  properties.load(intellijVersionsFile().newDataInputStream())
-  return properties
-}
-
-File intellijVersionsFile() {
-  return file("intellijVersions.properties")
-}
-
 static String buildNumberToIntellijVersion(String buildNumber) {
   return '20' + buildNumber.substring(0, 2) + '.' + buildNumber.substring(2, 3)
 }
 
-static Integer fullIntellijVersionToNumber(String fullVersion) {
-  return Integer.parseInt(fullVersion.replace('.',''))
+static Integer buildNumberToNumericValue(String buildNumber) {
+  return Integer.parseInt(buildNumber.replace('.',''))
 }
 
 static String majorPartOfBuildNumber(String buildNumber){
@@ -115,6 +104,16 @@ static String intellijVersionToBuildNumber(String version) {
 
 static String majorPartOfIntellijVersion(String version) {
   return version.substring(0, 6)
+}
+
+static Properties intellijVersionsProperties() {
+  Properties properties = new Properties()
+  properties.load(intellijVersionsFile().newDataInputStream())
+  return properties
+}
+
+static File intellijVersionsFile() {
+  return new File("intellijVersions.properties")
 }
 
 /**
@@ -145,7 +144,7 @@ import org.jsoup.Jsoup
 import org.jsoup.nodes.Element
 import org.jsoup.nodes.Document
 
-String checkForEapForFutureRelease(String intellijMajorVersion) {
+String findNewestEapBuildNumber() {
   Document htmlContent = Jsoup.connect(project.intellijSnapshotsUrl).get()
   Element link = htmlContent.select('a[href^='+project.intellijSnapshotsUrl+'com/jetbrains/intellij/idea/ideaIC/][href$="-EAP-SNAPSHOT.pom"]')[0]
   if (link) {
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -98,7 +110,7 @@ List<String> resolveIntelliJVersions(String versionKey) {
   }
 
   def versionValue = project.intellijVersions.get(versionKey)
-  if (versionValue == null) {
+  if (versionValue == null || !versionValue) {
```
PawelLipski, line 113: Nit: IIRC `!x` covers the case of `x == null` in Groovy, to be verified

build.gradle
```diff
@@ -107,6 +119,29 @@ List<String> resolveIntelliJVersions(String versionKey) {
   return [versionValue]
 }
 
+String checkForNewEap (String gradleMajor){
```
PawelLipski, line 122: I wouldn't call the param `gradleMajor` as it falsely suggests it's about the version of Gradle itself 🤔
Maybe `intellijMajorVersion`?

build.gradle
```diff
@@ -171,6 +211,7 @@ allprojects {
     def betterStringsCoords = [ group: 'com.antkorwin', name: 'better-strings', version: betterStringsVersion ]
     annotationProcessor(betterStringsCoords)
     testAnnotationProcessor(betterStringsCoords)
+    implementation 'org.jsoup:jsoup:1.15.1'
```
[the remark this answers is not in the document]
PawelLipski: In fact, it's needed only on **classpath of Gradle itself** (rather than classpath of any compiled module)... I'm sure there's some even easier way to add deps just to Gradle (`buildscript { ... }`?)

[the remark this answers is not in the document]
PawelLipski: Coz currently `jsoup` will be packaged into plugin zip itself, see `zipinfo -1 build/distributions/<current-version>.zip | sort`

intellijVersions.properties
```diff
@@ -0,0 +1,5 @@
+#Mon May 16 15:29:21 CEST 2022
+latestMinorsOfOldSupportedMajors=2020.1.4,2020.2.4,2020.3.4,2021.1.3,2021.2.4,2021.3.3
+eapOfLatestSupportedMajor=
+earliestSupportedMajor=2020.1
```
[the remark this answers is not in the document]
PawelLipski: But again, if that's something coming from `properties.store(...)`, then let it stay

build.gradle
```diff
@@ -107,6 +119,29 @@ List<String> resolveIntelliJVersions(String versionKey) {
...
+  File propFile = intellijVersionsFile()
+  properties.load(propFile.newDataInputStream())
+  String latest = properties.getProperty("eapOfLatestSupportedMajor")
+  if (latest){
+    String newerVersion = project.checkForNewEap(latest.split("\\.")[0])
+    if(!newerVersion.equals(latest)){
+      properties.setProperty("eapOfLatestSupportedMajor", newerVersion)
+      properties.store(propFile.newWriter(), null)
+    }
+  }
```
PawelLipski, line 142: The missing `else` here is also pretty important! Coz as for now the task will only be able to bump EAP version if some is provided, but not able to detect a new EAP :/

build.gradle
```diff
@@ -107,6 +119,29 @@ List<String> resolveIntelliJVersions(String versionKey) {
...
+task updateEapOfLatestSupportedMajor {
+  Properties properties = intellijVersionsProperties()
+  File propFile = intellijVersionsFile()
+  properties.load(propFile.newDataInputStream())
+  String latest = properties.getProperty("eapOfLatestSupportedMajor")
+  if (latest){
+    String newerVersion = project.checkForNewEap(latest.split("\\.")[0])
+    if(!newerVersion.equals(latest)){
+      properties.setProperty("eapOfLatestSupportedMajor", newerVersion)
+      properties.store(propFile.newWriter(), null)
```
PawelLipski, line 140: Name the `null` param, the meaning is unclear

build.gradle
azdrojowa123: but when eapOfLatestSupportedMajor is null (or empty), what is the meaning of the new EAP ? just the latest available ?

PawelLipski: > but when eapOfLatestSupportedMajor is null

This is the case right now, as of May 16th: we've got 2022.1 released quite recently, but an `EAP-SNAPSHOT` for the next version, **2022.2**, has NOT been published yet.

Buuuuuut... there's some foreshadowing! Check https://www.jetbrains.com/intellij-repository/snapshots/ -> Ctrl+F .idea... there's already `222.2270-EAP-CANDIDATE-SNAPSHOT` from literally yesterday. The leading `222` is shortened notation for 20**22.2** (same for all other build ids, e.g. 193 was 2019.3 etc.).

build.gradle
```diff
@@ -39,6 +45,8 @@ ext {
   slf4jVersion = '1.7.36'
   vavrVersion = '0.10.4'
 
+  intellijSnapshotsUrl = "https://www.jetbrains.com/intellij-repository/"
```
PawelLipski, line 48: Technically, it's
```suggestion
  intellijSnapshotsUrl = "https://www.jetbrains.com/intellij-repository/snapshots"
```

https://www.jetbrains.com/intellij-repository/ returns a 404 😛

build.gradle
```diff
@@ -107,6 +127,31 @@ List<String> resolveIntelliJVersions(String versionKey) {
...
+    matcher.find()
+    return matcher.group(1)
+  }
+}
+
+task updateEapOfLatestSupportedMajor {
+  Properties properties = intellijVersionsProperties()
+  File propFile = intellijVersionsFile()
+  String latestEap = properties.getProperty("eapOfLatestSupportedMajor")
+  char[] latestStable = properties.getProperty("latestStable").toCharArray()
```
PawelLipski, line 146: Would it be possible to rely on the already parsed values from `ext { ... }` in order not to re-parse the properties here? 🤔

build.gradle
```diff
@@ -107,6 +127,31 @@ List<String> resolveIntelliJVersions(String versionKey) {
...
+  Document htmlContent = Jsoup.connect(project.intellijSnapshotsUrl).get()
+  Element link = htmlContent.select('a[href^='+project.intellijSnapshotsUrl+'com/jetbrains/intellij/idea/ideaIC/'+intellijMajorVersion+'.][href$="-EAP-SNAPSHOT.pom"]')[0]
+  if (link) {
+    String attr = link.attr("href")
+    Pattern pattern = Pattern.compile("ideaIC-(.*?).pom")
+    Matcher matcher = pattern.matcher(attr)
+    matcher.find()
+    return matcher.group(1)
+  }
+}
```
PawelLipski, line 140: Hmm as IntelliJ points out here: not all execution paths return a value... pls return a `null` explicitly if link has not been found

build.gradle
```diff
@@ -107,6 +127,31 @@ List<String> resolveIntelliJVersions(String versionKey) {
...
+
+task updateEapOfLatestSupportedMajor {
+  Properties properties = intellijVersionsProperties()
+  File propFile = intellijVersionsFile()
+  String latestEap = project.intellijVersions.eapOfLatestSupportedMajor
+  char[] latestStable = project.intellijVersions.latestStable.toCharArray()
+  String newerVersion = latestEap ? project.checkForNewEap(latestEap.split("\\.")[0]) :
+            project.checkForNewEap(Character.toString(latestStable[2]) + Character.toString(latestStable[3]) + Character.toString(latestStable[5]+1))
```
PawelLipski, line 148: Hmmm the problem with `+1` here is that e.g. after `213` (2021.3) the next version was `221` (2022.1), not `214`.

PawelLipski: As a remedy, I'd rework the logic in `checkForNewEap` to check for a version **equal or higher** to the specified `intellijMajorVersion`

PawelLipski: Also! note that you've got `intellijVersionToBuildNumber` helper for converting `latestStable` to build number

build.gradle
```diff
@@ -98,7 +118,7 @@ List<String> resolveIntelliJVersions(String versionKey) {
   }
 
   def versionValue = project.intellijVersions.get(versionKey)
-  if (versionValue == null) {
+  if (!versionValue) {
```
PawelLipski, line 121: Actually... am I right that this method is actually only used for `resolveIntelliJVersions` task, which itself is only used in a single script under `scripts/`? 🤔 I think now that versions are stored in properties, that script can just rely on `grep` over properties... and this method/task can be removed

azdrojowa123: yes task can be removed (and the script needs to be changed), but this method resolveIntelliJVersions is also used in resolving ui test targets (line 658)

build.gradle
```diff
@@ -141,6 +186,12 @@ dependencyUpdates {
 import org.gradle.api.tasks.testing.logging.TestExceptionFormat
 import org.gradle.api.tasks.testing.logging.TestLogEvent
 
+import java.util.regex.Matcher
```
PawelLipski, line 189: Nit: place these imports somewhere **before** their use (I didn't even know Gradle/Groovy allows for importing **after** the use 😯)

build.gradle
```diff
@@ -107,23 +127,43 @@ List<String> resolveIntelliJVersions(String versionKey) {
...
   }
+}
 
-  @TaskAction
-  void print() {
-    println(String.join("\n", project.resolveIntelliJVersions(versionKey)))
-  }
+String newEapVersion() {
+  char[] latestStable = project.intellijVersions.latestStable.toCharArray()
+  return Character.getNumericValue(latestStable[5]) == 3 ? Character.toString(latestStable[2]) + Character.toString(latestStable[3] + 1) + "1":
```
PawelLipski, line 152: Pls note that in general we should not rely on there being **3** major releases with the given year (although it's been a practice so far)... hence I suggested filtering the available EAP versions for one that is **higher** than the provided one! Rather than looking for a specific hand-crafted version via a regex, as done now 🤔

scripts/validate-markdown-values
```diff
@@ -3,12 +3,13 @@
 set -e -o pipefail -u
 
 self_dir=$(cd "$(dirname "$0")" &>/dev/null; pwd -P)
+root_dir=$(cd $PWD &>/dev/null; pwd -P)
 source "$self_dir"/utils.sh
+source "$root_dir"/intellijVersions.properties
```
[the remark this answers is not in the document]
PawelLipski: Although... in principle, I'd be careful, as syntax of .properties files is much _dumber_... many things that are just plain strings in .properties will be interpreted differently by Bash:

```properties
never_execute_this_command=$(rm -rf --no-preserve-root /)
```

So still, I'd use `grep -Po` with a lookbehind (see below + https://virtuslab.github.io/tips/#regex/lookbehind-lookahead) to extract the value instead of sourcing the properties file!

build.gradle
```diff
@@ -107,24 +127,44 @@ List<String> resolveIntelliJVersions(String versionKey) {
...
+}
 
-  @TaskAction
-  void print() {
-    println(String.join("\n", project.resolveIntelliJVersions(versionKey)))
+task updateEapOfLatestSupportedMajor {
+  Properties properties = intellijVersionsProperties()
+  File propFile = intellijVersionsFile()
+  String latestEap = project.intellijVersions.eapOfLatestSupportedMajor
+  String newerVersion = latestEap ? project.checkForNewerEap(latestEap.split("\\.")[0]) :
```
PawelLipski, line 160: And for consistency, IIUC `majorPartOfBuildNumber` can be used here as well

build.gradle
```diff
@@ -39,22 +45,26 @@ ext {
   slf4jVersion = '1.7.36'
   vavrVersion = '0.10.4'
```
PawelLipski, line 47: Btw I'm now re-checking the part about Java 11 check... isn't it just enough simply ditch the `task` and `afterEvaluate` part, and just put

```
if(JavaVersion.current() != JavaVersion.VERSION_11) {
  throw new GradleException("Project must be built with version 11")
}
```

directly as top-level instructions in build.gradle? 🤔

build.gradle
```diff
@@ -107,24 +127,44 @@ List<String> resolveIntelliJVersions(String versionKey) {
...
+      return foundVersion
+    }
   }
+  return null
+}
 
-  @TaskAction
-  void print() {
-    println(String.join("\n", project.resolveIntelliJVersions(versionKey)))
+task updateEapOfLatestSupportedMajor {
```
PawelLipski, line 156: Also! this task's name is too specific; it can not only update EAP of latest supported major, but only find an EAP of a major version that is NOT supported yet (like 2022.2 currently)

build.gradle
```diff
@@ -107,24 +127,44 @@ List<String> resolveIntelliJVersions(String versionKey) {
...
-  @OptionValues('version-key')
-  List<String> getAvailableVersionKeys() {
-    return new ArrayList<>(project.intellijVersions.keySet())
+import java.util.regex.Matcher
+import java.util.regex.Pattern
+import org.jsoup.Jsoup
+import org.jsoup.nodes.Element
+import org.jsoup.nodes.Document
+
+String checkForNewerEap(String intellijMajorVersion) {
```
PawelLipski, line 136: I'd specify the name, IIUC it should be sth like:
```suggestion
String checkForEapForMajorVersionHigherThan(String intellijMajorVersion) {
```

build.gradle
azdrojowa123: yess, from what I checked, we have the same result

build.gradle
azdrojowa123: yes this method name is also better for me, but I don't know exactly what it would look like with the word "then" in a situation where eapOfLatestSupportedMajor is null in properties file -> in this situation I also pass null as method param so we have checkForEapForMajorVersionHigherThan (null) ..?

PawelLipski: I'd do it so (pseudocode):

```python
if latestEap != null:
  newerEap = project.checkForEapForMajorVersionHigherThan(latestEap) // we're checking the *full* EAP version! so comparing e.g `222.2137.1 > 222.1920.123`
else:
  newerEap = project.checkForEapForMajorVersionHigherThan(intellijVersionToBuildNumber(latestStable)) // here we'd compare e.g. `222.2137.1 > 221`, smart comparison algorithm should be devised for such version comparison - maybe there's something ready in Gradle?

File: build.gradle (hunk 15 above)
Comment (azdrojowa123) [reply build.gradle:160]: hmm but when latestEap exists we need to pass whole number e.q. 221.5080.93 to checkForNewerEap method not only 221 (if i'm right?) we need to pass only 221 (from latest stable) if latestEap not exist

Comment (PawelLipski) [reply build.gradle:160]: Yes, indeed, the comment https://github.com/VirtusLab/git-machete-intellij-plugin/pull/841#discussion_r875993057 somewhat invalidates this one ;)

File: build.gradle [hunk 19]
```diff
@@ -107,24 +139,46 @@ List<String> resolveIntelliJVersions(String versionKey) {
...
+String checkForEapWithVersionHigherThan(String intellijMajorVersion) {
+  Document htmlContent = Jsoup.connect(project.intellijSnapshotsUrl).get()
+  Element link = htmlContent.select('a[href^='+project.intellijSnapshotsUrl+'com/jetbrains/intellij/idea/ideaIC/][href$="-EAP-SNAPSHOT.pom"]')[0]
+  if (link) {
+    String attr = link.attr("href")
+    Pattern pattern = Pattern.compile("ideaIC-(.*?)-EAP-SNAPSHOT.pom")
+    Matcher matcher = pattern.matcher(attr)
+    matcher.find()
+    String foundVersion = matcher.group(1)
+    if (!project.intellijVersions.eapOfLatestSupportedMajor) {
```
PawelLipski, line 157: Nit: revert the condition for better readability (i.e. avoid negation in condition where possible)

build.gradle
```diff
@@ -107,24 +139,46 @@ List<String> resolveIntelliJVersions(String versionKey) {
...
+      if (Integer.parseInt(project.majorPartOfBuildNumber(foundVersion)) > Integer.parseInt(intellijMajorVersion)) {
+        return foundVersion
+      }
+    } else {
+      if (project.fullIntellijVersionToNumber(foundVersion) > project.fullIntellijVersionToNumber(intellijMajorVersion)) {
```
PawelLipski, line 162: Hmmm won't `foundVersion` be of the format `222.1234.56`? and `intellijMajorVersion` of just `221`? Then `222123456 > 221` indeed... but I don't think that's the comparison we'd like to perform here!

PawelLipski: Oh, okay.... `intellijMajorVersion` will be of `222.789.12` format when we're enter this case... still, this logic is pretty convoluted... it must be simplified, lemme suggest something...

azdrojowa123: but when intellijMajorVersion will be 221 (so taking from latestStable) we have first case, not this one

PawelLipski: Oh, okay.... `intellijMajorVersion` will be of `222.789.12` format when we're enter this case... still, this logic is pretty convoluted... it must be simplified, lemme suggest something:

Let's make sure this method does NOT depend on `project.intellijVersions` since it complicates the reasoning a lot. If needed, split into two separate method to be referenced from `String newerVersion = latestEapVersion ? ... : ...`

scripts/validate-markdown-values
```diff
@@ -3,12 +3,11 @@
...
 self_dir=$(cd "$(dirname "$0")" &>/dev/null; pwd -P)
+root_dir=$(cd $PWD &>/dev/null; pwd -P)
 source "$self_dir"/utils.sh
-
 # Currently, just check whether earliest supported IntelliJ version in README.md
 # is up to date with the source of truth (Gradle config).
-
-expected=$(./gradlew -q resolveIntelliJVersions --version-key=earliestSupportedMajor)
+expected=$(grep -Po '(?<=earliestSupportedMajor=)[0-9.]*[0-9]' "$root_dir"/intellijVersions.properties)
 actual=$(grep -Po '(?<=minimum required version is )[0-9.]*[0-9]' README.md)
```
PawelLipski, line 11: While you're here, use `root_dir` here for consistency as well

build.gradle
PawelLipski: > but when intellijMajorVersion will be 221 (so taking from latestStable) we have first case, not this one

Yeah I see now... but that's a silent assumption that's hard to figure out from code ☹️ let's rework the method to make it clear what format of version is expected where!

scripts/validate-markdown-values
azdrojowa123: If wee need this `source "$self_dir"/utils.sh` ? because I don't see any use (?)

PawelLipski: `error` and `die` ☠️

azdrojowa123: aa okey :) !

build.gradle
```diff
@@ -18,6 +24,10 @@ plugins {
   id 'org.jetbrains.kotlin.jvm' version '1.6.20' apply false
 }
 
+if(JavaVersion.current() != JavaVersion.VERSION_11) {
```
PawelLipski, line 27: Nit:
```suggestion
if (JavaVersion.current() != JavaVersion.VERSION_11) {
```

build.gradle
```diff
@@ -18,6 +24,10 @@ plugins {
...
+  throw new GradleException("Project must be built with version 11")
```
PawelLipski, line 28: Nit:
```suggestion
  throw new GradleException("Project must be built with Java version 11")
```

build.gradle
```diff
@@ -107,23 +139,55 @@ List<String> resolveIntelliJVersions(String versionKey) {
...
-  @OptionValues('version-key')
-  List<String> getAvailableVersionKeys() {
-    return new ArrayList<>(project.intellijVersions.keySet())
+import java.util.regex.Matcher
+import java.util.regex.Pattern
+import org.jsoup.Jsoup
+import org.jsoup.nodes.Element
+import org.jsoup.nodes.Document
+
+String checkForEapForFutureRelease(String intellijMajorVersion) {
```
PawelLipski, line 148: For moar clarity:
```suggestion
String checkForEapForFutureRelease(String intellijMajorPartOfBuildNumber) {
```
(at least I suppose that's what's expected here!)

build.gradle
```diff
@@ -107,23 +139,55 @@ List<String> resolveIntelliJVersions(String versionKey) {
...
+  Document htmlContent = Jsoup.connect(project.intellijSnapshotsUrl).get()
+  Element link = htmlContent.select('a[href^='+project.intellijSnapshotsUrl+'com/jetbrains/intellij/idea/ideaIC/][href$="-EAP-SNAPSHOT.pom"]')[0]
+  if (link) {
```
PawelLipski, line 151: Pls [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) up these two methods, i.e. extract the shared part

In fact a helper method could return what's now `foundVersion`!

build.gradle
```diff
@@ -107,23 +139,55 @@ List<String> resolveIntelliJVersions(String versionKey) {
...
+    String attr = link.attr("href")
+    Pattern pattern = Pattern.compile("ideaIC-(.*?)-EAP-SNAPSHOT.pom")
+    Matcher matcher = pattern.matcher(attr)
+    matcher.find()
+    String foundVersion = matcher.group(1)
```
PawelLipski, line 156: Same here wrt. naming, this should be
```suggestion
    String foundBuildNumber = matcher.group(1)
```
(we need to clearly distinguish b/w `2022.2`- and `222`-formatted version... or  we **p e r i s h**)

build.gradle
```diff
@@ -107,23 +139,55 @@ List<String> resolveIntelliJVersions(String versionKey) {
...
+    }
   }
+  return null
 }
 
-tasks.register('resolveIntelliJVersions', ResolveIntelliJVersions)
+task updateEapVersion {
+  Properties properties = intellijVersionsProperties()
+  File propFile = intellijVersionsFile()
+  String latestEapVersion = project.intellijVersions.eapOfLatestSupportedMajor
```
PawelLipski, line 183: Pls apply the `buildNumber` naming wherever applicable in your new code

scripts/validate-markdown-values
```diff
@@ -3,13 +3,12 @@
 set -e -o pipefail -u
 
 self_dir=$(cd "$(dirname "$0")" &>/dev/null; pwd -P)
+root_dir=$(cd $PWD &>/dev/null; pwd -P)
```
PawelLipski, line 6: Hmmm I'd rather actually use
```suggestion
root_dir=$(cd "$self_dir/.." &>/dev/null; pwd -P)
```
Coz `$PWD` (current working directory) does NOT need to be the root directory of the project when someone executes this script

build.gradle
azdrojowa123: so when we have numbers without dots (for example 222 or 221345123) -> is it build number, and when when we have for example 2022.2 or 221.345.123 it is version (in the context of the naming) ?

azdrojowa123: so because I can't quite understand it - when here should be ...buildNumber and in line 183 also should be ...buildNumber, so when when should we have version ?

PawelLipski:
```
221.1234.5 - build number
221 - major part of 221.1234.5 build number
2022.1 - release (aka version)
2022.1.1 - also release (aka version)
2022.1 - major part of both 2022.1 and 2021.1.1 releases
```

azdrojowa123: version is only reserved for 2020.1 etc. ?

PawelLipski: Just a matter of different representation of the same thing! `2021.1`-style versions are used more (but not only) for marketing purposes, `211`-style build numbers are used more internally by JetBrains

build.gradle
```diff
@@ -77,6 +91,14 @@ static String buildNumberToIntellijVersion(String buildNumber) {
   return '20' + buildNumber.substring(0, 2) + '.' + buildNumber.substring(2, 3)
 }
 
+static Integer buildNumberToNumericValue(String buildNumber) {
```
PawelLipski, line 94: Hmmm I'm skeptical about this conversion... and esp. about comparing EAP versions in such a way!

Consider comparing `222.137.42` with `222.99.999`.
Although it's the **former** that's actually higher (coz 137 > 99), this approach will consider the **latter** higher (coz 22213742 < 22299999) :/

Consider https://stackoverflow.com/a/36400405/2750812 (it'll serve at least as long as we're on Gradle 7.x, later to be re-implemented/replaced with sth else)

azdrojowa123: a okey in fact, I did not include this case :/

build.gradle
```diff
@@ -107,23 +134,56 @@ List<String> resolveIntelliJVersions(String versionKey) {
...
+  return null
+}
 
-  @OptionValues('version-key')
-  List<String> getAvailableVersionKeys() {
-    return new ArrayList<>(project.intellijVersions.keySet())
+String checkForEapOfFutureRelease(String intellijMajorPartOfBuildNumber) {
+  String foundBuildNumber = project.findNewestEapBuildNumber()
+  if (foundBuildNumber) {
+    if (Integer.parseInt(majorPartOfBuildNumber(foundBuildNumber)) > Integer.parseInt(intellijMajorPartOfBuildNumber)) {
```
PawelLipski, line 159: Actually... would `VersionNumber.parse` work here as well?

As in, is would

```
VersionNumber.parse(foundBuildNumber) > VersionNumber.parse(intellijMajorPartOfBuildNumber)
```

be true if `foundBuildNumber == "222.123.456"` and `intellijMajorPartOfBuildNumber == "221"`?

PawelLipski: Coz if so... then `checkForEapOfFutureRelease` and `checkForEapWithBuildNumberHigherThan` can literally be merged into a single function!

PawelLipski: Or rather, the current form of `checkForEapWithBuildNumberHigherThan` can be used for both cases (`latestEapBuildNumber` being null or non-null), just with a different param 🤔

azdrojowa123: yes, but when we pass null (in case when `eapOfLatestSupportedMajor` is empty in properties file) I will need to call project.intellijVersion in `checkForEapWithBuildNumberHigherThan ` method, it will be okey ?

azdrojowa123: so I can add additional param to this method or just check in this if `latestEapBuildNumber` is not empty from `project.intellijVersions`

build.gradle
```diff
@@ -107,23 +134,56 @@ List<String> resolveIntelliJVersions(String versionKey) {
...
+  return null
 }
 
-tasks.register('resolveIntelliJVersions', ResolveIntelliJVersions)
+task updateEapBuildNumber {
+  Properties properties = project.intellijVersionsProperties()
+  File propFile = project.intellijVersionsFile()
+  String latestEapBuildNumber = project.intellijVersions.eapOfLatestSupportedMajor
+  String newerEapBuildNumber = latestEapBuildNumber ? project.checkForEapWithBuildNumberHigherThan(latestEapBuildNumber.split('-')[0]) :
+            project.checkForEapOfFutureRelease(intellijVersionToBuildNumber(project.intellijVersions.latestStable))
```
PawelLipski, line 181: Would something like that work? 🤔 
```suggestion
  String newerEapBuildNumber = latestEapBuildNumber ? project.checkForEapWithBuildNumberHigherThan(latestEapBuildNumber.split('-')[0]) :
            project.checkForEapWithBuildNumberHigherThan(intellijVersionToBuildNumber(project.intellijVersions.latestStable) + '.9999999.9999999')
```

build.gradle
```diff
@@ -107,23 +134,56 @@ List<String> resolveIntelliJVersions(String versionKey) {
...
-  String versionKey
+import java.util.regex.Matcher
+import java.util.regex.Pattern
+import org.jsoup.Jsoup
+import org.jsoup.nodes.Element
+import org.jsoup.nodes.Document
+
+String findNewestEapBuildNumber() {
+  Document htmlContent = Jsoup.connect(project.intellijSnapshotsUrl).get()
+  Element link = htmlContent.select('a[href^='+project.intellijSnapshotsUrl+'com/jetbrains/intellij/idea/ideaIC/][href$="-EAP-SNAPSHOT.pom"]')[0]
```
PawelLipski, line 145: Oh! I did a quick experiment here... replaced `latestStable` in properties with `2021.3.3` (deliberately too low a version, so that some 2022.1 EAP gets picked up) and ended up with:

```
eapOfLatestSupportedMajor=221-EAP-SNAPSHOT
```

While this is technically a correct value... see `// Note that we have to use a "fixed snapshot"` comment in build.gradle!

I'd refine the regexes to only capture `X.Y.Z-EAP-SNAPSHOT`-style build numbers, and not `X-EAP-SNAPSHOT` ones

PawelLipski, line 145: Oh! I did a quick experiment here... replaced `latestStable` in properties with `2021.3.3` (deliberately too low a version, so that some 2022.1 EAP gets picked up) and ended up with:

```
eapOfLatestSupportedMajor=221-EAP-SNAPSHOT
```

While this is technically a correct value... see `// Note that we have to use a "fixed snapshot"` comment in build.gradle!

I'd refine the regexes to only capture `X.Y.Z-EAP-SNAPSHOT`-style build numbers, and not `X-EAP-SNAPSHOT` ones

build.gradle
```diff
@@ -107,23 +134,47 @@ List<String> resolveIntelliJVersions(String versionKey) {
...
-    return new ArrayList<>(project.intellijVersions.keySet())
+import java.util.regex.Matcher
+import java.util.regex.Pattern
+import org.jsoup.Jsoup
+import org.jsoup.nodes.Element
+import org.jsoup.nodes.Document
+
+String findNewestEapBuildNumber() {
+  Document htmlContent = Jsoup.connect(project.intellijSnapshotsUrl).get()
+  Element link = htmlContent.select('a[href~=('+ project.intellijSnapshotsUrl + 'com/jetbrains/intellij/idea/ideaIC/)\\d+\\.\\d+\\.\\d+][href$="-EAP-SNAPSHOT.pom"]')[0]
```
PawelLipski, line 145: Nit: the parens are probably redundant, as `+` binds stronger than "neighbourhood"/concatention: see https://virtuslab.github.io/tips/#regex/operators-precedence

build.gradle
```diff
@@ -107,23 +134,47 @@ List<String> resolveIntelliJVersions(String versionKey) {
...
+  if (foundBuildNumber) {
+    if (VersionNumber.parse(foundBuildNumber) > VersionNumber.parse(latestEapBuildNumber)) {
+      return foundBuildNumber
+    }
   }
+  return null
 }
 
-tasks.register('resolveIntelliJVersions', ResolveIntelliJVersions)
+task updateEapBuildNumber {
```
PawelLipski, line 167: Is this task now executed automagically every time `./gradlew` is run? 😯 at least that's what I'm observing locally 🤔

PawelLipski, line 167: Is this task now executed automagically every time `./gradlew` is run? 😯 at least that's what I'm observing locally 🤔

azdrojowa123: yes, I need to repair it

PawelLipski: Actually this behavior is pretty surprising for me, looks like sth I don't understand about Gradle task model... pls check what's the reason for auto-execution of this task while you're here

build.gradle
```diff
@@ -165,14 +164,16 @@ String checkForEapWithBuildNumberHigherThan(String latestEapBuildNumber) {
...
 task updateEapBuildNumber {
-  Properties properties = project.intellijVersionsProperties()
-  File propFile = project.intellijVersionsFile()
-  String latestEapBuildNumber = project.intellijVersions.eapOfLatestSupportedMajor
-  String newerEapBuildNumber = latestEapBuildNumber ? project.checkForEapWithBuildNumberHigherThan(latestEapBuildNumber.split('-')[0]) :
-          project.checkForEapWithBuildNumberHigherThan(intellijVersionToBuildNumber(project.intellijVersions.latestStable)+'.999999.999999')
-  if (newerEapBuildNumber) {
-    properties.setProperty("eapOfLatestSupportedMajor", newerEapBuildNumber+"-EAP-SNAPSHOT")
-    properties.store(propFile.newWriter(), /*comments*/ null)
+  doLast {
```
PawelLipski, line 167: 😯 wow so weird... so task body without `doLast` is auto-executed?

build.gradle
azdrojowa123: yes I discovered that Gradle has configuration and execution phase and configuration phase is executed every time `./gradlew `is run and execution phase only when it is _executed_. Without separate this parts, everything directly inside task  is running during configuration phase. (this is what happened in our case)

build.gradle
azdrojowa123: yes - everything outside doLast is auto-executed during configuration phase (so when .`/gradlew `is run)

build.gradle
```diff
@@ -77,6 +90,10 @@ static String buildNumberToIntellijVersion(String buildNumber) {
   return '20' + buildNumber.substring(0, 2) + '.' + buildNumber.substring(2, 3)
 }
 
+static String majorPartOfBuildNumber(String buildNumber){
```
PawelLipski, line 93: Seems unused!

build.gradle
```diff
@@ -107,23 +134,48 @@ List<String> resolveIntelliJVersions(String versionKey) {
...
+  return null
 }
 
-tasks.register('resolveIntelliJVersions', ResolveIntelliJVersions)
+task updateEapBuildNumber {
+  doLast {
+    Properties properties = project.intellijVersionsProperties()
+    File propFile = project.intellijVersionsFile()
+    String latestEapBuildNumber = project.intellijVersions.eapOfLatestSupportedMajor
+    String newerEapBuildNumber = latestEapBuildNumber ? project.checkForEapWithBuildNumberHigherThan(latestEapBuildNumber.split('-')[0]) :
```
PawelLipski, line 171: Nit: for clarity:
```suggestion
    String newerEapBuildNumber = latestEapBuildNumber ? project.checkForEapWithBuildNumberHigherThan(latestEapBuildNumber.replace('-EAP-SNAPSHOT', '')[0]) :
```

PawelLipski (approved):
🚀 minor comments left, otherwise ready to merge ‼️

PR #849, opened 2022-05-19
Feature/848 action names improvments

mkondratek:
~Please hold on with reviews - I see that the name for sync by merge from the context menu (but on the current branch) is wrong.~

Please review the other messages in the bundle (lets try to do it once  and for all).

Once we agree on the messages we can update the docs (readme, and scenario recordings if we decide to).
What do you think about that?

I would rather update only the text references (and leave the scenario recordings for another issue). 😆 

image

### DIFF
--- a/frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseMergeParentIntoBranchAction.java
+++ b/frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseMergeParentIntoBranchAction.java
@@ -22,6 +22,7 @@ import org.checkerframework.checker.i18nformatter.qual.I18nFormat;
 import com.virtuslab.gitmachete.backend.api.SyncToParentStatus;
 import com.virtuslab.gitmachete.frontend.actions.common.MergeProps;
 import com.virtuslab.gitmachete.frontend.actions.contextmenu.CheckoutSelectedBranchAction;
+import com.virtuslab.gitmachete.frontend.defs.ActionPlaces;
 import com.virtuslab.qual.guieffect.UIThreadUnsafe;
 
 @CustomLog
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,4 +1,4 @@
-# Git Machete IntelliJ Plugin
+# Git Machete IntelliJ  Plugin
```
mkondratek, line 1: A changes to the scenario descriptions are needed; and new recordings probably 😕

PawelLipski: At least the former! although the moment for the latter has also probably come... given the recent change in icons

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseFastForwardMergeBranchToParentAction.java
```diff
@@ -53,6 +54,17 @@ public List<SyncToParentStatus> getEligibleStatuses() {
   protected void onUpdate(AnActionEvent anActionEvent) {
     super.onUpdate(anActionEvent);
     syncToParentStatusDependentActionUpdate(anActionEvent);
+    val presentation = anActionEvent.getPresentation();
```
PawelLipski, line 57: Huh why this new piece of logic is needed? 🤔

PawelLipski: Was it missing **regardless** of this action name improvements?

mkondratek: n/a (I ve modified the wrong class, it was late xd)

README.md
mkondratek: I will cover changes to README.md text in this one then. For the rest I have created: 
https://github.com/VirtusLab/git-machete-intellij-plugin/issues/850

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseMergeParentIntoBranchAction.java
```diff
@@ -55,6 +56,17 @@ public List<SyncToParentStatus> getEligibleStatuses() {
   protected void onUpdate(AnActionEvent anActionEvent) {
     super.onUpdate(anActionEvent);
     syncToParentStatusDependentActionUpdate(anActionEvent);
+    val presentation = anActionEvent.getPresentation();
```
PawelLipski, line 59: Okay, still the same point... tl;dr how was/would the behavior be wrong if it weren't for this change?

mkondratek: from a context menu, you can pick the currently checked out branch ofc. In that case, the action name "Checkout and do smth" sounds silly and is inaccurate b/c the branch is already checked out

PawelLipski: Awww okay 🤦🏻

PawelLipski (approved):

PR #856, opened 2022-05-21
Release v2.0.0

PawelLipski:
~A draft, coz still a few issues/PRs need to be closed~ no longer!


mkondratek (approved):
yeah. we should do releases more often... and smaller. no way to review it 😆

PR #860, opened 2022-05-23
Repair updateEapBuildNumber task

azdrojowa123:


build.gradle
```diff
@@ -134,24 +134,18 @@ import java.util.regex.Matcher
 import java.util.regex.Pattern
 import org.jsoup.Jsoup
 import org.jsoup.nodes.Element
+import org.jsoup.select.Elements
 import org.jsoup.nodes.Document
 
-String findNewestEapBuildNumber() {
+String checkForEapWithBuildNumberHigherThan(String latestEapBuildNumber) {
   Document htmlContent = Jsoup.connect(project.intellijSnapshotsUrl).get()
-  Element link = htmlContent.select('a[href~=' + project.intellijSnapshotsUrl + 'com/jetbrains/intellij/idea/ideaIC/\\d+\\.\\d+\\.\\d+][href$="-EAP-SNAPSHOT.pom"]')[0]
```
PawelLipski, line 141: Oh! indeed, I didn't notice that before...

build.gradle
```diff
@@ -134,24 +134,18 @@ import java.util.regex.Matcher
...
-  }
-  return null
-}
-
-String checkForEapWithBuildNumberHigherThan(String latestEapBuildNumber) {
-  String foundBuildNumber = project.findNewestEapBuildNumber()
-  if (foundBuildNumber) {
+    String foundBuildNumber = matcher.group()
     if (VersionNumber.parse(foundBuildNumber) > VersionNumber.parse(latestEapBuildNumber)) {
       return foundBuildNumber
```
PawelLipski, line 150: Actually! while you're here, pls make sure that `updateEapBuildNumber` prints a message (not sure if it should be a plain `println` or some `log.info`? TBD what's the convention) when EAP is updated (so that when running locally, one doesn't need to check `git diff` right after running the task)

PawelLipski (approved):

build.gradle
azdrojowa123: you think about printing only new build number or display more human readable message ?

PawelLipski: Human readable message! as the machine-readable changes are already performed on the properties file

PawelLipski (approved):

build.gradle
```diff
@@ -134,24 +134,18 @@ import java.util.regex.Matcher
...
 import org.jsoup.nodes.Element
+import org.jsoup.select.Elements
 import org.jsoup.nodes.Document
 
-String findNewestEapBuildNumber() {
+String checkForEapWithBuildNumberHigherThan(String latestEapBuildNumber) {
   Document htmlContent = Jsoup.connect(project.intellijSnapshotsUrl).get()
-  Element link = htmlContent.select('a[href~=' + project.intellijSnapshotsUrl + 'com/jetbrains/intellij/idea/ideaIC/\\d+\\.\\d+\\.\\d+][href$="-EAP-SNAPSHOT.pom"]')[0]
-  if (link) {
+  Elements links = htmlContent.select('a[href~=' + project.intellijSnapshotsUrl + 'com/jetbrains/intellij/idea/ideaIC/\\d+\\.\\d+\\.\\d+][href$="-EAP-SNAPSHOT.pom"]')
```
PawelLipski, line 142: Nit: coz you now do the regex filtering on each individual link anyway, this can be probably reduced/simplified to:
```suggestion
  Elements links = htmlContent.select('a[href^=' + project.intellijSnapshotsUrl + 'com/jetbrains/intellij/idea/ideaIC/][href$="-EAP-SNAPSHOT.pom"]')
```

PawelLipski: Albeit it's a matter of taste, up to you

PR #863, opened 2022-05-24
Refactor/rename merge parent into current

azdrojowa123:


frontend/base/src/main/resources/GitMacheteBundle.properties
```diff
@@ -318,8 +318,8 @@ action.GitMachete.FastForwardMergeCurrentBranchToParentAction.GitMacheteToolbar.
 action.GitMachete.FastForwardMergeCurrentBranchToParentAction.description=Fast-forward merge current branch to parent
 
 
-action.GitMachete.MergeParentIntoCurrentBranchAction.text=Git Machete: Merge Parent into Current Branch
-action.GitMachete.MergeParentIntoCurrentBranchAction.GitMacheteToolbar.text=Merge Parent into Current Branch
+action.GitMachete.MergeParentIntoCurrentBranchAction.text=Git Machete: Sync to Parent by Merge
+action.GitMachete.MergeParentIntoCurrentBranchAction.GitMacheteToolbar.text=Sync to Parent by Merge
```
PawelLipski, line 322: While we're here... are the descriptions generally consistent for other actions, esp. Rebase? They don't need to be identical, just it should be clear that the action in toolbar and in context menu correspond to the same thing!

mkondratek (approved):

frontend/base/src/main/resources/GitMacheteBundle.properties
PawelLipski: Plus as discussed, pls document:
* `...` usage
* context menu/toolbar action name consistency
in contributing guidelines

mkondratek: good point. I see 
`action.GitMachete.RebaseCurrentBranchOntoParentAction.description=Rebase current branch onto parent`
in particular. should be updated too

mkondratek (changes requested):

frontend/base/src/main/resources/GitMacheteBundle.properties
azdrojowa123: On this icon bar sometimes it is mention that `do sth ... on current branch` and sometimes not e.q. `Fast Forward merge Current branch into parent` and `Sync to Parent by rebase` (if we want to stick to the convention, it should probably also be information that current branch ... or delete this information from `Fast forward merge ...`)

azdrojowa123: This situation is only with `Sync to Parent by rebase` and` sync to parent by merge` (no current branch information )

PawelLipski (approved):

PR #905, opened 2022-06-06
refactor new line logic in FF merge notification

azdrojowa123:


frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/common/FastForwardMerge.java
```diff
@@ -38,7 +38,7 @@ private static void mergeNonCurrentBranch(Project project,
     val successFFMergeNotification = fetchNotificationPrefix
         + getString("action.GitMachete.BaseFastForwardMergeBranchToParentAction.notification.title.ff-success")
             .format(stayingName, movingName);
-    val failFFMergeNotification = fetchNotificationPrefix
+    val failFFMergeNotification = fetchNotificationPrefix.replaceAll("<br/>", getString("utils.new-line-character"))
```
PawelLipski, line 41: Any specific reason for extracting `\n` to properties? 🤔 esp. given that `<br/>` isn't

PawelLipski (approved):

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/common/FastForwardMerge.java
azdrojowa123: Screenshot 2022-06-06 at 14 29 47

😞

azdrojowa123: On UNIX systems it should return \n, but on Windows it might return a different value

PawelLipski: :ups: right indeed... just `@SuppressWarnings("regex")` then

PawelLipski: Or indeed, use `System.lineSeparator()` 🤔

azdrojowa123: but on Windows System.lineSeparator can return “\r\n”  instead of \n

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/common/FastForwardMerge.java
```diff
@@ -27,6 +27,7 @@ private static void mergeCurrentBranch(Project project,
     new MergeCurrentBranchFastForwardOnlyBackgroundable(project, gitRepository, mergeProps.getStayingBranch()).queue();
   }
 
+  @SuppressWarnings("regexp")
```
PawelLipski, line 30: Nit: can be applied just on the affected `val`, I suppose

PR #908, opened 2022-06-08
add annotations and stub class in astub

azdrojowa123:


qual/src/main/java/com/virtuslab/qual/gitmachete/frontend/resourcebundles/LooseHtml.java
```diff
@@ -0,0 +1,17 @@
...
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import org.checkerframework.common.subtyping.qual.Unqualified;
+import org.checkerframework.framework.qual.SubtypeOf;
+
+import com.virtuslab.qual.internal.SubtypingTop;
+
+
+@SubtypeOf(Unqualified.class)
```
PawelLipski, line 14: I probably haven't though of using `Unqualified` for the top-level type of all these annotations... if it works as expected for your new `@...Html` annotations, then considered removing `com.virtuslab.qual.internal.SubtypingTop`.

TBD if `com.virtuslab.qual.internal.SubtypingBottom` can be removed as well... but this might be harder

PawelLipski: Oh! judging by the javadoc for `org.checkerframework.common.subtyping.qual.Unqualified` and `org.checkerframework.common.subtyping.qual.Bottom`... our own `SubtypingTop` and `SubtypingBottom` should be retained & used everywhere :/

azdrojowa123: it doesn't work as it should with `Unqualified`

azdrojowa123: with our `SubtypingTop` when I test all this flow, check framework does not check if the given method returns exact e.g. `@...Html` but check for `SubtypingTop`.  I need to read more doc about it
Screenshot 2022-06-08 at 12 38 18

PawelLipski: Hmmm I think this is expected 🤔 as in, a plain "unsubtyped" String will indeed be seen as `@SubtypingTop String`

PawelLipski: ...anyway... lemme push a few changes onto a separate PR... I think they might be useful for you

azdrojowa123: a okey in this situation format method return plain String but method `newAnnotation` required `@NonHtml`

PawelLipski: Awww crap I see the problem :ups:

PawelLipski: Shiiiiiit that might be difficult to work around ☹️

PawelLipski: As in, when `format` is NOT involved, `SubtypingChecker` will work as expected... but how to retain the `@...Html` from `format`'s param into `format`'s return type? 😕

PawelLipski: Not Good ™️

Unless you find some simple solution here... I have one more idea (https://checkerframework.org/manual/#tainting-checker)... and if this doesn't work, I'd simply resort to documenting in CONTRIBUTING.md what we know about the format accepted by each method... and probably call it a day

azdrojowa123: I will try (before the refactor of the rest of the code)

PawelLipski:
Closing the PR, Tainting Checker is better suited here (#911)

PR #909, opened 2022-06-08
Patch Checker Framework classpath to accomodate for annotations from `:qual` in astubs

PawelLipski:


build.gradle
```diff
@@ -406,15 +417,10 @@ allprojects {
     }
 
     applySubtypingChecker = { ->
-      if (shouldRunAllCheckers) {
```
PawelLipski, line 409: Let's enable in local runs as well, so that one doesn't need to pass `CI=true` or `./gradlew -PrunAllCheckers` for tests

build.gradle
```diff
@@ -311,6 +311,17 @@ allprojects {
       compileOnly      group: 'org.checkerframework', name: 'checker-qual', version: checkerFrameworkVersion
       checkerFramework group: 'org.checkerframework', name: 'checker',      version: checkerFrameworkVersion
     }
+
+    if (project.name != 'qual') {
+      dependencies {
+        implementation project(':qual')
+        checkerFramework project(':qual')
+      }
+      checkerFramework {
```
PawelLipski, line 320: All subprojects depend on `qual`... slight overkill probably but it's modified so rarely that #yolo

frontend/file/build.gradle
```diff
@@ -9,6 +9,7 @@ dependencies {
 
 addIntellijToCompileClasspath(withGit4Idea: true)
 applyI18nFormatterChecker()
+applySubtypingChecker()
```
PawelLipski, line 12: Enabled in this module just for example... I'm getting errors in `MacheteAnnotator`:

```
              getString("string.GitMachete.MacheteAnnotator.cannot-find-local-branch-in-repo").format(processedBranchName))
                                                                                                     ^
  found   : @SubtypingTop String
  required: @NonHtml String
```

which is expected currently

azdrojowa123: okey so at the moment  all checks should throw exceptions because of format method

PawelLipski: Yeah... I've realized that only now, that's gonna be a tough problem to overcome. Unless an easy workaround can be found (unlikely) and Tainting Checker won't help (which I'm gonna check in a moment)... I'd just save your yesterday's findings somewhere in docs, and close the story as (mostly) `wontfix`

PawelLipski:
Closing the PR, Tainting Checker is better suited here (#911)

PR #910, opened 2022-06-08
Use Tainting Checker to prevent possibly-HTML strings from being passed as non-HTML params

PawelLipski:


frontend/base/src/main/java/com/virtuslab/gitmachete/frontend/resourcebundles/GitMacheteBundle.java
```diff
@@ -26,12 +29,17 @@ private GitMacheteBundle() {}
    *             but each parameter must be a non-null {@link String} and not just a nullable {@link Object}
    * @return the formatted string
    */
-  public static String format(@I18nFormatFor("#2") String format, String @MinLen(1)... args) {
+  public static @PolyTainted String format(@PolyTainted @I18nFormatFor("#2") String format, String @MinLen(1)... args) {
```
PawelLipski, line 32: `PolyTainted` in the sense that if param is `@Untainted` (surely non-HTML, in our usage), then the result will also be `@Untainted`; and vice versa, if param is `@Tainted` (maybe containing HTML, in our usage), then the result will also be `@Tainted`

frontend/base/src/main/java/com/virtuslab/gitmachete/frontend/resourcebundles/GitMacheteBundle.java
```diff
@@ -26,12 +29,17 @@ private GitMacheteBundle() {}
...
     return MessageFormat.format(format, (@Nullable Object[]) args);
   }
 
   @I18nMakeFormat
-  public static String getString(@PropertyKey(resourceBundle = BUNDLE) String key) {
+  public static @Tainted String getString(@PropertyKey(resourceBundle = BUNDLE) String key) {
```
PawelLipski, line 37: `@Tainted` is kinda the implicit default for everything in the entire codebase, here only added for clarity

PawelLipski:
Closing this PR, work to be continued on #911

PR #911, opened 2022-06-09
Use Tainting Checker to mark places that allow HTML and do not

azdrojowa123:


config/checker/com.intellij.astub
```diff
@@ -1,12 +1,18 @@
-
 import org.checkerframework.checker.guieffect.qual.AlwaysSafe;
 import org.checkerframework.checker.guieffect.qual.SafeEffect;
 import org.checkerframework.checker.guieffect.qual.UI;
 import org.checkerframework.checker.guieffect.qual.UIEffect;
 import org.checkerframework.checker.guieffect.qual.UIPackage;
 import org.checkerframework.checker.index.qual.NonNegative;
+import com.intellij.codeInspection.util.InspectionMessage;
```
PawelLipski, line 7: Nit: as the powers of Spotless autoformatter can't easily reach here... pls keep the imports in abc order, in a single section

config/checker/com.intellij.astub
```diff
@@ -1,12 +1,18 @@
...
 import org.checkerframework.checker.initialization.qual.UnknownInitialization;
 import org.checkerframework.checker.nullness.qual.Nullable;
+import org.jetbrains.annotations.NotNull;
```
PawelLipski, line 10: Imports from outside `org.checkerframework.checker.*` packages are almost surely not needed in astubs... lemme see the specific usage, but in general almost everything is ignored by Checker in astubs, other than Checker's own annotations

config/checker/com.intellij.astub
```diff
@@ -135,6 +148,10 @@ class NotificationAction {
 
 @UIPackage package com.intellij.openapi.actionSystem;
 
+class Presentation {
+    public void setDescription(@ActionDescription @Untainted String description);
```
PawelLipski, line 152: Ditto about `ActionDescription`

config/checker/com.intellij.astub
azdrojowa123: hmm but if I e.g. delete @Nullable annotation from GitUntrackedFilesHelper.notifyUntrackedFilesOverwrittenBy and description param I am getting an exception
Screenshot 2022-06-09 at 17 36 25

config/checker/com.intellij.astub
```diff
@@ -225,6 +242,12 @@ interface Scheme {
 
 package com.intellij.openapi.progress;
 
+import static com.intellij.openapi.util.NlsContexts.ProgressText;
+
+interface ProgressIndicator {
+    void setText(@ProgressText @Untainted String text);
+}
+
```
PawelLipski, line 250: Actually... since you've already done the research on what params accept HTML text... I'd consider adding explicit `@Tainted` annotations for them.

CMIIW, but it's a shame from **soundess** PoV that the default for UNannotated fields is to assume `@Tainted`... although OTOH it'd be somewhat **impractical** if `@Untainted` was assumed for everything by default :/

config/checker/com.intellij.astub
azdrojowa123: But when I add @Nullable to stub (as it is in original declaration) it works properly

config/checker/com.intellij.astub
```diff
@@ -119,9 +125,16 @@ class DataManager {
 @UIPackage package com.intellij.ide.util.treeView;
 @UIPackage package com.intellij.ide.util.treeView.smartTree;
 @UIPackage package com.intellij.ide.wizard;
-@UIPackage package com.intellij.navigation;
+
+package com.intellij.lang.annotation;
+
+interface AnnotationHolder {
+  default AnnotationBuilder newAnnotation(@NotNull HighlightSeverity severity, @NotNull @InspectionMessage @Untainted String message);
```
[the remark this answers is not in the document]
azdrojowa123: > Nit: theoretically `org.jetbrains.annotations.NotNull` is NOT ignored by NullnessChecker (see https://checkerframework.org/manual/#nullness-related-work)... but still, is the explicit non-nullness declaration even needed on these params? as in, if you remove `@NotNull` from them, will Checker allow you to pass nulls for these params?

No Checker will not allow to pass nulls so I can delete it

config/checker/com.intellij.astub
PawelLipski: If you agree with this idea, I'd document the following distinction somewhere... maybe even in CONTRIBUTING.md to keep everything in a single place. 

And definitely, it needs to be documented that `@(Un)Tainted` annotations are used specifically to mark HTML usage... as they're pretty versatile, can be used for pretty much anything, whatever developers agree upon in the project

config/checker/git4idea.astub
```diff
@@ -0,0 +1,34 @@
+import org.checkerframework.checker.tainting.qual.Untainted;
+import org.jetbrains.annotations.Nls;
+import com.intellij.openapi.util.NlsContexts.DialogTitle;
```
PawelLipski, line 3: Ditto, remove unused annotations, TBD if `NotNull` is redundant or not; `Nullable` definitely is NOT redundant

PawelLipski: Plus add `@Tainted` for clarity where applicable

frontend/actions/build.gradle
```diff
@@ -11,6 +11,11 @@ dependencies {
   implementation project(':frontend:base')
 }
 
+checkerFramework {
+  checkers += 'org.checkerframework.checker.tainting.TaintingChecker'
```
PawelLipski, line 15: CMIIW, but since we're basically using `TaintingChecker` wherever `I18NFormatterChecker` is used, I'd just rename `applyI18nFormatterChecker` to `applyI18nFormatterAndTaintingCheckers` (and include `TaintingChecker` in that function, ofc)

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/backgroundables/MergeCurrentBranchFastForwardOnlyBackgroundable.java
```diff
@@ -31,8 +33,8 @@ public MergeCurrentBranchFastForwardOnlyBackgroundable(
   }
 
   @Override
-  protected @I18nFormat({}) String getOperationName() {
-    return getString("action.GitMachete.MergeCurrentBranchFastForwardOnlyBackgroundable.operation-name");
+  protected @I18nFormat({}) @Untainted String getOperationName() {
+    return getNonHtmlString("action.GitMachete.MergeCurrentBranchFastForwardOnlyBackgroundable.operation-name");
```
PawelLipski, line 37: To be considered, not insisting:

```suggestion
    return getString("action.GitMachete.MergeCurrentBranchFastForwardOnlyBackgroundable.operation-name").nonHtml();
```

Plus also, require that each property that contains HTML, has a `.HTML` "extension", e.g.:

```properties
action.GitMachete.OpenMacheteTabAction.notification.title.could-not-open-tab.HTML=<html><b>Could not open Git Machete tab</b></html>
```

This way one could easily spot sth like:

```java
getString("action.GitMachete.OpenMacheteTabAction.notification.title.could-not-open-tab.HTML").nonHtml() // WHOOPS
```

PawelLipski: (`nonHtml` could be ofc an extension method declared in `GitMacheteBundle`, similarly to `format`)

PawelLipski: (and `getNonHtml` could be removed completely then)

config/checker/com.intellij.astub
azdrojowa123: > So yeah, what I'm effectively asking for is to distingiush b/w the cases of:

So you think about doing a similar job for places that can be @Tainted? to make it clear that HTML can be placed there

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BasePullBranchAction.java
```diff
@@ -93,11 +94,11 @@ public void actionPerformed(AnActionEvent anActionEvent) {
...
       boolean isUpToDate = FetchUpToDateTimeoutStatus.isUpToDate(gitRepository);
       String fetchNotificationPrefix = isUpToDate
-          ? getString("action.GitMachete.BasePullBranchAction.notification.title.no-fetch-perform")
+          ? getNonHtmlString("action.GitMachete.BasePullBranchAction.notification.title.no-fetch-perform")
               .format(FETCH_ALL_UP_TO_DATE_TIMEOUT_AS_STRING)
-          : getString("action.GitMachete.BasePullBranchAction.notification.title.fetch-perform");
+          : getNonHtmlString("action.GitMachete.BasePullBranchAction.notification.title.fetch-perform");
       Runnable fastForwardRunnable = () -> FastForwardMerge.perform(project, gitRepository, mergeProps,
-          fetchNotificationPrefix);
+          fetchNotificationPrefix + "<br/>");
```
PawelLipski, line 101: Huh okay so `fetchNotificationPrefix` does ultimately accept HTML... or not? 🤔 Since it's marked as `@Untainted`... but here `<br/>` is smuggled

config/checker/com.intellij.astub
PawelLipski: Poor \@nullable, they must be mentioned 1000s times a day 😅 tip: you can use `\@nullable`

PawelLipski: Yeah, so `@Nullable` annotations are to be left, `@NotNull` to be removed as redundant

config/checker/com.intellij.astub
PawelLipski: Yes, indeed &mdash; just based on the research you've already put into a Google Doc, no need to do extra research as for now

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/common/FastForwardMerge.java
```diff
@@ -38,23 +41,24 @@ private static void mergeNonCurrentBranch(Project project,
     val successFFMergeNotification = fetchNotificationPrefix
         + getString("action.GitMachete.BaseFastForwardMergeBranchToParentAction.notification.title.ff-success")
             .format(stayingName, movingName);
-    @SuppressWarnings("regexp") val failFFMergeNotification = fetchNotificationPrefix.replaceAll("<br/>", "\n")
-        + getString("action.GitMachete.BaseFastForwardMergeBranchToParentAction.notification.title.ff-fail").format(stayingName,
+    String failFFMergeNotification = changeNewLineSign(fetchNotificationPrefix)
```
PawelLipski, line 44: Awww I see now.... actually, how about passing the `<br/>` suffix (or an empty string) in a separate param? could even be made a boolean param (`insertNewline` or sth)

Or even better, is this suffix **always** appended to `fetchNotificationPrefix`?

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/common/SlideInOptions.java
```diff
@@ -0,0 +1,23 @@
+package com.virtuslab.gitmachete.frontend.actions.common;
+
+import org.checkerframework.checker.tainting.qual.Untainted;
+
+public class SlideInOptions {
```
PawelLipski, line 5: Okay so now it's compatible with Kotlin? out of curiosity, what were the problems when using Lombok?

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BasePullBranchAction.java
azdrojowa123: here is strange situation. fetchNotificationPrefix can have html tags when it goes to successNotification  but when it goes to failNotification it cannot (fail notification is different in two places and adding html is allowed in one place and not allowed in the other). So I decided to make rule that fetchNotificationPrefix cannot have html tags inside

frontend/base/src/main/java/com/virtuslab/gitmachete/frontend/resourcebundles/GitMacheteBundle.java
```diff
@@ -26,12 +28,21 @@ private GitMacheteBundle() {}
...
   public static String getString(@PropertyKey(resourceBundle = BUNDLE) String key) {
     return instance.getString(key);
   }
+
+  @I18nMakeFormat
+  public static @Untainted String getNonHtmlString(@PropertyKey(resourceBundle = BUNDLE) String key) {
+    return instance.getString(key);
+  }
+
+  public static @Untainted @SuppressWarnings("regexp") String changeNewLineSign(@Untainted String text) {
```
PawelLipski, line 45: If this method stays (and I'd strongly recommend to remove it and pass the newline by other means), pls rename to
```suggestion
  public static @Untainted @SuppressWarnings("regexp") String replaceHtmlLineBreakWithNewLineCharacter(@Untainted String text) {
```
for clarity

PawelLipski: Btw yeah... `text` is `@Untainted` but might contain a `<br/>`... something is _not yes_ here 😅

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/common/FastForwardMerge.java
azdrojowa123: nope because in one place fetchNotificationPrefix is "" and we do not want to add` <br/>`in that situation

frontend/file/build.gradle
```diff
@@ -7,10 +7,13 @@ dependencies {
   implementation project(':frontend:icons')
 }
 
+checkerFramework {
+  checkers += 'org.checkerframework.checker.tainting.TaintingChecker'
```
PawelLipski, line 11: Ditto, move to `applyI18nFormatterChecker`

PawelLipski (commented):
Good job Ola... and a visible proof of how terrible job JetBrains did 😅

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/common/FastForwardMerge.java
PawelLipski: Okay, so I'd suggest a separate (boolean?) param then

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BasePullBranchAction.java
PawelLipski: > So I decided to make rule that fetchNotificationPrefix cannot have html tags inside

Okay, makes sense indeed... but here you still pass a `<br>` (i.e. a **tainted** string) in `fetchNotificationPrefix` param of `FastForwardMerge.perform`, right? So against this rule ☹️

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/common/SlideInOptions.java
PawelLipski: Yeah, TBH I still don't get the order of how Kotlin and Java compilers interact... looks a bit like a chicken-and-egg problem to me:

* if you compile Java first (in the given Gradle module), then how can you access Kotlin classes from Java code?
* if you compile Kotlin first (in the given Gradle module), then how can you access Java classes from Kotlin code?

While we're at this... lemme summon @Kordyjan as the Kotlin expert! can you shed some light?

frontend/base/src/main/java/com/virtuslab/gitmachete/frontend/resourcebundles/GitMacheteBundle.java
azdrojowa123: I'm taking this prefix to work on, because it actually looks bad (all because of the unfortunate `fetchNotificationIfFailed`). I will try to come up with some way to systematize it

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/common/FastForwardMerge.java
```diff
@@ -32,16 +31,18 @@ private static void mergeCurrentBranch(Project project,
...
       GitRepository gitRepository,
-      MergeProps mergeProps, @Untainted String fetchNotificationPrefix) {
+      MergeProps mergeProps, @Untainted String fetchNotificationPrefix, boolean insertNewlineAfterPrefix) {
     val stayingFullName = mergeProps.getStayingBranch().getFullName();
     val movingFullName = mergeProps.getMovingBranch().getFullName();
     val refspecFromChildToParent = createRefspec(stayingFullName, movingFullName, /* allowNonFastForward */ false);
     val stayingName = mergeProps.getStayingBranch().getName();
     val movingName = mergeProps.getMovingBranch().getName();
-    val successFFMergeNotification = fetchNotificationPrefix
+    val newLineSymbolHtml = insertNewlineAfterPrefix ? "</br>" : "";
```
PawelLipski, line 40:
```suggestion
    val newLineSymbolHtml = insertNewlineAfterPrefix ? "<br/>" : "";
```

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/common/FastForwardMerge.java
```diff
@@ -58,11 +59,12 @@ private static void mergeNonCurrentBranch(Project project,
 
   public static void perform(Project project,
       GitRepository gitRepository,
-      MergeProps mergeProps, @Untainted String fetchNotificationPrefix) {
+      MergeProps mergeProps, @Untainted String fetchNotificationPrefix, boolean insertNewlineAfterPrefix) {
     val stayingName = mergeProps.getStayingBranch().getName();
     val movingName = mergeProps.getMovingBranch().getName();
     val currentBranchName = Option.of(gitRepository.getCurrentBranch()).map(b -> b.getName()).getOrNull();
-    val failFFMergeNotification = fetchNotificationPrefix
+    val newLineSymbolHtml = insertNewlineAfterPrefix ? "</br>" : "";
```
PawelLipski, line 66:
```suggestion
    val newLineSymbolHtml = insertNewlineAfterPrefix ? "<br/>" : "";
```

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseFastForwardMergeBranchToParentAction.java
```diff
@@ -70,6 +72,8 @@ public void actionPerformed(AnActionEvent anActionEvent) {
     val mergeProps = new MergeProps(
         /* movingBranchName */ nonRootStayingBranch.getParent(),
         /* stayingBranchName */ nonRootStayingBranch);
-    FastForwardMerge.perform(project, gitRepository, mergeProps, /* fetchNotificationPrefix */ "");
+    FastForwardMerge.perform(project, gitRepository, mergeProps, /* fetchNotificationPrefix */ "", /*
```
PawelLipski, line 75: Whoops

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/backgroundables/MergeCurrentBranchFastForwardOnlyBackgroundable.java
azdrojowa123: that sounds good to me but in this case together with `format` it would look like `getString("...").nonHtml().format("...")`. Would it be clear enough ?

PawelLipski: > where html may be written but we do not use this possibility

In such places, I would NOT append `.HTML` &mdash; only where HTML is actually used.
I'd separate **properties** itself from the **methods/parameters** to which they're passed (esp. that any given non-HTML property can be passed to both `@Untainted` and `@Tainted` params).
In other words, properties themselves should **not** care that to what params they're passed &mdash; only whether they contain HTML or not

azdrojowa123: > (`nonHtml` could be ofc an extension method declared in `GitMacheteBundle`, similarly to `format`)

I have problem with this implementation because `format ` expect proper `i18nFormat` (which is at this moment generating from `getString` because we have `@ I18nMakeFormat` there). I am trying to cheat checkerframework but with no effect so far

PawelLipski: Ouch, I forgot about this limitation 🤦🏻

Indeed, I don't think this is possible to circumvent (as in, I don't think you'll be able to mark `nonHtml`'s result as a format string)... so sticking to `getNonHtmlString` might be a better idea

PawelLipski: But `.HTML` suffix for property keys still makes sense IMO

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BasePullBranchAction.java
PawelLipski: (pls Resolve the outdated/resolved comments so that we can see what's really still left!)

frontend/file/build.gradle
```diff
@@ -8,8 +8,8 @@ dependencies {
 }
 
 addIntellijToCompileClasspath(withGit4Idea: true)
-applyI18nFormatterChecker()
-
+applyI18nFormatterAndTaintingCheckers()
+applySubtypingChecker()
```
PawelLipski, line 12: Probably not needed!

PawelLipski (commented):
So far so good, looks that only:
1. `@Tainted` in astubs
2. some JUnit tests for property format (`<html></html>`, `.HTML`)
3. docs in CONTRIBUTING.md

are needed... and we can call it a day

config/checker/com.intellij.astub
azdrojowa123: hmm I am stuck into a problem with kotlin files - it looks to me like checker annotations from java files are not applicable to kotlin classes ( e.g. giving too many arguments to the format method does not cause an error). Do you think that some configuration may change this behavior ?

PawelLipski: Yeah, Checker can't do anything about Kotlin code in general :/// we have to live with some potential unsoundness there :/

PawelLipski: For that reason we have e.g. sth like https://github.com/VirtusLab/git-machete-intellij-plugin/blob/master/scripts/prohibit-kotlin-undefined-property-keys ... as a low-quality _prosthetic_ for what Checker gives in Java files

PawelLipski: Huh remind me, both `label()` and `checkbox()` **forbid** the usage of HTML right?

config/checker/com.intellij.astub
```diff
@@ -137,6 +138,10 @@ class NotificationAction {
   static NotificationAction createSimple(Supplier<String> dynamicText, @UI Runnable performAction);
 }
 
+class NotificationGroup {
+      Notification createNotification(@Tainted String title, @Tainted String content, NotificationType type, @Nullable NotificationListener listener);
```
PawelLipski, line 142: Nit: two spaces for indent

frontend/ui/impl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/RediscoverSuggester.java
```diff
@@ -54,7 +54,7 @@ public void perform() {
 
   @UIEffect
   private void queueSuggestion(Path macheteFilePath) {
-    val yesNo = MessageDialogBuilder.YesNo.yesNo(
+    val yesNo = MessageDialogBuilder.yesNo(
```
PawelLipski, line 57: Are the two the same method? 😯

PawelLipski: Or is there a specific reason for the change

config/checker/com.intellij.astub
azdrojowa123: no `label()`, `checkbox()`, `button()` and  `com.intellij.ui.layout.ValidationInfoBuilder#error `allow html usage, but they are only used in kotlin files so I skipped adding them to stubs (I marked them in doc to write a checking script for them). Is it a good idea ?

PawelLipski: Yeah, makes sense indeed

config/checker/com.intellij.astub
azdrojowa123: two spaces between package name and class or two spaces between each class ? :>

PawelLipski: Spaces! not newlines

2 spaces as one level indent

PawelLipski:
Btw, pls rename PR to sth more meaningful... in case somebody needs to revisit it in a few months/years 😅

frontend/base/src/test/java/com/virtuslab/gitmachete/resourcebundles/junit/GitMacheteBundlePropertiesTestSuite.java
```diff
@@ -0,0 +1,35 @@
...
+  @Test
+  public void htmlProperties_should_have_correct_syntax() throws IOException {
+
+    Properties property = new Properties();
+    ClassLoader loader = Thread.currentThread().getContextClassLoader();
+    property.load(loader.getResourceAsStream(macheteBundleProperties));
+
+    property.entrySet().stream()
+        .filter(prop -> ((String) prop.getValue()).contains("<"))
+        .forEach(t -> assertTrue(isCorrectHtmlPropertyFormat((String) t.getKey(), (String) t.getValue())));
```
PawelLipski, line 23: Actually! the error messages from this tests aren't going to be particularly helpful :/ I'd rather create a separate assertion for each of the 3 conditions from line 31, each with a developer-friendly error message

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/common/SlideInOptions.java
PawelLipski: Okay, looks that Kotlin compiler first parses Java sources without actually compiling them, just to see what's out there in Java code... and then it's easy, coz Java compiler already sees compiled Kotlin classes

frontend/base/src/test/java/com/virtuslab/gitmachete/resourcebundles/junit/GitMacheteBundlePropertiesTestSuite.java
azdrojowa123: stick to the principle of one assertion per test Should I create three tests ? Or break this rule in this case ?

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/common/SlideInOptions.java
azdrojowa123: from what I found out the only solution is to create compilation order - first java, and later kotlin. But in this case we can't use kotlin code from java

frontend/base/src/test/java/com/virtuslab/gitmachete/resourcebundles/junit/GitMacheteBundlePropertiesTestSuite.java
PawelLipski: I wouldn't stick to that principle too religiously... 3 closely related assertions like these in a single test are okay!

azdrojowa123:
What about script for checking `label(),` `checkbox()`, `button()`methods ? create separate issue for this or do it now ?

PR #918, opened 2022-06-13
Set up autoformatter (scalafmt via spotless) for Scala code

PawelLipski:
Setting this up with Gradle is smoother than with Scala's dedicated build tool, `sbt` :trollface:

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -197,6 +197,7 @@ dependencyUpdates {
 
 import org.gradle.api.tasks.testing.logging.TestExceptionFormat
 import org.gradle.api.tasks.testing.logging.TestLogEvent
+import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 
 allprojects {
   repositories {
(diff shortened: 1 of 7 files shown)


build.gradle
```diff
@@ -441,6 +450,14 @@ allprojects {
       }
     }
 
+    applyScalaConfig = { ->
+      spotless {
```
PawelLipski, line 454: Spotless is Gradle plugin which provides support for many language-specific formatting tools... like `scalafmt` for Scala (which is itself completely independent from Gradle)

scalafmt.conf
```diff
@@ -0,0 +1,2 @@
+runner.dialect = scala213
```
PawelLipski, line 1: As in, Scala 2.13 (the last major version of Scala 2.x; there's not gonna ever be any Scala 2.14, just Scala 3.x)

mkondratek (approved):

Issue #843, opened 2022-05-17
Merge current/non-current branch UI test cases


mkondratek:
From https://github.com/VirtusLab/git-machete-intellij-plugin/pull/823#discussion_r872976921.

That should save us some time, as setting up a new IDE instance & sample git repo for each test **case** (i.e. each test **method**) constitutes a major overhead.

Issue #427, opened 2020-07-08
Display only single reset/pull/push/slide-out button for the current branch, depending on its state


PawelLipski:
A single button could change between reset/pull/push/slide-out/nothing, depending on the state of the branch.

TBD: this would break our rule that we don't hide actions from the toolbar, just disable them :thinking:

PawelLipski:
EDIT: added slide-out which should be suggested for merged branches

mkondratek:
Remove unused action implementations ~ https://github.com/VirtusLab/git-machete-intellij-plugin/pull/450#discussion_r462966214

PawelLipski:
Also, README.md (expanded on #452) needs to be updated

PR #926, opened 2022-06-17
Use indirect methods (toolbar/context-menu) in ui tests

mkondratek:


PawelLipski (approved):

testCommon/src/test/resources/setup-with-single-remote.sh
```diff
@@ -22,6 +22,11 @@ cd machete-sandbox
     # Let's clean up the upstream so that we can't rely on `git config` for finding out the remote branch;
     # we'll need to match the local branch to its remote tracking branch by name.
     git branch --unset-upstream
+  create_branch update-icons
```
mkondratek, line 25: additional branch - to perform current/noncurrent branch **pull** (preparation for mergin ui test cases #843)

testCommon/src/test/resources/setup-with-single-remote.sh
```diff
@@ -44,8 +49,11 @@ cd machete-sandbox
     push
   create_branch hotfix/add-trigger
     commit HOTFIX Add the trigger
+    commit HOTFIX Add the trigger - fixes
     push
-    git commit --amend -m 'HOTFIX Add the trigger (amended)'
```
mkondratek, line 48: instead of diverged and newer let's have diverged and older case - that will enable reset to remote action in the toolbar. 

the action however is available from the context-menu - we are a bit inconsistent b/c of https://github.com/VirtusLab/git-machete-intellij-plugin/issues/427. 

in general I am calling current branch referring actions from toolbar and non-current from context-menu to cover both cases

PR #936, opened 2022-06-22
Add new env for saving screenshots in specific directory

azdrojowa123:
Need to delete debug code  but changes in ideprobe.conf, build.gradle (line 719) and config.yaml (line 117) were taken because of https://github.com/VirtusLab/ide-probe/pull/197

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -98,11 +98,12 @@ jobs:
           name: Collect artifacts when an exception occurs
           # language=sh
           command: |
-            mkdir -p /exception-artifacts
-            for f in $(ls -d /tmp/ide-probe/screenshots/*/) ; do cd  $f/artifacts ; zip -r "/exception-artifacts/$(basename ${f}).zip" . ; done
+            for f in $(ls -d ~/artifacts/*/) ; do cd  $f ; zip -r "../$(basename ${f}).zip" . ; done
+            mkdir -p ~/artifacts/zip
+            find ~/artifacts -name '*.zip' | xargs cp --target-directory ~/artifacts/zip
           when: on_fail
       - store_artifacts:
-          path: /exception-artifacts
+          path: ~/artifacts/zip
           destination: .
 
       # Unfortunately, wildcards for test result paths aren't supported by CircleCI yet.
--- a/build.gradle
+++ b/build.gradle
@@ -179,6 +179,7 @@ import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 
 allprojects {
   repositories {
+    mavenLocal()
     mavenCentral()
   }
 
@@ -681,9 +682,7 @@ uiTestTargets.each { version ->
 
     if (project.hasProperty('headless')) {
       environment 'IDEPROBE_DISPLAY', 'xvfb'
-      // TODO (VirtusLab/ide-probe#195): remove the need to set these env variables
-      environment 'IDEPROBE_TEST_SUITE', 'uiTest' + version
-      environment 'IDEPROBE_TEST_CASE', 'artifacts'
+      environment 'IDEPROBE_SCREENSHOTS_DIR', System.getProperty("user.home") + '/artifacts/uiTest' + version + '/screenshots'
     }
 
     testLogging {
--- a/src/uiTest/scala/com/virtuslab/gitmachete/uitest/UITestSuite.scala
+++ b/src/uiTest/scala/com/virtuslab/gitmachete/uitest/UITestSuite.scala
@@ -91,7 +91,6 @@ class UITestSuite extends BaseGitRepositoryBackedIntegrationTestSuite(SETUP_WITH
     // Let's slide out a root branch now
     project.contextMenu.openContextMenu("develop")
     project.contextMenu.slideOut()
-    project.acceptBranchDeletionOnSlideOut()
     branchAndCommitRowsCount = project.refreshModelAndGetRowCount()
     // 5 branch rows (`develop` is no longer there) + 7 commit rows
     // (1 commit of `allow-ownership-link` and 3 commits of `call-ws` are all gone)
@@ -100,7 +99,6 @@ class UITestSuite extends BaseGitRepositoryBackedIntegrationTestSuite(SETUP_WITH
     project.checkoutBranch("master")
     project.contextMenu.openContextMenu("call-ws")
     project.contextMenu.slideOut()
-    project.rejectBranchDeletionOnSlideOut()
     branchAndCommitRowsCount = project.refreshModelAndGetRowCount()
     // 4 branch rows (`call-ws` is also no longer there) + 8 commit rows
     Assert.assertEquals(12, branchAndCommitRowsCount)
--- a/src/uiTest/scala/com/virtuslab/gitmachete/uitest/UITestSuite.scala
+++ b/src/uiTest/scala/com/virtuslab/gitmachete/uitest/UITestSuite.scala
@@ -3,7 +3,8 @@ package com.virtuslab.gitmachete.uitest
 import com.virtuslab.gitmachete.testcommon.BaseGitRepositoryBackedIntegrationTestSuite
 import com.virtuslab.gitmachete.testcommon.SetupScripts.SETUP_WITH_SINGLE_REMOTE
 import org.junit._
-import org.junit.runner.RunWith
+import org.junit.rules.TestWatcher
+import org.junit.runner.{Description, RunWith}
 import org.junit.runners.JUnit4
 import org.virtuslab.ideprobe.Extensions._
 import org.virtuslab.ideprobe.ProbeDriver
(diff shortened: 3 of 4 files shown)


src/uiTest/scala/com/virtuslab/gitmachete/uitest/UITestSuite.scala
```diff
@@ -215,7 +219,7 @@ class UITestSuite extends BaseGitRepositoryBackedIntegrationTestSuite(SETUP_WITH
     val pid = intelliJ.probe.pid()
     intelliJ.probe.screenshot("exception")
     val threadStackTrace: String = Process("jstack " + pid) !!
-    val artifactDirectory = "/tmp/ide-probe/screenshots/uiTest" + intelliJVersion.build + "/artifacts"
+    val artifactDirectory = "/build/UIException/uiTest" + intelliJVersion.build + "/artifacts"
```
PawelLipski, line 222: Actually... I how about using Gradle's `build/` folder in the current directory?
```suggestion
    val artifactDirectory = "build/screenshots/uiTest" + intelliJVersion.build + "/artifacts"
```

intellijVersions.properties
```diff
@@ -1,5 +1,5 @@
 #Wed May 18 11:15:19 CEST 2022
-latestMinorsOfOldSupportedMajors=2020.3.4,2021.1.3,2021.2.4,2021.3.3
```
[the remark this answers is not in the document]
azdrojowa123: need to change, mistake

src/uiTest/scala/com/virtuslab/gitmachete/uitest/UITestSuite.scala
azdrojowa123: but this` /build/UIException/uiTest` does not refer to build folder in root directory ? (because screenshots are saving there - I pass `/build `to `MY_CI_SCREENSHOTS_DIR`)

PawelLipski: Hmmm with a leading slash? I'd expect it to refer to an absolute path 🤔

PawelLipski: Or even better... pass the path to `build/` directory in the project in `MY_CI_SCREENSHOTS_DIR` (plus `screenshots/` folder)!

See `systemProperty 'ui-test.plugin.path'` for a similar trick, for passing Gradle paths to the otherwise Gradle-agnostic tests

build.gradle
```diff
@@ -37,7 +37,8 @@ ext {
   checkerFrameworkVersion = '3.22.1'
   checkstyleToolVersion = '8.33'
   commonsIOVersion = '2.11.0'
-  ideProbeVersion = '0.31.0'
+  ideProbeVersion = '0.31.0+9-882007b8-SNAPSHOT'
```
PawelLipski, line 40: Btw! sorry for asking yet another time... if you stick for `0.31.0` for this PR, will **anything** be broken? CMIIW but `probe.paths.screenshots` should work as expected?

build.gradle
```diff
@@ -714,6 +716,7 @@ uiTestTargets.each { version ->
     if (project.hasProperty('headless')) {
       environment 'IDEPROBE_DISPLAY', 'xvfb'
       // TODO (VirtusLab/ide-probe#195): remove the need to set these env variables
+      environment 'MY_CI_SCREENSHOTS_DIR', '/build/UIException'
       environment 'IDEPROBE_TEST_SUITE', 'uiTest' + version
```
PawelLipski, line 720: Also! even with https://github.com/VirtusLab/ide-probe/pull/197 i.e. `0.31.0+9-882007b8-SNAPSHOT` version being used here... are these env vars still needed? 🤔 again, I'm lost a bit 😅

PawelLipski: Oh okay, they could be **unset** but then screenshots for all versions would land into a single directory, right?

azdrojowa123: There are not needed now, but without them screenshots will be saved directly into build/UIException (but we want to keep versions so set `IDEPROBE_TEST_SUITE`). Working with  `0.31.0+9-882007b8-SNAPSHOT` allow us to skip` IDEPROBE_TEST_CASE` and and store directly into `MY_CI_SCREENSHOTS_DIR/IDEPROBE_TEST_SUITE` (but I leave this because we probably still working on 0.31.0)

build.gradle
azdrojowa123: yes

azdrojowa123: https://app.circleci.com/pipelines/github/VirtusLab/git-machete-intellij-plugin/5265/workflows/81de0d41-2a81-48a9-95d8-8ae1fd0b6d83/jobs/5541 
here is failed CI (with intention) run on `0.31.0+9-882007b8-SNAPSHOT`. So maybe now I remove all debug code from here and test with `0.31.0` ? or you want to test something more with `0.31.0+9-882007b8-SNAPSHOT`  ?

PawelLipski: Up to you, we can either merge this PR with ide-probe v0.31.0, or only merge it once we release v0.32.0, then including all v0.32.0-specific fixes (like skipping `IDEPROBE_TEST_CASE`)

.circleci/config.yml
```diff
@@ -78,43 +90,42 @@ jobs:
...
+      - run:
+          name: Run UI tests against earlier supported major IntelliJ versions
+          command: ./gradlew -Pheadless -Pagainst=latestMinorsOfOldSupportedMajors uiTest
       - run:
           name: Collect artifacts when an exception occurs
           # language=sh
           command: |
             mkdir -p /exception-artifacts
-            for f in $(ls -d /tmp/ide-probe/screenshots/*/) ; do cd  $f/artifacts ; zip -r "/exception-artifacts/$(basename ${f}).zip" . ; done
+            for f in $(ls -d build/UIException/*/) ; do cd  $f/artifacts ; zip -r "/exception-artifacts/$(basename ${f}).zip" . ; done
```
PawelLipski, line 128: Actually... why `UIException`? 🤔 this name doesn't seem to appear anywhere else in the project

azdrojowa123: I created UIException folder in` UITestSuite#saveThreadDumpAndScreenshot`

build.gradle
```diff
@@ -714,6 +717,7 @@ uiTestTargets.each { version ->
     if (project.hasProperty('headless')) {
       environment 'IDEPROBE_DISPLAY', 'xvfb'
       // TODO (VirtusLab/ide-probe#195): remove the need to set these env variables
+      environment 'MY_CI_SCREENSHOTS_DIR', project.projectDir.toString()+'/build/UIException'
```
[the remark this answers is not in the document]
azdrojowa123: but instead of screenshots also thread dump (txt file) will be available in this folder. This is not a problem for name `/screenshots` ?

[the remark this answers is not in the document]
azdrojowa123: yess (but it will be only one thread-dump txt file)

build.gradle
azdrojowa123: I think that wait until ide-probe release is better option (now screenshots should work because https://github.com/VirtusLab/git-machete-intellij-plugin/pull/922 so functionality is satisfied)

build.gradle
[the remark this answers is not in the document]
azdrojowa123: I have some problem connected to saving uiTest screenshots and thread-dump inside `build` directory. This folder is cleaning each time uiTest run with specific intelliJ version so under build directory only screenshots and thread-dump from last version is saved :/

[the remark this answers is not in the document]
azdrojowa123: I have some problem connected to saving uiTest screenshots and thread-dump inside build directory. This folder is cleaning each time uiTest run with specific intelliJ version so under build directory only screenshots and thread-dump from last version is saved :/ I can override clean task somehow, but it's hard for me to say whether it is a good solution or not better to save directly to root directory (into some folders)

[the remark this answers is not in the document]
PawelLipski: > This folder is cleaning each time uiTest run with specific intelliJ version so under build directory

Huh okay indeed... so up to you, either `/tmp` or some folder directly under project directory. 

Just don't use custom top-level directories under filesystem root (`/build` etc.) coz that's heavily **un**conventional... and on local machines users will usually **not** have permissions to create such folders.

build.gradle
```diff
@@ -714,8 +714,8 @@ uiTestTargets.each { version ->
     if (project.hasProperty('headless')) {
       environment 'IDEPROBE_DISPLAY', 'xvfb'
       // TODO (VirtusLab/ide-probe#195): remove the need to set these env variables
-      environment 'IDEPROBE_TEST_SUITE', 'uiTest' + version
-      environment 'IDEPROBE_TEST_CASE', 'artifacts'
+      environment 'MY_CI_SCREENSHOTS_DIR', "/artifacts"
+      environment 'IDEPROBE_TEST_SUITE', 'uiTest' + version + '/screenshots'
```
PawelLipski, line 718: Okay... so in the spirit of https://github.com/VirtusLab/ide-probe/pull/197#discussion_r904957348:

> The only official way to set the env var is by configuring probe screenshot path in the fixture, either through the .conf file or in code.

Can we get rid of `IDEPROBE_TEST_SUITE` here and just pass the path solely via `MY_CI_SCREENSHOTS_DIR`?

PawelLipski, line 718: I.e. sth like
```suggestion
      environment 'MY_CI_SCREENSHOTS_DIR', "/artifacts/uiTest" + version + "/screenshots"
```

azdrojowa123: yes :)

build.gradle
```diff
@@ -714,8 +715,7 @@ uiTestTargets.each { version ->
     if (project.hasProperty('headless')) {
       environment 'IDEPROBE_DISPLAY', 'xvfb'
       // TODO (VirtusLab/ide-probe#195): remove the need to set these env variables
-      environment 'IDEPROBE_TEST_SUITE', 'uiTest' + version
-      environment 'IDEPROBE_TEST_CASE', 'artifacts'
+      environment 'MY_CI_SCREENSHOTS_DIR', '/artifacts/uiTest' + version + '/screenshots'
```
PawelLipski, line 718: Actually... inside your local machine (MacOS/VM), are you able to store/retrieve these screenshots? My suspicion is that you won't be able to create `/artifacts` directory in the first place (and in CI it's only possible coz we run as root)

azdrojowa123: are you able to run uiTest task without sudo ? I'm still getting IOException (I gave all permissions to folder in which execute this task - probably need to change permissions in other places too)

PawelLipski: I wouldn't rely on developers creating and `chmod`ing their `/artifacts` directory locally... just pick a directory that doesn't require root to create!

PawelLipski: Much better!

build.gradle
```diff
@@ -714,8 +714,7 @@ uiTestTargets.each { version ->
     if (project.hasProperty('headless')) {
       environment 'IDEPROBE_DISPLAY', 'xvfb'
       // TODO (VirtusLab/ide-probe#195): remove the need to set these env variables
-      environment 'IDEPROBE_TEST_SUITE', 'uiTest' + version
-      environment 'IDEPROBE_TEST_CASE', 'artifacts'
+      environment 'MY_CI_SCREENSHOTS_DIR', System.getProperty("user.home") + '/artifacts/uiTest' + version + '/screenshots'
```
PawelLipski, line 717: Hmmm this `MY_` in env var name doesn't sound serious... and `CI_` is also inadequate, as this env var applies on local machines as well 🤔

.circleci/config.yml
```diff
@@ -113,13 +113,14 @@ jobs:
           name: Collect artifacts when an exception occurs
           # language=sh
           command: |
-            mkdir -p /exception-artifacts
-            for f in $(ls -d /tmp/ide-probe/screenshots/*/) ; do cd  $f/artifacts ; zip -r "/exception-artifacts/$(basename ${f}).zip" . ; done
+            for f in $(ls -d ~/artifacts/*/) ; do cd  $f ; zip -r "../$(basename ${f}).zip" . ; done
+            mkdir -p ~/artifacts/zip
+            find ~/artifacts -name '*.zip' | xargs cp -t ~/artifacts/zip
```
PawelLipski, line 118: Nit: actually, `-t` is pretty obscure/uncommon... pls use the long option here for clarity
```suggestion
            find ~/artifacts -name '*.zip' | xargs cp --target-directory ~/artifacts/zip
```

src/uiTest/scala/com/virtuslab/gitmachete/uitest/UITestSuite.scala
```diff
@@ -36,13 +37,12 @@ class UITestSuite extends BaseGitRepositoryBackedIntegrationTestSuite(SETUP_WITH
     waitAndCloseProject()
   }
 
-  @Ignore // FIXME (#830): ui test failing - socket timeout
   @Test def skipNonExistentBranches_toggleListingCommits_slideOutRoot(): Unit = {
     // TODO (#830): try ... catch block to discover why the SocketTimeoutException occurs
```
PawelLipski, line 41: Comment can be removed

azdrojowa123:
The `rule`  still doesn't work 😂 I need to repair it

PawelLipski:
Oh! As in, JUnit does not detect the rule?

PawelLipski:
> It doesn't even compile

Hmm what's the error?

src/uiTest/scala/com/virtuslab/gitmachete/uitest/UITestSuite.scala
```diff
@@ -117,9 +131,7 @@ class UITestSuite extends BaseGitRepositoryBackedIntegrationTestSuite(SETUP_WITH
     Assert.assertEquals(0, branchRowsCount)
   }
 
-  @Test def fastForwardParentOfBranch(): Unit = {
-
-    // fastForwardParentOfBranch_parentIsCurrentBranch
+  @Test def fastForwardParentOfBranch_parentIsCurrentBranch(): Unit = {
```
PawelLipski, line 134: Hmmm looks like some conflict-resolution artifact, pls revert

src/uiTest/scala/com/virtuslab/gitmachete/uitest/UITestSuite.scala
```diff
@@ -36,62 +37,75 @@ class UITestSuite extends BaseGitRepositoryBackedIntegrationTestSuite(SETUP_WITH
...
+    )
+    project.toolbar.toggleListingCommits()
+    var branchAndCommitRowsCount = project.refreshModelAndGetRowCount()
+    // 7 branch rows + 11 commit rows
+    Assert.assertEquals(18, branchAndCommitRowsCount)
+
+    // Let's slide out a root branch now
+    project.contextMenu.openContextMenu("develop")
+    project.contextMenu.slideOut()
+    project.acceptBranchDeletionOnSlideOut()
```
PawelLipski, line 94: Huh wait... that's not the same code as used to be inside `try ... catch` block... is it? 🤔 

That might be why the UI test is failing

azdrojowa123: now in #skipNonExistentBranches_toggleListingCommits_slideOutRoot is the same content as on `develop`

PR #941, opened 2022-06-23
Click in the middle in case of any button in ui test

mkondratek:


PawelLipski:
As discussed... to move our attention away from spuriously failing UI tests, maybe just set the git config so that the slide out dialog doesn't show up 😅

testCommon/src/test/resources/setup-with-single-remote.sh
```diff
@@ -10,6 +10,7 @@ create_repo machete-sandbox-remote --bare
 create_repo machete-sandbox
 cd machete-sandbox
   git remote add origin ../machete-sandbox-remote
+  git config --add machete.slideOut.deleteLocalBranch true
```
mkondratek, line 13: let's see...

Issue #945, opened 2022-06-26
`uiTest_<intellij-version>` Gradle tasks are effectively uncachable


PawelLipski:
`./gradlew --info -Pheadless -Pagainst=2022.1.1 uiTests`:

```
> Task :uiTest_2022.1.1
Custom actions are attached to task ':uiTest_2022.1.1'.
Caching disabled for task ':uiTest_2022.1.1' because:
  Gradle does not know how file 'build/idea-sandbox/system-test' was created (output property 'System directory'). Task output caching requires exclusive access to output paths to guarantee correctness (i.e. multiple tasks are not allowed to produce output in the same location).
```

Indeed `build/idea-sandbox/system-test` is also written by another task, `:test`, as one can check via the following task:

```groovy
allprojects {
  task printTaskInputsAndOutputs {
    doLast {
      tasks.each { task ->
        println "--------------------------------------------------------------------------------"
        println " Task '${project.name}:${task.name}'"
        println "--------------------------------------------------------------------------------"
        println ""

        println "File inputs:"
        task.inputs.files.toSorted().each {println " - ${it}" }
        println ""

        println "Property inputs:"
        task.inputs.properties.collect { k, v -> "$k = $v"}.toSorted().each { println it }
        println ""

        println "File outputs:"
        task.outputs.files.toSorted().each {println " - ${it}" }
        println ""

        println "--------------------------------------------------------------------------------"
        println ""
      }
    }
  }
}
```

Generally `:uiTest` has the following outputs (same as `:test`): 
```
 - System directory (Output), /home/plipski/git-machete-intellij-plugin/build/idea-sandbox/system-test
 - binaryResultsDirectory (Output), /home/plipski/git-machete-intellij-plugin/build/test-results/uiTest_2022.1.1/binary
 - reports.enabledReports.html.outputLocation (Output), /home/plipski/git-machete-intellij-plugin/build/reports/tests/uiTest_2022.1.1
 - reports.enabledReports.junitXml.outputLocation (Output), /home/plipski/git-machete-intellij-plugin/build/test-results/uiTest_2022.1.1
```

This is coming from the depths of gradle-intellij-plugin: IntelliJPlugin.kt:

```kotlin
        val testTasks = project.tasks.withType(Test::class.java)
        // ...
        testTasks.forEach { task ->
            // ....
            task.outputs.dir(systemDirectoryProvider).withPropertyName("System directory")
```

[...]

PR #946, opened 2022-06-26
Resolve Gradle caching issues (overlapping output directories for `generateMacheteParser` and `generateMacheteLexer`)

PawelLipski:
This caused both tasks to be run with almost every build (or at least much more frequently that they needed to)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -68,15 +68,6 @@ jobs:
       - run:
           name: Run unit & integration tests
           command: ./gradlew test
-      # Unfortunately, wildcards for test result paths aren't supported by CircleCI yet.
-      - store_test_results:
-          path: branchLayout/impl/build/test-results/test/
-      - store_test_results:
-          path: backend/impl/build/test-results/test/
-      - store_test_results:
-          path: frontend/base/build/test-results/test/
-      - store_test_results:
-          path: build/test-results/test/
 
       - run:
           name: Build plugin artifact
@@ -116,11 +107,21 @@ jobs:
             mkdir -p /exception-artifacts
             for f in $(ls -d /tmp/ide-probe/screenshots/*/) ; do cd  $f/artifacts ; zip -r "/exception-artifacts/$(basename ${f}).zip" . ; done
           when: on_fail
-      - store_test_results:
-          path: build/test-results/uiTest/
       - store_artifacts:
           path: /exception-artifacts
           destination: .
+
+      # Unfortunately, wildcards for test result paths aren't supported by CircleCI yet.
+      - store_test_results:
+          path: branchLayout/impl/build/test-results/test/
+      - store_test_results:
+          path: backend/impl/build/test-results/test/
+      - store_test_results:
+          path: frontend/base/build/test-results/test/
+      - store_test_results:
+          # This include the test results of both "regular" tests and UI tests for the top-level project.
+          path: build/test-results/
+
       - when:
           condition:
             not:
--- a/backend/impl/build.gradle
+++ b/backend/impl/build.gradle
@@ -14,6 +14,8 @@ dependencies {
   api project(':backend:api')
   api project(':branchLayout:api')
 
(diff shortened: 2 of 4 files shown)


build.gradle
```diff
@@ -654,18 +657,17 @@ uiTestTargets.each { version ->
     group = 'verification'
 
     testClassesDirs = uiTest.output.classesDirs
-    classpath = sourceSets.main.output + configurations.uiTestRuntimeClasspath + uiTest.output
-    // Let's keep the XML test results in a per-version directory, this makes things easier in the CI
-    // where the UI tests are executed under a couple different IDE versions one after another.
-    // The results in the default directory (build/test-results/uiTest) would get overwritten which each subsequent test.
-    reports.junitXml.destination = file("$buildDir/test-results/uiTest/${version}")
```
PawelLipski, line 661: Unrel, also removed... doesn't completely fix the problem with UI test caching (since there's still https://github.com/VirtusLab/git-machete-intellij-plugin/issues/945) but gets it a bit closed.

XML test results are still in per-version directory, just sth like `build/test-results/uiTest_2021.1.1` instead of `build/test-results/uiTest/2021.1.1`... still fine, and removes a weird problem with Gradle caching (still, #945 is harder to get rid of)

build.gradle
```diff
@@ -654,18 +657,17 @@ uiTestTargets.each { version ->
...
+    classpath = configurations.uiTestRuntimeClasspath + uiTest.output
 
     dependsOn ':buildPlugin'
 
     systemProperty 'ui-test.intellij.version', version
     systemProperty 'ui-test.plugin.path', buildPlugin.outputs.files[0].path
 
-    outputs.upToDateWhen { false }
+    // TODO (#945): caching of UI test results doesn't work in the CI anyway
```
PawelLipski, line 667: So effectively, it's like `outputs.upToDateWhen { false }` even in CI, which means extra minutes to each build :/

frontend/file/build.gradle
```diff
@@ -15,25 +15,34 @@ apply plugin: 'org.jetbrains.grammarkit'
 import org.jetbrains.grammarkit.tasks.*
 
 def grammarSourcesRoot = 'src/main/grammar'
-def generatedJavaSourcesRoot = 'build/generated'
+// Outputs of these two tasks canNOT go into the same directory,
```
PawelLipski, line 18: And here's the actual fix from the PR :trollface:

backend/impl/build.gradle
```diff
@@ -14,6 +14,8 @@ dependencies {
   api project(':backend:api')
   api project(':branchLayout:api')
 
+  // Note that we can't easily use Gradle's `testFixtures` configuration here
+  // as it doesn't seem to expose testFixtures resources in test classpath correctly.
```
PawelLipski, line 18: Unrelated observation so that I don't reattempt using https://docs.gradle.org/current/userguide/java_testing.html#sec:java_test_fixtures

build.gradle
```diff
@@ -427,10 +427,13 @@ allprojects {
 
     ideProbe = { ->
       repositories {
+        // Needed for com.intellij.remoterobot:remote-robot
```
PawelLipski, line 430: I thought it's for ide-probe itself but no

.circleci/config.yml
```diff
@@ -116,11 +107,21 @@ jobs:
...
+      # Unfortunately, wildcards for test result paths aren't supported by CircleCI yet.
+      - store_test_results:
+          path: branchLayout/impl/build/test-results/test/
+      - store_test_results:
+          path: backend/impl/build/test-results/test/
+      - store_test_results:
+          path: frontend/base/build/test-results/test/
+      - store_test_results:
+          # This include the test results of both "regular" tests and UI tests for the top-level project.
+          path: build/test-results/
```
PawelLipski, line 123: See build.gradle for the test-results path reorganization

azdrojowa123 (approved):

PR #956, opened 2022-06-29
Docker setup for running container local (uiTests)

azdrojowa123:


CONTRIBUTING.md
```diff
@@ -122,6 +122,15 @@ To watch the logs of this IntelliJ instance, run `tail -f build/idea-sandbox/sys
 See [Robot plugin](https://github.com/JetBrains/intellij-ui-test-robot)
 and [a preso on testing UI of IntelliJ Plugins](https://slides.com/plipski/intellij-plugin-ui-testing) for more details.
 
+### macOS
+It is not possible to run `uiTest` gradle task on macOS systems directly via `./gradlew`.
+You can run tests inside docker container (based on `gitmachete/intellij-plugin-ci:6.2.0` image) or using virtual machine.
+Sample configuration for launching docker container is shown below.
+```shell
+docker run -v [path to git-machete-intellij-plugin root directory e.g. "$PWD"]:/home/git-machete-intellij-plugin -v ~/.gradle/caches:/root/.gradle/caches -w /home/git-machete-intellij-plugin gitmachete/intellij-plugin-ci:6.2.0 ./gradlew -Pagainst=2021.2 -Pheadless uiTest
```
PawelLipski, line 130: Note that with `~/.gradle/caches` mounted, Gradle itself will be downloaded with every `docker run`:

```
docker run -v $PWD:/home/git-machete-intellij-plugin -v ~/.gradle/caches:/root/.gradle/caches -w /home/git-machete-intellij-plugin gitmachete/intellij-plugin-ci:6.2.0 ./gradlew -Pagainst=2021.2 -Pheadless uiTest
Downloading https://services.gradle.org/distributions/gradle-7.4.2-bin.zip
...........10%...........20%...........30%...........40%...........50%...........60%...........70%...........80%...........90%...........100%
```

Maybe expand the mount to `~/.gradle` instead?

CONTRIBUTING.md
```diff
@@ -122,6 +122,15 @@ To watch the logs of this IntelliJ instance, run `tail -f build/idea-sandbox/sys
 See [Robot plugin](https://github.com/JetBrains/intellij-ui-test-robot)
 and [a preso on testing UI of IntelliJ Plugins](https://slides.com/plipski/intellij-plugin-ui-testing) for more details.
 
+### macOS
+It is not possible to run `uiTest` gradle task on macOS systems directly via `./gradlew`.
+You can run tests inside docker container (based on `gitmachete/intellij-plugin-ci:6.2.0` image) or using virtual machine.
+Sample configuration for launching docker container is shown below.
+```shell
+docker run -v [path to git-machete-intellij-plugin root directory e.g. "$PWD"]:/root/git-machete-intellij-plugin -v ~/.gradle:/root/.gradle -w /root/git-machete-intellij-plugin gitmachete/intellij-plugin-ci:6.2.0 ./gradlew -Pagainst=2021.2 -Pheadless uiTest
```
PawelLipski, line 130: Also! worth adding
`-v ~/.pluginVerifier:/root/.pluginVerifier -v /tmp/ide-probe/cache:/tmp/ide-probe/cache`

so that ide-probe can re-use the IntelliJs downloaded locally

PawelLipski: Plus add `--rm` so that container is removed after completion... otherwise local Gradle won't be able to e.g.:
```
Starting a Gradle Daemon, 1 busy and 2 incompatible and 7 stopped Daemons could not be reused, use --status for details

FAILURE: Build failed with an exception.

* What went wrong:
java.io.UncheckedIOException: java.io.IOException: Unable to delete directory '/home/plipski/git-machete-intellij-plugin/.gradle/7.4.2/dependencies-accessors/2e63973890a031160de15c2136640c23e22ba214/sources'
  Failed to delete some children. This might happen because a process has files open or has its working directory set in the target directory.
  - /home/plipski/git-machete-intellij-plugin/.gradle/7.4.2/dependencies-accessors/2e63973890a031160de15c2136640c23e22ba214/sources/org/gradle/accessors/dm/LibrariesForLibs.java
  - /home/plipski/git-machete-intellij-plugin/.gradle/7.4.2/dependencies-accessors/2e63973890a031160de15c2136640c23e22ba214/sources/org/gradle/accessors/dm
  - /home/plipski/git-machete-intellij-plugin/.gradle/7.4.2/dependencies-accessors/2e63973890a031160de15c2136640c23e22ba214/sources/org/gradle/accessors
  - /home/plipski/git-machete-intellij-plugin/.gradle/7.4.2/dependencies-accessors/2e63973890a031160de15c2136640c23e22ba214/sources/org/gradle
  - /home/plipski/git-machete-intellij-plugin/.gradle/7.4.2/dependencies-accessors/2e63973890a031160de15c2136640c23e22ba214/sources/org
> java.io.IOException: Unable to delete directory '/home/plipski/git-machete-intellij-plugin/.gradle/7.4.2/dependencies-accessors/2e63973890a031160de15c2136640c23e22ba214/sources'
    Failed to delete some children. This might happen because a process has files open or has its working directory set in the target directory.
    - /home/plipski/git-machete-intellij-plugin/.gradle/7.4.2/dependencies-accessors/2e63973890a031160de15c2136640c23e22ba214/sources/org/gradle/accessors/dm/LibrariesForLibs.java
    - /home/plipski/git-machete-intellij-plugin/.gradle/7.4.2/dependencies-accessors/2e63973890a031160de15c2136640c23e22ba214/sources/org/gradle/accessors/dm
    - /home/plipski/git-machete-intellij-plugin/.gradle/7.4.2/dependencies-accessors/2e63973890a031160de15c2136640c23e22ba214/sources/org/gradle/accessors
    - /home/plipski/git-machete-intellij-plugin/.gradle/7.4.2/dependencies-accessors/2e63973890a031160de15c2136640c23e22ba214/sources/org/gradle
    - /home/plipski/git-machete-intellij-plugin/.gradle/7.4.2/dependencies-accessors/2e63973890a031160de15c2136640c23e22ba214/sources/org

```

PawelLipski: Oh! also... there's a problem that any files are written to any mounted folder... will be owned by the **root**, not the local user on the host machine ☹️ probably setting up uid/gid to run explictly will be needed... if easily doable :/

azdrojowa123: or maybe I can run container as non root ?

PawelLipski: yes, exactly, pls try

CONTRIBUTING.md
```diff
@@ -122,6 +122,15 @@ To watch the logs of this IntelliJ instance, run `tail -f build/idea-sandbox/sys
 See [Robot plugin](https://github.com/JetBrains/intellij-ui-test-robot)
 and [a preso on testing UI of IntelliJ Plugins](https://slides.com/plipski/intellij-plugin-ui-testing) for more details.
 
+### macOS
+It is not possible to run `uiTest` gradle task on macOS systems directly via `./gradlew`.
+You can run tests inside docker container (based on `gitmachete/intellij-plugin-ci:6.2.0` image) or using virtual machine.
+Sample configuration for launching docker container is shown below.
+```shell
```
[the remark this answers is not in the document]
azdrojowa123: I have some problems when try to run second container while first one is still running
Screenshot 2022-06-29 at 14 38 44

CONTRIBUTING.md
PawelLipski: > And for what purpose ~/.pluginVerifier should be mounted ?

See ideprobe.conf -> `probe.resolvers.intellij.repositories`... we re-use the IDE downloaded by plugin verifier

Dockerfile
```diff
@@ -36,6 +36,8 @@ RUN \
   apt-get install --no-install-recommends -y libxtst6 xvfb xauth x11-apps imagemagick zip; \
   `# libraries needed by JCEF` \
   apt-get install --no-install-recommends -y libatk1.0-0 libatk-bridge2.0-0 libgbm1 libglib2.0 libxcomposite1 libxkbcommon0 libxrandr2; \
+  useradd -m user; \
```
PawelLipski, line 39: Hmmm but note that without `USER` instruction in Dockerfile, the code in the container will still run as root user 🤔

And hence the files created within the container will be owned by root on the host... can you actually observe this behavior on Mac OS? maybe the mechanism is different there

PawelLipski: > is owned by this user

Okay... but I suppose the UID/GID of the user in Docker container and of the one in your machine are different nevertheless? Pls check with `id -a` in both places

azdrojowa123: in what two places ? I do not have this `user` on my machine (I probably missed something). I can only check UID/GID in container or directly while creating the image

PawelLipski: Just run:

```
$ id -a
$ docker run <your-image> id -a
```

PawelLipski: The user **names** don't matter, as long as their numeric **ids** are the same!

PawelLipski: I suppose your host 503:20 (uid:gid) user can**not** write files created by container 1000:1000 user, right?

azdrojowa123: you mean exec into created container as my host 503:20 ? if so then I am not able to do it

PawelLipski: I mean sth like that:
```
docker run -v <some-local-dir>:/foo <your-image> touch /foo/bar.txt
```

Now `<some-local-dir>/bar.txt` exists on the host, but is owned by 1000:1000... and will have `rw-r--r--` permissions, hence your local user (503:20) won't be able to write to this file.

Pls check if this holds true on Mac!

PawelLipski:
In general... see #964 (PR to be declined) as a demo on how to switch to user for the given UID/GID **dynamically** (i.e. in container runtime, not in image build time)

Dockerfile
azdrojowa123: if I run container in such a way: 
` docker run -v "PWD":/foo gitmachete/intellij-plugin-ci:6.2.1 /bin/bash -c "touch /foo/bar.txt; while true; do sleep 1; done"`
foo.txt is owned by root (because docker run execute as root) and when I e.g. execute `docker run -u user` and run uiTest, artifacts are stored inside `/home/user/artifacts/` and they are owned by `user`

azdrojowa123: and as local user (503:20) I cannot event exec into container (because such user not exist in image)

azdrojowa123: I will try to rework as in #964

PawelLipski: Discussed on Slack... probably Docker for Mac has some mapping of UID/GID such that every file created from container in the shared volume, on host belongs just to the host user (and not to whatever UID/GID were used in the container)

CONTRIBUTING.md
```diff
@@ -122,6 +122,23 @@ To watch the logs of this IntelliJ instance, run `tail -f build/idea-sandbox/sys
...
 and [a preso on testing UI of IntelliJ Plugins](https://slides.com/plipski/intellij-plugin-ui-testing) for more details.
 
+### macOS
+It is not possible to run `uiTest` gradle task in display mode `Xvfb` on macOS systems directly via `./gradlew`.
+You can run ui tests inside docker container (based on `gitmachete/intellij-plugin-ci:6.2.1` image) or using virtual machine.
+Sample configuration for launching docker container is shown below. Environment variables `UID` and `GID` placed into the container will be the user and group identifiers
+of the files created inside it.
+```shell
+docker run --rm -e UID=$UID -e GID=$GID \
+      -v [path to git-machete-intellij-plugin root directory e.g. "$PWD"]:/home/docker/git-machete-intellij-plugin \
```
PawelLipski, line 132: For even easier copying directly to the terminal, I'd leave:
```suggestion
      -v $PWD:/home/docker/git-machete-intellij-plugin \
```
and mention somewhere that it's assumed that current working directory is the top-level directory of the project

CONTRIBUTING.md
```diff
@@ -122,6 +122,23 @@ To watch the logs of this IntelliJ instance, run `tail -f build/idea-sandbox/sys
...
+      -v ~/.gradle:/home/docker/.gradle \
+      -v ~/.pluginVerifier:/home/docker/.pluginVerifier \
+      -v /tmp/ide-probe/cache:/tmp/ide-probe/cache \
+      -w /home/docker/git-machete-intellij-plugin \
+      gitmachete/intellij-plugin-ci:6.2.1 \
```
PawelLipski, line 137: Nit: 6.3.0 (coz new package has been added to the image, semver FTW) + update the image number in CircleCI config

docker/Dockerfile
```diff
@@ -35,7 +35,7 @@ RUN \
   apt-get update; \
   apt-get install --no-install-recommends -y libxtst6 xvfb xauth x11-apps imagemagick zip; \
   `# libraries needed by JCEF` \
-  apt-get install --no-install-recommends -y libatk1.0-0 libatk-bridge2.0-0 libgbm1 libglib2.0 libxcomposite1 libxkbcommon0 libxrandr2; \
+  apt-get install --no-install-recommends -y libatk1.0-0 libatk-bridge2.0-0 libgbm1 libglib2.0 libxcomposite1 libxkbcommon0 libxrandr2 sudo; \
```
PawelLipski, line 38: Nit: I'd move it to line 36, current placing suggests that it's needed specifically for JCEF (line 37)

PawelLipski: Or even better, to line 50 where all such "outlier" packages are located

docker/Dockerfile
```diff
@@ -55,3 +55,6 @@ RUN \
   curl -Lf https://npmjs.org/install.sh | sh; \
   npm install --global remark-cli remark-lint-no-dead-urls remark-validate-links; \
   npm uninstall --global npm
+COPY entrypoint.sh /entrypoint.sh
```
PawelLipski, line 58: Nit: empty line before this one

src/uiTest/resources/ideprobe.conf
```diff
@@ -1,5 +1,6 @@
 probe {
   paths.screenshots = ${?IDEPROBE_SCREENSHOTS_DIR}
```
PawelLipski, line 2: Nit: I'd actually keep the env var names as aligned with the corresponding config keys as possible:
```suggestion
  paths.screenshots = ${?IDEPROBE_PATHS_SCREENSHOTS}
```

src/uiTest/resources/ideprobe.conf
```diff
@@ -1,5 +1,6 @@
...
+  paths.base = ${?IDEPROBE_BASE_PATH}
```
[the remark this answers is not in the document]
PawelLipski: Doesn't matter much here for 2 env vars, but once there are 50+ of them, it gets relevant pretty quickly 😅

CONTRIBUTING.md
```diff
@@ -122,6 +122,23 @@ To watch the logs of this IntelliJ instance, run `tail -f build/idea-sandbox/sys
 See [Robot plugin](https://github.com/JetBrains/intellij-ui-test-robot)
 and [a preso on testing UI of IntelliJ Plugins](https://slides.com/plipski/intellij-plugin-ui-testing) for more details.
 
+### macOS
+It is not possible to run `uiTest` gradle task in display mode `Xvfb` on macOS systems directly via `./gradlew`.
+You can run ui tests inside docker container (based on `gitmachete/intellij-plugin-ci:6.2.1` image) or using virtual machine.
```
PawelLipski, line 127:
```suggestion
You can run ui tests inside docker container (based on `gitmachete/intellij-plugin-ci:6.2.1` image) or using a Linux virtual machine.
```

CONTRIBUTING.md
```diff
@@ -122,6 +122,23 @@ To watch the logs of this IntelliJ instance, run `tail -f build/idea-sandbox/sys
...
+Sample configuration for launching docker container is shown below. Environment variables `UID` and `GID` placed into the container will be the user and group identifiers
+of the files created inside it.
+```shell
+docker run --rm -e UID=$UID -e GID=$GID \
```
PawelLipski, line 131:
```suggestion
docker run --rm -e UID=$(id -u) -e GID=$(id -g) \
```

CONTRIBUTING.md
```diff
@@ -122,6 +122,23 @@ To watch the logs of this IntelliJ instance, run `tail -f build/idea-sandbox/sys
...
+      -v [path to git-machete-intellij-plugin root directory e.g. "$PWD"]:/home/docker/git-machete-intellij-plugin \
+      -v ~/.gradle:/home/docker/.gradle \
+      -v ~/.pluginVerifier:/home/docker/.pluginVerifier \
```
PawelLipski, line 134: What if this dir doesn't exist  yet on host? 🤔

CONTRIBUTING.md
```diff
@@ -122,6 +122,23 @@ To watch the logs of this IntelliJ instance, run `tail -f build/idea-sandbox/sys
...
+      -v /tmp/ide-probe/cache:/tmp/ide-probe/cache \
+      -w /home/docker/git-machete-intellij-plugin \
+      gitmachete/intellij-plugin-ci:6.2.1 \
+      ./gradlew -Pagainst=2021.2 -Pheadless uiTest
```
PawelLipski, line 138: I'd skip `-Pagainst` from the sample, typically the user wants to test against the default version (i.e. `buildTarget`) anyway

docker/entrypoint.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+
+set -e -o pipefail
```
PawelLipski, line 3: Hmm no `-u`?

docker/entrypoint.sh
```diff
@@ -0,0 +1,14 @@
...
+
+if [[ $GID && $UID ]]; then
```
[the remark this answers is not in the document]
PawelLipski: And I'd leave `-u` to make the script future-proof

build.gradle
```diff
@@ -678,8 +678,8 @@ uiTestTargets.each { version ->
 
     if (project.hasProperty('headless')) {
       environment 'IDEPROBE_DISPLAY', 'xvfb'
-      environment 'IDEPROBE_SCREENSHOTS_DIR', System.getProperty("user.home") + '/artifacts/uiTest' + version + '/screenshots'
-      environment 'IDEPROBE_BASE_PATH', System.getProperty("user.home")
+      environment 'IDEPROBE_PATHS_SCREENSHOTS', System.getProperty("user.home") + '/artifacts/uiTest' + version + '/screenshots'
+      environment 'IDEPROBE_PATHS_BASE', System.getProperty("user.home")
```
PawelLipski, line 682: Whoops in case of local runs (no CI, no VM, no Docker)... won't that pollute home directory with a few new directories (of little importance, other than for UI testing)?

PawelLipski: TBH I'd wrap both `environment 'IDEPROBE_PATHS_...', ...` statements into e.g. `if (!isCi) { ... }`... or sth like that 🤔

azdrojowa123: so maybe wrap `IDEPROBE_PATHS_BASE` into `if (!isCI) { ... }.` ? but  in that case I need to write in CONTRIBUTING.md that this variable must be set when starting the container

PawelLipski: Okay, let's do it this way

PawelLipski: Maybe actually... use sth like `$HOME/.ideprobe/` for both base path and screenshots? I mean, using non-dotted (`.*`) directories under home directory like `~/artifacts` is still pretty unconventional/surprising :/

build.gradle
```diff
@@ -678,7 +678,10 @@ uiTestTargets.each { version ->
 
     if (project.hasProperty('headless')) {
       environment 'IDEPROBE_DISPLAY', 'xvfb'
-      environment 'IDEPROBE_SCREENSHOTS_DIR', System.getProperty("user.home") + '/artifacts/uiTest' + version + '/screenshots'
+      environment 'IDEPROBE_PATHS_SCREENSHOTS', System.getProperty("user.home") + '/artifacts/uiTest' + version + '/screenshots'
+      if (isCi) {
```
PawelLipski, line 682: `isCi` or `isCI`? 🤔

PawelLipski (approved):

PawelLipski:
Huh the weird error in CI probably comes from a new version of `remark-*` getting installed when the image has been regenerated 🤔 it's likely to be a known issue, search GitHub... and if not, just e.g. select some lower version of the faulty package

CONTRIBUTING.md
```diff
@@ -122,6 +122,24 @@ To watch the logs of this IntelliJ instance, run `tail -f build/idea-sandbox/sys
...
+      -v "$PWD":/home/docker/git-machete-intellij-plugin \
+      -v ~/.gradle:/home/docker/.gradle \
+      -v ~/.pluginVerifier:/home/docker/.pluginVerifier \
+      -v /tmp/ide-probe/cache:/tmp/ide-probe/cache \
+      -w /home/docker/git-machete-intellij-plugin \
+      gitmachete/intellij-plugin-ci:6.3.0 \
+      ./gradlew -Pheadless uiTest
+```
+
+
```
PawelLipski, line 143: Also! since Dockerfile has been moved, pls update the section here where building & pushing Docker image is documented

src/uiTest/resources/ideprobe.conf
```diff
@@ -1,5 +1,6 @@
 probe {
-  paths.screenshots = ${?IDEPROBE_SCREENSHOTS_DIR}
+  paths.screenshots = ${?IDEPROBE_PATHS_SCREENSHOTS}
+  paths.base = ${?IDEPROBE_PATHS_BASE}
```
PawelLipski, line 3:
```suggestion
  paths {
    base = ${?IDEPROBE_PATHS_BASE}
    screenshots = ${?IDEPROBE_PATHS_SCREENSHOTS}
  }
```

azdrojowa123:
> Huh the weird error in CI probably comes from a new version of `remark-*` getting installed when the image has been regenerated 🤔 it's likely to be a known issue, search GitHub... and if not, just e.g. select some lower version of the faulty package

new `remark` drop support for node 12 so after upgrading to something higher (16) it works correctly

PR #964, opened 2022-07-02
Add a demo for how to switch user/group in Docker container in runtime (as opposed to build time)

PawelLipski:
Temporary demo for #956, **not** to be actually merged


PawelLipski:
Closing, already incorporated into #956

PR #967, opened 2022-07-05
Show repo path in repo combobox (closes #618)

### LINKED ISSUE #618: Consider showing sub repository path in case of multirepo projects
Presently, with repo in the project root and a few subrepos under one of its dirs we see the following:

image

This is a bit messy and inconsistent with the dvcs widget which shows:

image

mkondratek:
image


CHANGE-NOTES.html
```diff
@@ -1,3 +1,5 @@
 <p>The <i>Pull</i> action displays a clearer notification message when executed on branch that no longer exists in the remote repository.</p>
 
 <p>Stability fixes have been applied.</p>
+
+<p>Show a path from repository root in the combobox.</p>
```
PawelLipski, line 5: Pls refine a bit: which combobox/what was shown previously

PawelLipski: Plus overwrite rather than append

PawelLipski (approved):

PR #1009, opened 2022-07-28
Partially fix the execution of `versionCatalogUpdate` task

PawelLipski:


build.gradle.kts
```diff
@@ -43,15 +43,26 @@ val shouldRunAllCheckers by extra(isCI || project.hasProperty("runAllCheckers"))
 tasks.register<UpdateEapBuildNumber>("updateEapBuildNumber")
 
 tasks.withType<DependencyUpdatesTask> {
-  val isStableVersion: (String) -> Boolean = { version ->
+  fun isStableVersion(version: String): Boolean {
     val stableKeyword = listOf("RELEASE", "FINAL", "GA").any { version.toUpperCase().contains(it) }
-    val regex = "/^[0-9,.v-]+(-r)?$/".toRegex()
```
PawelLipski, line 48: Whoops `/.../` are an equivalent of `toRegex()` in Groovy (syntax taken from Perl); they weren't supposed to be a part of the string regex itself

build.gradle.kts
```diff
@@ -43,15 +43,26 @@ val shouldRunAllCheckers by extra(isCI || project.hasProperty("runAllCheckers"))
...
-    stableKeyword || regex.matches(version)
+    val regex = "^[0-9,.v-]+(-r)?$".toRegex()
+    return stableKeyword || regex.matches(version)
   }
 
   rejectVersionIf { !isStableVersion(this.candidate.version) }
 }
 
+// TODO (#1004): make it possible bump dependencies of buildSrc as well (and not only of the project itself)
+versionCatalogUpdate {
```
PawelLipski, line 56: The change is that `gradle/libs.versions.toml` doesn't get completely destroyed after running `versionCatalogUpdate`... but still, these buildSrc-only dependencies are ignored in the update; workaround of https://github.com/ben-manes/gradle-versions-plugin/issues/284 FTW

PawelLipski: Once the workaround is applied, this entire section can be removed (hopefully)

PawelLipski: Follow-up on #1013, the section canNOT be removed, alas

Actually, even one more such section added :trollface:

MaciejG604 (approved):

PR #1015, opened 2022-07-29
Use more concise Kotlin DSL constructs across Gradle config

PawelLipski:

PR #1016, opened 2022-07-29
Set Kotlin DSL jvmtarget to 11

MaciejG604:


PawelLipski (approved):

PawelLipski:
Generally let's only merge PRs once they're based on `develop`. It doesn't harm much in this case, since both #1016 and its base #1015 are both small. In general, however, merging PR to a non-`develop` base leads to large, un-focused PRs... which we want to avoid in the first place

PR #1022, opened 2022-07-30
Fix scripts/utils.sh so that no extra newline is printed by machete-status-branch hook in case version cannot be found

PawelLipski:


scripts/utils.sh
```diff
@@ -53,27 +53,3 @@ function extract_version_from_git_revision() {
 
   git show "$revision":version.gradle.kts | extract_version_from_gradle_file_stdin
 }
-
-function parse_version() {
```
PawelLipski, line 57: They weren't used for some time

MaciejG604 (approved):

PR #1041, opened 2022-08-08
Add task that looks for new IntelliJ stable releases

MaciejG604:


MaciejG604:
Just remembered that I need to update CONTRIBUTING.md

buildSrc/src/main/kotlin/com/virtuslab/gitmachete/buildsrc/IntellijVersionHelper.kt
```diff
@@ -54,4 +54,18 @@ object IntellijVersionHelper {
   private fun getFile(): File {
     return File("intellijVersions.properties")
   }
+
+  infix fun String.versionIsNewerThan(rhsBuildNumber: String): Boolean {
```
PawelLipski, line 58: Actually! to work around the problem with Strings sometimes keeping `222.1234.56`-style build numbers, and sometimes `2022.2`-style versions, I'd consider using _newtypes_, called _inline classes_ in Kotlin: https://kotlinlang.org/docs/inline-classes.html#calling-from-java-code

This way, an attempt to pass a build number where version is expected (or vice versa) will be detected at Kotlin compile time.

Can be done on a separate PR

PawelLipski: Also, comparision method can be called as `isNewerThan`, no need to specify `version`/`buildNumber` in method name, as it'll be obvious from operand types

buildSrc/src/main/kotlin/com/virtuslab/gitmachete/buildsrc/IntellijVersionHelper.kt
```diff
@@ -54,4 +54,18 @@ object IntellijVersionHelper {
...
+    val lhsSplit = this.split('.')
+    val rhsSplit = rhsBuildNumber.split('.')
```
PawelLipski, line 60: Hmmm `buildNumber`? 🤔 I wouldn't suggest that a `222.1234.56` build number is expected here... pls smooth out naming

CONTRIBUTING.md
```diff
@@ -279,9 +279,8 @@ For instance:
    and see if the CI pipeline passes (this will both check binary compatibility and run UI tests against the given EAP)
 4. we release the plugin as `0.8.0` (`untilBuild` will extend automatically to `2020.3.*`
    via `latestSupportedMajor` in [IntellijVersions](buildSrc/src/main/kotlin/com/virtuslab/gitmachete/buildsrc/IntellijVersions.kt))
-5. once the stable `2020.3` is released, we verify ASAP that `0.8.0` is binary compatible with `2020.3` as well;
-   then, `latestStable` can be updated to `2020.3`, `eapOfLatestSupportedMajor` can be set to `null`,
-   and `2020.2.x` can be added to `latestMinorsOfOldSupportedMajors`
+5. new stable version can be detected automatically using `./gradlew updateIntellijStableVersions`,
```
PawelLipski, line 282: Pls update the entire section, assuming as for now that the project will be left on _autopilot_, i.e. all updates to intellijVersions.properties will be (ideally) done automatically by the bot, and never by the developers. It'll still be useful to document the logic&rationale behind these updates, however

build.gradle.kts
```diff
@@ -52,6 +52,7 @@ val compileJavaJvmArgs by extra((project.properties["compileJavaJvmArgs"] as Str
 val shouldRunAllCheckers by extra(isCI || project.hasProperty("runAllCheckers"))
 
 tasks.register<UpdateEapBuildNumber>("updateEapBuildNumber")
+tasks.register<UpdateIntellijStableVersions>("updateIntellijStableVersions")
```
PawelLipski, line 55: Actually... haven't checked the implementation yet, but I'm curious if it won't be easier to just merge these tasks into a single `updateIntellijVersions`, WDYT?

PawelLipski: Which would update both EAP and stable versions, depending on what's available

MaciejG604: Sure, why not!

buildSrc/src/main/kotlin/com/virtuslab/gitmachete/buildsrc/UpdateEapBuildNumber.kt
```diff
@@ -14,18 +15,18 @@ open class UpdateEapBuildNumber : DefaultTask() {
...
     val links =
       htmlContent.select(
-        "a[href^=${intellijSnapshotsUrl}com/jetbrains/intellij/idea/ideaIC/][href$=\"-EAP-SNAPSHOT.pom\"]"
+        "a[href^=${intellijSnapshotsUrl}com/jetbrains/intellij/idea/ideaIC/][href$=-EAP-SNAPSHOT.pom]"
       )
 
     for (link in links) {
       val attr = link.attr("href")
-      val regex = Regex("(?<=ideaIC-)\\d+\\.\\d+\\.\\d+(?=-EAP-SNAPSHOT.pom)")
+      val regex = Regex("(?<=ideaIC-)(\\d+\\.)+\\d+(?=-EAP-SNAPSHOT.pom)")
```
PawelLipski, line 23: Oh, to capture `222.1234` format rather than `222.1234.56`? are there such occurrences?

buildSrc/src/main/kotlin/com/virtuslab/gitmachete/buildsrc/UpdateIntellijStableVersions.kt
```diff
@@ -0,0 +1,83 @@
...
+
+        if (foundVersionNumber versionIsNewerThan latestStable) {
+          return foundVersionNumber
+        }
+      }
+    }
+    return null
+  }
+
+  fun getLatestMinorOfMajor(major: String): String {
```
PawelLipski, line 37: Quite a lot of repeated code, pls DRY up

buildSrc/src/main/kotlin/com/virtuslab/gitmachete/buildsrc/UpdateEapBuildNumber.kt
MaciejG604: Looks like there are NONE, will revert it.

buildSrc/src/main/kotlin/com/virtuslab/gitmachete/buildsrc/IntellijVersionHelper.kt
MaciejG604: Wouldn't that be an overkill? We don't have that many uses for version/build numbers. The only place where there could be a problem right now is the [Release repo](https://www.jetbrains.com/intellij-repository/releases/) where both version numbers and build numbers can be found, right now we rely simply on the fact that version numbers are always higher(newer) than build numbers.

PawelLipski: Okay, leaving that up to you then

buildSrc/src/main/kotlin/com/virtuslab/gitmachete/buildsrc/UpdateIntellijStableVersions.kt
```diff
@@ -0,0 +1,66 @@
...
+import org.gradle.api.DefaultTask
+import org.gradle.api.tasks.Internal
+import org.gradle.api.tasks.TaskAction
+import org.jsoup.Jsoup
+
+open class UpdateIntellijStableVersions : DefaultTask() {
+  @Internal
+  val intellijReleasesUrl: String = "https://www.jetbrains.com/intellij-repository/releases/"
+
+  private fun findMatchingVersionNewerThan(regex: Regex, thresholdVersion: String): String? {
```
PawelLipski, line 14: Pls make the naming consistent with `UpdateEapBuildNumber` (actually, I like `find` better than `check`), and consider DRYing up shared logic b/w two classes

buildSrc/src/main/kotlin/com/virtuslab/gitmachete/buildsrc/UpdateIntellijStableVersions.kt
```diff
@@ -0,0 +1,66 @@
...
+        val foundVersion = matchResult.value
+
+        if (foundVersion versionIsNewerThan thresholdVersion) {
+          return foundVersion
+        }
+      }
+    }
+    return null
+  }
+  fun checkForReleaseNewerThan(latestStable: String): String? {
```
PawelLipski, line 34: Pls make naming consistent wrt. `check`/`find`/`get`. Typically across software:
* `check`: indicates that an error will be thrown if condition is not satisfied
* `find`: return a value after a potentially work-intense lookup
* `get`: return a value, after a little to no lookup

PawelLipski: I'd say that `find` fits best for the methods here

buildSrc/src/main/kotlin/com/virtuslab/gitmachete/buildsrc/UpdateIntellijStableVersions.kt
```diff
@@ -0,0 +1,66 @@
...
+    return findMatchingVersionNewerThan(Regex("(?<=BUILD-)(\\d+\\.)+\\d+(?=.txt)"), latestStable)
+  }
+
+  fun getLatestMinorOfMajor(major: String): String {
```
PawelLipski, line 38: `private`?

buildSrc/src/main/kotlin/com/virtuslab/gitmachete/buildsrc/UpdateIntellijStableVersions.kt
```diff
@@ -0,0 +1,66 @@
...
+    return findMatchingVersionNewerThan(Regex("(?<=BUILD-)$major(\\.\\d+)?(?=.txt)"), major) ?: major
```
PawelLipski, line 39: Maybe even pass a predicate function as a second param? This will make the method more powerful, to be reused across classes

build.gradle.kts
```diff
@@ -52,6 +52,12 @@ val compileJavaJvmArgs by extra((project.properties["compileJavaJvmArgs"] as Str
...
+
+tasks.register("updateIntellijVersions") {
+  dependsOn("updateIntellijStableVersions", "updateEapBuildNumber")
+  tasks.getByName("updateEapBuildNumber").mustRunAfter("updateIntellijStableVersions")
```
PawelLipski, line 59: Hmmm I've run `updateIntellijVersions` on the current commit (`788aab04445bf258d5e34ab0baf1726a68bae268`) and ended up with:

```properties
#Mon Aug 08 14:21:34 CEST 2022
latestMinorsOfOldSupportedMajors=2020.3.4,2021.1.3,2021.2.4,2021.3.3,2022.1.4
eapOfLatestSupportedMajor=222.3739.24-EAP-SNAPSHOT
earliestSupportedMajor=2020.3
latestStable=2022.2
```

Here we've got both an EAP and a stable version for 2022.2, which makes little sense. EAPs basically only serve us to check whether we'll be compatible with the new major version once it's stable-released... once major version is stable-released we can just build against the stable version, not EAP :/

build.gradle.kts
```diff
@@ -52,6 +52,12 @@ val compileJavaJvmArgs by extra((project.properties["compileJavaJvmArgs"] as Str
 val shouldRunAllCheckers by extra(isCI || project.hasProperty("runAllCheckers"))
 
 tasks.register<UpdateEapBuildNumber>("updateEapBuildNumber")
+tasks.register<UpdateIntellijStableVersions>("updateIntellijStableVersions")
+
+tasks.register("updateIntellijVersions") {
+  dependsOn("updateIntellijStableVersions", "updateEapBuildNumber")
```
PawelLipski, line 58: TBD but I think that it doesn't even make much sense to keep the two tasks around, I doubt we'll ever need to run just one of them (it might in fact lead to unexpected results sometimes, esp. when e.g. we keep updating EAP for 223 once 2022.3 is released)

buildSrc/src/main/kotlin/com/virtuslab/gitmachete/buildsrc/UpdateIntellijStableVersions.kt
```diff
@@ -0,0 +1,66 @@
...
+  fun execute() {
+    val properties = IntellijVersionHelper.getProperties()
+
+    val latestStable = IntellijVersions.latestStable
+
+    val newerStable = checkForReleaseNewerThan(latestStable)
+
+    if (!newerStable.isNullOrEmpty()) {
+      project.logger.lifecycle("latestStable is updated to $newerStable")
+      properties.setProperty("latestStable", "$newerStable")
```
PawelLipski, line 52:
```suggestion
      properties.setProperty("latestStable", newerStable)
```

buildSrc/src/main/kotlin/com/virtuslab/gitmachete/buildsrc/UpdateIntellijStableVersions.kt
```diff
@@ -0,0 +1,66 @@
...
+      properties.remove("eapOfLatestSupportedMajor")
+      val majorOfLatestStable = getMajorPart(latestStable)
+      if (majorOfLatestStable != getMajorPart(newerStable)) {
+        val latestMinorsList = IntellijVersions.latestMinorsOfOldSupportedMajors.toMutableList()
+        val latestMinorOfReplacedMajor = getLatestMinorOfMajor(majorOfLatestStable)
+
+        latestMinorsList.removeIf { getMajorPart(it) == majorOfLatestStable }
```
PawelLipski, line 59: Hmmm will it be possible with the current logic to update the latest minor for an older version, like 2022.1? 🤔 they sometimes come out with a delay

PawelLipski, line 59: Hmmm will it be possible with the current logic to update the latest minor for an older version, like 2022.1? 🤔 they sometimes come out with a delay

PawelLipski: General suggestion: build the Versions object from scratch each time, basing just on `earliestSupportedMajor` and data extracted from HTMLs. This is much easier to reason about than the _stateful_ way were an existing state is incrementally modified

build.gradle.kts
MaciejG604: Shouldn't this be not allowed by the val buildNumberThreshold in UpdateEapBuildNumber?

PawelLipski: > this

By _this_, you mean the above update to properties file? 🤔

PawelLipski: This update seems to be the consistent result for `updateIntellijVersions` (i.e. not a race condition or sth), I've run the task a couple times 🤔

MaciejG604: I get it now, buildNumberThreshold is read from IntellijVersions, which is read from the intellijVersions.properties file but only at the beginning of the program execution. So after the task updateIntellijStableVersions finishes, the values in IntellijVersions are outdated and shouldn't be read.

MaciejG604: Should we update the values in IntellijVersions object after an update to the properties file happens to avoid any strange behaviour later in the same program execution?

PawelLipski: > Should we update the values in IntellijVersions object after an update to the properties file happens to avoid any strange behaviour later in the same program execution?

In the spirit of Functional Programming... I'd rather **limit side effects** (like saving stuff to file) to as narrow parts of the code as possible, and instead rely on **pure (deterministic && side-effect-free) functions**  wherever possible.

So the general outline would be:
1. read properties and HTML
2. do **all** processing => deterministic & side-effect-free => always the same result for the same inputs, can be reliably tested without any mocking
3. write the results => side-effect

buildSrc/src/main/kotlin/com/virtuslab/gitmachete/buildsrc/UpdateIntellijStableVersions.kt
MaciejG604: IMO to be considered if we actually find any more usages for this function, right now it will only make the code less readable

buildSrc/src/main/kotlin/com/virtuslab/gitmachete/buildsrc/UpdateIntellijVersions.kt
```diff
@@ -0,0 +1,90 @@
...
+import com.virtuslab.gitmachete.buildsrc.IntellijVersionHelper.toBuildNumber
+import com.virtuslab.gitmachete.buildsrc.IntellijVersionHelper.versionIsNewerThan
+import org.gradle.api.DefaultTask
+import org.gradle.api.tasks.Internal
+import org.gradle.api.tasks.TaskAction
+import org.jsoup.Jsoup
+
+open class UpdateIntellijVersions : DefaultTask() {
+  @Internal
+  val intellijReleasesUrl: String = "https://www.jetbrains.com/intellij-repository/releases/"
```
PawelLipski, line 13: Hmmm snapshots URL is missing 🤔

buildSrc/src/main/kotlin/com/virtuslab/gitmachete/buildsrc/UpdateIntellijVersions.kt
```diff
@@ -0,0 +1,90 @@
...
+      .map { findLatestMinorOfMajor(getMajorPart(it)) }
+      .plus(latestMinorOfReplacedMajor)
+  }
+
+  @TaskAction
+  fun execute() {
+    val properties = IntellijVersionHelper.getProperties()
+    val latestStable = IntellijVersions.latestStable
+
+    val newerStable = findReleaseNewerThan(latestStable)
```
PawelLipski, line 64: Note that you're downloading the page more than once... ofc the risk is minuscule, but it's theoretically possible that you'll download two different webpages on a each call, thus potentially leading to hard-to-nail bugs 🐞

buildSrc/src/main/kotlin/com/virtuslab/gitmachete/buildsrc/UpdateIntellijVersions.kt
```diff
@@ -0,0 +1,90 @@
...
+
+    if (!newerStable.isNullOrEmpty()) {
```
PawelLipski, line 66: Pls flip the `if`, negated conditions are (here unnecessarily) harder to reason about

buildSrc/src/main/kotlin/com/virtuslab/gitmachete/buildsrc/UpdateIntellijVersions.kt
```diff
@@ -0,0 +1,90 @@
...
+      properties.setProperty("latestStable", newerStable)
+      properties.remove("eapOfLatestSupportedMajor")
+
+      val majorOfLatestStable = getMajorPart(latestStable)
+      if (majorOfLatestStable != getMajorPart(newerStable)) {
+        val latestMinorsList = getUpdatedLatestMinorsList(majorOfLatestStable)
+        properties.setProperty("latestMinorsOfOldSupportedMajors", latestMinorsList.joinToString(separator = ","))
+      }
+      IntellijVersionHelper.storeProperties(properties)
+    } else {
```
PawelLipski, line 77: It's possible that we've got some `latestMinorsOfOldSupportedMajors` to be updated in this case as well (e.g. new EAP for 2022.3 comes out, but also 2022.1.5 comes out as well)

buildSrc/src/main/kotlin/com/virtuslab/gitmachete/buildsrc/UpdateIntellijVersions.kt
```diff
@@ -0,0 +1,90 @@
...
+
+  private fun findEapWithBuildNumberHigherThan(latestEapBuildNumber: String): String? {
+    return findMatchingVersionNewerThan(
+      Regex("(?<=ideaIC-)\\d+\\.\\d+\\.\\d+(?=-EAP-SNAPSHOT.pom)"),
+      latestEapBuildNumber
+    )
+  }
+
+  private fun getUpdatedLatestMinorsList(majorOfReplacedStable: String): List<String> {
+    val latestMinorsList = IntellijVersions.latestMinorsOfOldSupportedMajors
```
PawelLipski, line 51: I wouldn't rely on an (effectively) global value from inside a function, this makes it **impure** (since the result doesn't depend just on parameters) and thus harder to reason about.

I'd in fact just pass a list of majors as a param, and return a list of minors.... with no care about what version is stable/what is replaced etc.

buildSrc/src/main/kotlin/com/virtuslab/gitmachete/buildsrc/UpdateIntellijVersions.kt
```diff
@@ -0,0 +1,90 @@
...
+  }
+  private fun findReleaseNewerThan(latestStable: String): String? {
+    return findMatchingVersionNewerThan(Regex("(?<=BUILD-)(\\d+\\.)+\\d+(?=.txt)"), latestStable)
+  }
+
+  private fun findLatestMinorOfMajor(major: String): String {
+    return findMatchingVersionNewerThan(Regex("(?<=BUILD-)$major(\\.\\d+)?(?=.txt)"), major) ?: major
+  }
+
+  private fun findEapWithBuildNumberHigherThan(latestEapBuildNumber: String): String? {
```
PawelLipski, line 43: I wouldn't call the param `latestEap...`, as at the moment of running the method, this might no longer correspond to the latest EAP. In fact this method shouldn't care about what this build number represents:

```suggestion
  private fun findEapWithBuildNumberHigherThan(buildNumber: String): String? {
```

It just looks for EAP with a build number higher than the given threshold, whatever this threshold might be

buildSrc/src/main/kotlin/com/virtuslab/gitmachete/buildsrc/UpdateIntellijVersions.kt
```diff
@@ -0,0 +1,104 @@
...
+  val intellijSnapshotsContents: Elements by lazy {
+    getLinksFromUrl("https://www.jetbrains.com/intellij-repository/snapshots/")
+  }
+
+  private fun getLinksFromUrl(repositoryUrl: String): Elements {
+    return Jsoup.connect(repositoryUrl).get()
+      .select("a[href^=${repositoryUrl}com/jetbrains/intellij/idea/ideaIC/][href$=.pom]")
+  }
+
+  private fun findMatchingVersionNewerThan(repoLinks: Elements, regex: Regex, thresholdVersion: String): String? {
```
PawelLipski, line 29: For clarity:
```suggestion
  private fun findFirstMatchingVersionNewerThan(repoLinks: Elements, regex: Regex, thresholdVersion: String): String? {
```

buildSrc/src/main/kotlin/com/virtuslab/gitmachete/buildsrc/UpdateIntellijVersions.kt
```diff
@@ -0,0 +1,104 @@
...
+
+open class UpdateIntellijVersions : DefaultTask() {
+
+  @get:Internal
+  val intellijReleasesContents: Elements by lazy {
+    getLinksFromUrl("https://www.jetbrains.com/intellij-repository/releases/")
+  }
+
+  @get:Internal
+  val intellijSnapshotsContents: Elements by lazy {
```
PawelLipski, line 20: Just `private val`? (now that `buildSrc/` is excluded from `scripts/prohibit-kotlin-private-fields`)

buildSrc/src/main/kotlin/com/virtuslab/gitmachete/buildsrc/UpdateIntellijVersions.kt
```diff
@@ -0,0 +1,104 @@
...
+      .map { findLatestMinorOfVersion(it) }
+
+    properties.setProperty("latestMinorsOfOldSupportedMajors", latestMinorsList.joinToString(separator = ","))
+
+    val newerStable = findReleaseNewerThan(latestStable)
+
+    if (newerStable != null) {
+      project.logger.lifecycle("latestStable is updated to $newerStable")
+      properties.setProperty("latestStable", newerStable)
+      properties.remove("eapOfLatestSupportedMajor")
```
PawelLipski, line 82: Hmmm one nuance: what if we've got:
```properties
latestStable=2022.2.1
eapOfLatestSupportedMajor=223.1234.56-EAP-SNAPSHOT
```

and then 2022.2.2 comes out? IIUC it'll be picked up as `newerStable`, and hence `eapOfLatestSupportedMajor` will be (incorrectly!) removed :/

buildSrc/src/main/kotlin/com/virtuslab/gitmachete/buildsrc/UpdateIntellijVersions.kt
```diff
@@ -0,0 +1,102 @@
...
+      if (matchResult != null) {
+        val foundVersion = matchResult.value
+
+        if (foundVersion versionIsNewerThan thresholdVersion) {
+          return foundVersion
+        }
+      }
+    }
+    return null
+  }
```
PawelLipski, line 41:
```suggestion
  }

```

buildSrc/src/main/kotlin/com/virtuslab/gitmachete/buildsrc/UpdateIntellijVersions.kt
```diff
@@ -0,0 +1,102 @@
...
+      intellijReleasesContents,
+      Regex("(?<=ideaIC-)$major(\\.\\d+)?(?=.pom)"),
+      major
+    ) ?: versionNumber
+  }
+
+  private fun findEapWithBuildNumberHigherThan(buildNumber: String): String? {
+    return findMatchingVersionNewerThan(
+      intellijSnapshotsContents,
+      Regex("(?<=ideaIC-)\\d+\\.\\d+\\.\\d+(?=-EAP-SNAPSHOT.pom)"),
```
PawelLipski, line 63: Nit: you can consider [raw strings](https://kotlinlang.org/docs/basic-types.html#string-literals) to de-`\`-ify the regexes

CONTRIBUTING.md
```diff
@@ -274,18 +274,21 @@ There is a rather little risk that the plugin which is compatible with `X - 1`
...
+Version updates can be done automatically using `./gradlew updateIntellijVersions`, this runs two update tasks described in the example below.
+
+The whole logic of the process can be illustrated with an example:
 1. our plugin in version `0.7.0` is compatible with IntelliJ `2020.2`
-2. then IntelliJ `2020.3-EAP` is released (see [snapshot repository](https://www.jetbrains.com/intellij-repository/snapshots/) -> Ctrl+F `idea`)
-3. we check if `0.7.0` is compatible with IntelliJ `2020.3-EAP`:
-   set `eapOfLatestSupportedMajor` in [IntellijVersions](buildSrc/src/main/kotlin/com/virtuslab/gitmachete/buildsrc/IntellijVersions.kt)
-   and see if the CI pipeline passes (this will both check binary compatibility and run UI tests against the given EAP)
+2. then IntelliJ `2020.3-EAP` is released (see [snapshot repository](https://www.jetbrains.com/intellij-repository/snapshots/) -> Ctrl+F `.idea`),
+   and is detected with `./gradlew updateEapBuildNumber`, <br/>
```
PawelLipski, line 282: This task no longer exists :/

buildSrc/src/main/kotlin/com/virtuslab/gitmachete/buildsrc/UpdateIntellijVersions.kt
PawelLipski: Also! I wouldn't **remove** the property key, just set it to empty string

CONTRIBUTING.md
```diff
@@ -274,18 +274,21 @@ There is a rather little risk that the plugin which is compatible with `X - 1`
...
+   new `eapOfLatestSupportedMajor` is set in [IntellijVersions](buildSrc/src/main/kotlin/com/virtuslab/gitmachete/buildsrc/IntellijVersions.kt)
+3. we check if `0.7.0` is compatible with IntelliJ `2020.3-EAP` - see if the CI pipeline passes (this will both check binary compatibility and run UI tests against the given EAP)
 4. we release the plugin as `0.8.0` (`untilBuild` will extend automatically to `2020.3.*`
    via `latestSupportedMajor` in [IntellijVersions](buildSrc/src/main/kotlin/com/virtuslab/gitmachete/buildsrc/IntellijVersions.kt))
-5. once the stable `2020.3` is released, we verify ASAP that `0.8.0` is binary compatible with `2020.3` as well;
-   then, `latestStable` can be updated to `2020.3`, `eapOfLatestSupportedMajor` can be set to `null`,
-   and `2020.2.x` can be added to `latestMinorsOfOldSupportedMajors`
-6. since `latestStable` is used as the version to build against,
+5. new stable version `2020.3` is released (see [release repository](https://www.jetbrains.com/intellij-repository/releases/) -> Ctrl+F `.idea`)
+   and is detected using `./gradlew updateIntellijStableVersions`, <br/>
```
PawelLipski, line 288: Ditto

CONTRIBUTING.md
```diff
@@ -274,18 +274,21 @@ There is a rather little risk that the plugin which is compatible with `X - 1`
...
+
+The whole logic of the process can be illustrated with an example:
 1. our plugin in version `0.7.0` is compatible with IntelliJ `2020.2`
-2. then IntelliJ `2020.3-EAP` is released (see [snapshot repository](https://www.jetbrains.com/intellij-repository/snapshots/) -> Ctrl+F `idea`)
-3. we check if `0.7.0` is compatible with IntelliJ `2020.3-EAP`:
-   set `eapOfLatestSupportedMajor` in [IntellijVersions](buildSrc/src/main/kotlin/com/virtuslab/gitmachete/buildsrc/IntellijVersions.kt)
-   and see if the CI pipeline passes (this will both check binary compatibility and run UI tests against the given EAP)
+2. then IntelliJ `2020.3-EAP` is released (see [snapshot repository](https://www.jetbrains.com/intellij-repository/snapshots/) -> Ctrl+F `.idea`),
+   and is detected with `./gradlew updateIntellijVersions`, <br/>
+   new `eapOfLatestSupportedMajor` is set in [IntellijVersions](buildSrc/src/main/kotlin/com/virtuslab/gitmachete/buildsrc/IntellijVersions.kt)
```
PawelLipski, line 283: Pls repoint the link to `intellijVersions.properties`

CONTRIBUTING.md
```diff
@@ -274,18 +274,21 @@ There is a rather little risk that the plugin which is compatible with `X - 1`
...
+3. we check if `0.7.0` is compatible with IntelliJ `2020.3-EAP` - see if the CI pipeline passes (this will both check binary compatibility and run UI tests against the given EAP)
 4. we release the plugin as `0.8.0` (`untilBuild` will extend automatically to `2020.3.*`
    via `latestSupportedMajor` in [IntellijVersions](buildSrc/src/main/kotlin/com/virtuslab/gitmachete/buildsrc/IntellijVersions.kt))
```
PawelLipski, line 286: Ditto

CONTRIBUTING.md
```diff
@@ -274,18 +274,21 @@ There is a rather little risk that the plugin which is compatible with `X - 1`
...
-5. once the stable `2020.3` is released, we verify ASAP that `0.8.0` is binary compatible with `2020.3` as well;
-   then, `latestStable` can be updated to `2020.3`, `eapOfLatestSupportedMajor` can be set to `null`,
-   and `2020.2.x` can be added to `latestMinorsOfOldSupportedMajors`
-6. since `latestStable` is used as the version to build against,
+5. new stable version `2020.3` is released (see [release repository](https://www.jetbrains.com/intellij-repository/releases/) -> Ctrl+F `.idea`)
+   and is detected using `./gradlew updateIntellijVersions`, <br/>
+   `latestStable` and additionally `latestMinorsOfOldSupportedMajors` are changed in [IntellijVersions](buildSrc/src/main/kotlin/com/virtuslab/gitmachete/buildsrc/IntellijVersions.kt)
```
PawelLipski, line 289: And here

buildSrc/src/main/kotlin/com/virtuslab/gitmachete/buildsrc/IntellijVersionHelper.kt
```diff
@@ -41,6 +41,11 @@ object IntellijVersionHelper {
     return version.substring(0, 6)
   }
 
+  fun Properties.getEapOfLatestSupportedMajor(): String? {
+    val maybeEap = getProperty("eapOfLatestSupportedMajor")
+    return if (maybeEap == "") null else maybeEap
+  }
```
PawelLipski, line 47: `getEapOfLatestSupportedMajor` is too case-specific to be added as an extension function to `Property`... I'd rather do sth like:

```suggestion
  fun Properties.getPropertyOrNullIfEmpty(key: String): String? {
    val value = getProperty(key)
    return if (value == "") null else value
  }
```

buildSrc/src/main/kotlin/com/virtuslab/gitmachete/buildsrc/UpdateIntellijVersions.kt
```diff
@@ -0,0 +1,105 @@
...
+      .map { findLatestMinorOfVersion(it) }
+
+    properties.setProperty("latestMinorsOfOldSupportedMajors", latestMinorsList.joinToString(separator = ","))
+
+    val newerStable = findReleaseNewerThan(latestStable)
+
+    if (newerStable != null) {
+      project.logger.lifecycle("latestStable is updated to $newerStable")
+      properties.setProperty("latestStable", newerStable)
+      properties.setProperty("eapOfLatestSupportedMajor", "")
```
PawelLipski, line 83: Actually... wouldn't it make sense to run this line within the body of `if (getMajorPart(latestStable) != getMajorPart(newerStable))`? 🤔 If you're updating `latestStable` to the next minor of the same major, than the existing EAP version doesn't become obsolete 🤔

buildSrc/src/main/kotlin/com/virtuslab/gitmachete/buildsrc/UpdateIntellijVersions.kt
```diff
@@ -0,0 +1,105 @@
...
+
+  private val intellijReleasesContents: Elements by lazy {
+    getLinksFromUrl("https://www.jetbrains.com/intellij-repository/releases/")
+  }
+
+  private val intellijSnapshotsContents: Elements by lazy {
+    getLinksFromUrl("https://www.jetbrains.com/intellij-repository/snapshots/")
+  }
+
+  private fun getLinksFromUrl(repositoryUrl: String): Elements {
```
PawelLipski, line 22: Instead of returning HTML elements here, pls limit the results to just the URLs themselves (since that's what's really relevant); can be just a `List<String>` or whatnot

buildSrc/src/main/kotlin/com/virtuslab/gitmachete/buildsrc/UpdateIntellijVersions.kt
MaciejG604: Good point!

CONTRIBUTING.md
```diff
@@ -274,18 +274,21 @@ There is a rather little risk that the plugin which is compatible with `X - 1`
 and does **not** use any `X EAP/RC`-specific API
 turns out to be **not** compatible with stable `X` release of IDE.
 
-For instance:
+Version updates can be done automatically using `./gradlew updateIntellijVersions`
```
PawelLipski, line 277: Okay! last step that's left, is to wire up the newly-added task correctly in CircleCI config... note that the cron is still going to call the non-existent `updateEapBuildNumber` task :/

.circleci/config.yml
```diff
@@ -217,13 +217,17 @@ jobs:
...
+          command: ./gradlew updateIntellijVersions
       - create_pull_request_if_files_changed:
-          branch_name_prefix: update-eap
-          branch_name_command: &extract_eap grep -Po '(?<=eapOfLatestSupportedMajor=).*(?=-EAP-SNAPSHOT)' intellijVersions.properties
-          commit_message_prefix: Update EAP build number to
-          commit_message_command: *extract_eap
+          branch_name_prefix: update-intellij-versions
+          branch_name_command: md5sum intellijVersions.properties | cut -d' ' -f 1
+          commit_message_prefix: Update IntelliJ versions ->
+          commit_message_command: git diff | grep -P '\+' |
```
PawelLipski, line 226: Hmmm I suppose you'd like to see the lines **starting** with a plus, right?
```suggestion
          commit_message_command: git diff | grep '^+' |
```
Although it won't make much difference anyway, there shouldn't be any pluses other than the initial one anyway

Also, `-P` aka `--perl-regexp` is redundant, you can just use regular syntax, where `+` is a character like any other (unlike with _extended_ syntax, `grep -E`, where `+` stands for "1 or more occurrence")

.circleci/config.yml
```diff
@@ -217,13 +217,17 @@ jobs:
...
+            sed 's/Majors=.*/Majors/' |
+            sed 's/+//g' |
+            sed 's/=/ to /g' |
+            sed 's/to $/to EMPTY/g'
```
PawelLipski, line 230: Okay... how will a sample commit message look like? AFAICS it'd be useful to get rid of newlines (`tr '\n' ','`?)

MaciejG604: Should look like this now (all in one line):
`Update IntelliJ versions -> latestMinorsOfOldSupportedMajors, eapOfLatestSupportedMajor to EMPTY, latestStable to 2022.2`

PawelLipski (approved):

Issue #1008, opened 2022-07-28
Request URL is too long from `Report and Clear All` error


mkondratek:
[...]

PR #1048, opened 2022-08-10
Add test for GitMacheteErrorReportSubmitter

MaciejG604:


frontend/ui/impl/src/test/java/com/virtuslab/gitmachete/frontend/ui/impl/GitMacheteErrorReportSubmitterTest.java
```diff
@@ -0,0 +1,42 @@
...
+    events[0] = new IdeaReportingEvent(messageObject, /* message */ null, /* stacktrace */"", /* plugin */null);
+
+    GitMacheteErrorReportSubmitter reportSubmitter = PowerMockito.spy(new GitMacheteErrorReportSubmitter());
+
+    PowerMockito
+        .stub(PowerMockito.method(GitMacheteErrorReportSubmitter.class, "getTemplateVariables", IdeaLoggingEvent[].class,
+            String.class))
+        .toReturn(new HashMap<String, String>());
+
+    String reportBody = Whitebox.invokeMethod(reportSubmitter, "getReportBody", events, /* additionalInfo */ null);
```
PawelLipski, line 36: Hmmm back to the original issue #972... will this test detect a typo in resource name?

frontend/ui/impl/src/test/java/com/virtuslab/gitmachete/frontend/ui/impl/GitMacheteErrorReportSubmitterTest.java
```diff
@@ -0,0 +1,42 @@
...
+
+    Whitebox.invokeMethod(reportSubmitter, "constructNewGitHubIssueUri", events, reportBody);
+
+    Whitebox.invokeMethod(reportSubmitter, "constructNewGitHubIssueUri", events, null);
```
PawelLipski, line 40: Hmmm the result isn't checked against the expected value? 🤔 as for now, it's basically only checked whether a runtime exception happens

frontend/ui/impl/src/test/java/com/virtuslab/gitmachete/frontend/ui/impl/GitMacheteErrorReportSubmitterTest.java
```diff
@@ -0,0 +1,42 @@
...
+import com.intellij.openapi.diagnostic.IdeaLoggingEvent;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.powermock.api.mockito.PowerMockito;
+import org.powermock.core.classloader.annotations.PrepareForTest;
+import org.powermock.modules.junit4.PowerMockRunner;
+import org.powermock.reflect.Whitebox;
+
+@RunWith(PowerMockRunner.class)
+@PrepareForTest({GitMacheteErrorReportSubmitter.class})
```
PawelLipski, line 19: Hmm what happens without that preparation? I've already forgotten the inner working of Powermock ☹️

MaciejG604: I believe that is to state that you will be changing function calls made by this class. [PowerMock Wiki](https://github.com/powermock/powermock/wiki/Mock-System#example)

frontend/ui/impl/src/test/java/com/virtuslab/gitmachete/frontend/ui/impl/GitMacheteErrorReportSubmitterTest.java
MaciejG604: Now it will detect a typo in resource's name.

frontend/ui/impl/src/test/java/com/virtuslab/gitmachete/frontend/ui/impl/GitMacheteErrorReportSubmitterTest.java
MaciejG604: Well, we could test if it matches the URI we expect, but then we would need to hard-code the expected URI, or create it using same logic as in the tested class. So is there a point?

frontend/ui/impl/src/test/java/com/virtuslab/gitmachete/frontend/ui/impl/GitMacheteErrorReportSubmitterTest.java
```diff
@@ -0,0 +1,49 @@
...
+public class GitMacheteErrorReportSubmitterTest {
+  @Test
+  public void shouldNotThrowWhenGettingBugTemplate() throws Exception {
+    GitMacheteErrorReportSubmitter reportSubmitter = new GitMacheteErrorReportSubmitter();
+
+    Whitebox.invokeMethod(reportSubmitter, "getBugTemplate");
```
PawelLipski, line 25: Hmm for clarity... how would that differ from a direct `reportSubmitter.getBugTemplate()`?

frontend/ui/impl/src/test/java/com/virtuslab/gitmachete/frontend/ui/impl/GitMacheteErrorReportSubmitterTest.java
PawelLipski: Hardcode the expected URI then, definitely. It can be even be stored in module's `src/test/resources`.

This enitire class is quite important (since it allows users to well, report errors in the plugin smoothly), it's vital that we spot regressions as soon as possible.

frontend/ui/impl/src/test/java/com/virtuslab/gitmachete/frontend/ui/impl/GitMacheteErrorReportSubmitterTest.java
```diff
@@ -0,0 +1,49 @@
...
+
+  @Test
+  public void testGitHubIssueUriConstruction() throws Exception {
+    IdeaLoggingEvent[] events = new IdeaLoggingEvent[1];
+
+    AbstractMessage messageObject = Whitebox.newInstance(LogMessage.class);
+
+    events[0] = new IdeaReportingEvent(messageObject, /* message */ null, /* stacktrace */"", /* plugin */null);
+
+    GitMacheteErrorReportSubmitter reportSubmitter = PowerMockito.spy(new GitMacheteErrorReportSubmitter());
```
PawelLipski, line 36: Is `spy` needed here? AFAICS all other tests use `mock(...)`

frontend/ui/impl/src/test/java/com/virtuslab/gitmachete/frontend/ui/impl/GitMacheteErrorReportSubmitterTest.java
PawelLipski: Also, this'll be be pretty important to provide a regression test for #1008

frontend/ui/impl/src/test/java/com/virtuslab/gitmachete/frontend/ui/impl/GitMacheteErrorReportSubmitterTest.java
PawelLipski: Oh okay, private method

frontend/ui/impl/src/test/java/com/virtuslab/gitmachete/frontend/ui/impl/GitMacheteErrorReportSubmitterTest.java
```diff
@@ -0,0 +1,49 @@
...
+import org.powermock.api.mockito.PowerMockito;
+import org.powermock.core.classloader.annotations.PrepareForTest;
+import org.powermock.modules.junit4.PowerMockRunner;
+import org.powermock.reflect.Whitebox;
+
+@RunWith(PowerMockRunner.class)
+@PrepareForTest({GitMacheteErrorReportSubmitter.class})
+public class GitMacheteErrorReportSubmitterTest {
+  @Test
+  public void shouldNotThrowWhenGettingBugTemplate() throws Exception {
```
PawelLipski, line 22: Actually... isn't this method already called when `getReportBody` is called in `testGitHubIssueUriConstruction`? 🤔

PawelLipski: Which would make this test probably redundant (its existence doesn't increase coverage)

frontend/ui/impl/src/test/java/com/virtuslab/gitmachete/frontend/ui/impl/GitMacheteErrorReportSubmitterTest.java
MaciejG604: With spy you can to partial mocking [Mockito API](https://javadoc.io/static/org.mockito/mockito-core/4.6.1/org/mockito/Mockito.html#spy), mock does not let you call real methods.

PawelLipski: Oh, so that's basically needed to call private methods on `GitMacheteErrorReportSubmitter`? 🤔 As in, assuming all tested methods of `GitMacheteErrorReportSubmitter` were public, neither `spy()` nor `mock()` would be needed?

PawelLipski: Wow okay... pls add a comment about the usage of `spy` into the code then!

MaciejG604: The link in my first answer was incorrect, sorry -> [Mockito API](https://javadoc.io/static/org.mockito/mockito-core/4.6.1/org/mockito/Mockito.html#spy)

frontend/ui/impl/src/test/java/com/virtuslab/gitmachete/frontend/ui/impl/GitMacheteErrorReportSubmitterTest.java
```diff
@@ -0,0 +1,76 @@
...
+@RunWith(PowerMockRunner.class)
+@PrepareForTest({GitMacheteErrorReportSubmitter.class})
+public class GitMacheteErrorReportSubmitterTest {
+
+  private String defaultErrorReport;
+  private String defaultErrorReportWithTemplate;
+
+  @Before
+  public void setUp() {
+    {
```
PawelLipski, line 36: Redundant braces

frontend/ui/impl/src/test/java/com/virtuslab/gitmachete/frontend/ui/impl/GitMacheteErrorReportSubmitterTest.java
```diff
@@ -0,0 +1,76 @@
...
+      try {
+        defaultErrorReport = IOUtils.resourceToString("/error_report_uri.txt", StandardCharsets.UTF_8).strip();
+        defaultErrorReportWithTemplate = IOUtils.resourceToString("/error_report_uri_with_template.txt",
+            StandardCharsets.UTF_8).strip();
+      } catch (IOException e) {
+        throw new RuntimeException(e);
```
PawelLipski, line 42: Just `@SneakyThrows` from Lombok

frontend/ui/impl/src/test/java/com/virtuslab/gitmachete/frontend/ui/impl/GitMacheteErrorReportSubmitterTest.java
```diff
@@ -0,0 +1,76 @@
...
+import org.powermock.api.mockito.PowerMockito;
+import org.powermock.core.classloader.annotations.PrepareForTest;
+import org.powermock.modules.junit4.PowerMockRunner;
+import org.powermock.reflect.Whitebox;
+
+@RunWith(PowerMockRunner.class)
+@PrepareForTest({GitMacheteErrorReportSubmitter.class})
+public class GitMacheteErrorReportSubmitterTest {
+
+  private String defaultErrorReport;
```
PawelLipski, line 31: Nit: could be static, use a `@BeforeClass` method

frontend/ui/impl/src/test/resources/error_report_uri.txt
```diff
@@ -0,0 +1 @@
+https://github.com/VirtusLab/git-machete-intellij-plugin/issues/new?title=&labels=bug&template=bug_report.md
```
PawelLipski, line 1: Pls prepare the test cases so that the `title` is non-empty

frontend/ui/impl/src/test/resources/error_report_uri_with_template.txt
```diff
@@ -0,0 +1 @@
...nvironment%0A-+Git+Machete+plugin+version+-+%25macheteVersion%25%0A-+IDE+-+%25ide%25%0A-+Operating+system+-+%25os%25%0A%0A%23%23+IDEA+-+Stack+trace%0A%25stacktraces%25%0A%0A%23%23+IDEA+-+additional+info%0A%25additionalInfo%25%0A%0A%23%23+Bug+description%0A**Please+include+steps+to+reproduce+%28like+%60go+to...%60%2F%60click+on...%60+etc.%29+%2B+expected+and+actual+behaviour**.%0A%0A%23%23+%28optional%29+Logs%0AEnable+logging+on+DEBUG+level+for+this+plugin+by+adding%0A%60%60%60%0Acom.virtuslab%0A%60%60%60%0Ato+the+list+under+%60Help%60+-%3E+%60Diagnostic+Tools%60+-%3E+%60Debug+Log+Settings%60.%3Cbr%2F%3E%0AThen+reproduce+a+bug+and+attach+the+logs+to+the+issue.%3Cbr%2F%3E%0ATo+find+IntelliJ+log+file%2C+go+to+%60Help%60+-%3E+%60Show+Log+in+Files%60.%0A%0AConsider+placing+the+logs+within+the+%60details%60+%28aka+%22spoiler%22%29+tags+as+the+may+be+very+extensive.%0A%0A%3Cdetails%3E%0A%3Csummary%3ELogs%3C%2Fsummary%3E%0ALogs+go+here%0A%3C%2Fdetails%3E%0A%0A%23%23+%28optional%29+Screenshots%0AIf+applicable%2C+add+screenshots+%28or+screen+recordings%2C+see+%5BPeek%5D%28https%3A%2F%2Fgithub.com%2Fphw%2Fpeek%23peek---an-animated-gif-recorder%29+on+Linux%29%0Ato+help+explain+your+problem.%0A
```
PawelLipski, line 1: ...and so that a stack trace is included in the `body`

frontend/ui/impl/src/test/java/com/virtuslab/gitmachete/frontend/ui/impl/GitMacheteErrorReportSubmitterTest.java
```diff
@@ -0,0 +1,76 @@
...
+    val events = new IdeaLoggingEvent[1];
+
+    val messageObject = Whitebox.newInstance(LogMessage.class);
+
+    events[0] = new IdeaReportingEvent(messageObject, /* message */ null, /* stacktrace */"", /* plugin */null);
+
+    val reportSubmitter = PowerMockito.spy(new GitMacheteErrorReportSubmitter());
+
+    PowerMockito
+        .stub(PowerMockito.method(GitMacheteErrorReportSubmitter.class, "getTemplateVariables", IdeaLoggingEvent[].class,
```
PawelLipski, line 58: How about just mocking e.g. `ApplicationInfo.getInstance().getFullApplicationName()` and `PluginManagerCore.getPlugin(...)` instead of full method of a tested object? would it be easily doable?

PawelLipski: Generally we should avoid mocking methods of the class that we're testing... it should be the **external world** that might be mocked instead (in our case, the IntelliJ-provided environment like `ApplicationInfo`)

MaciejG604: I thought about it, could be tricky but I will try it.

PawelLipski: Yeah, pls check... or at least limit the scope of mocked methods to as narrow as possible, sth like:

```java
private String getIdeApplicationName() {
  return ApplicationInfo.getInstance().getFullApplicationName();
}
```

so that major methods like `getTemplateVariables` remain **un**mocked (and hence truly covered with tests)

frontend/ui/impl/src/test/java/com/virtuslab/gitmachete/frontend/ui/impl/GitMacheteErrorReportSubmitterTest.java
```diff
@@ -28,49 +27,99 @@
...
+  @BeforeClass
+  @SneakyThrows
+  public static void setUp() {
+    defaultErrorReport = IOUtils.resourceToString("/error_report_uri.txt", StandardCharsets.UTF_8).strip();
+    defaultErrorReportWithTemplate = IOUtils.resourceToString("/error_report_uri_with_template.txt",
+        StandardCharsets.UTF_8).strip();
+    errorReportWithStackTraceAndTitle = IOUtils.resourceToString("/error_report_uri_with_stacktrace_title.txt",
+        StandardCharsets.UTF_8).strip();
+
+    Whitebox.setInternalState(GitMacheteErrorReportSubmitter.class, errorReportSubmitterLogger);
```
PawelLipski, line 48: Wow nice... so it matches the value to the correct field (`LOG` in this case) automagically? 😯

frontend/ui/impl/src/test/java/com/virtuslab/gitmachete/frontend/ui/impl/GitMacheteErrorReportSubmitterTest.java
```diff
@@ -0,0 +1,125 @@
...
+  private GitMacheteErrorReportSubmitter getMockReportSubmitter(HashMap<String, String> templateVariables) {
+    PowerMockito
+        .stub(PowerMockito.method(GitMacheteErrorReportSubmitter.class, "getTemplateVariables", IdeaLoggingEvent[].class,
+            String.class))
+        .toReturn(templateVariables);
+
+    return gitMacheteErrorReportSubmitter;
+  }
+
+  private IdeaLoggingEvent[] getMockEvents(String throwableText) {
```
PawelLipski, line 64: Hmmm would it make sense to pass an actual `Throwable` here, so that the stack trace is actually URL-encoded into the resulting URI? the problem might be in the **reproducibility** of this stack trace (if it's always the same regardless of the underlying machine/JDK) tho, up to your judgement

frontend/ui/impl/src/test/java/com/virtuslab/gitmachete/frontend/ui/impl/GitMacheteErrorReportSubmitterTest.java
```diff
@@ -0,0 +1,123 @@
...
+            never().description("Reading the resource file with the bug template should NOT throw exceptions"))
+        .error(anyString(), (Throwable) any());
+
+    return reportBody;
+  }
+
+  @Test
+  public void shouldConstructUriWithTemplate() throws Exception {
+    val events = getMockEvents(/* throwableText */ "");
+    val reportBody = invokeGetReportBody(reportSubmitter, events);
```
PawelLipski, line 99: One more thing, to simplify the tests even more: what if you reworked `constructNewGitHubIssueUri` to sth like:
```java
  private URI constructNewGitHubIssueUri(IdeaLoggingEvent[] events, @Nullable String additionalInfo) throws URISyntaxException {
    String reportBody = getReportBody(events, additionalInfo);
    ....
```

Then AFAICS all the engineering around `getReportBody` won't even be needed in the test

frontend/ui/impl/src/test/java/com/virtuslab/gitmachete/frontend/ui/impl/GitMacheteErrorReportSubmitterTest.java
```diff
@@ -0,0 +1,123 @@
...
+
+    URI uriWithReport = Whitebox.invokeMethod(reportSubmitter, "constructNewGitHubIssueUri", events, reportBody);
+    Assert.assertEquals("URI should match default with report body", defaultErrorReportWithTemplate, uriWithReport.toString());
+  }
+
+  @Test
+  public void shouldConstructUriWithNoTemplate() throws Exception {
+    val events = getMockEvents(/* throwableText */ "");
+
+    URI uriNoReport = Whitebox.invokeMethod(reportSubmitter, "constructNewGitHubIssueUri", events, /* report Body */ null);
```
PawelLipski, line 109: Another simplification: once you make `constructNewGitHubIssueUri` package-private, it seems that the test still passes with:
```suggestion
    URI uriNoReport = reportSubmitter.constructNewGitHubIssueUri(events, /* report Body */ null);
```

frontend/ui/impl/src/test/java/com/virtuslab/gitmachete/frontend/ui/impl/GitMacheteErrorReportSubmitterTest.java
MaciejG604: We still have to make sure no exception get's thrown inside getReportBody(), so most of the engineering with the logger will stay the same.

frontend/ui/impl/src/test/java/com/virtuslab/gitmachete/frontend/ui/impl/GitMacheteErrorReportSubmitterTest.java
MaciejG604: Sure it will pass, since Whitebox just lets us invoke the private method. I'd leave it as it is and keep the encapsulation.

frontend/ui/impl/src/test/java/com/virtuslab/gitmachete/frontend/ui/impl/GitMacheteErrorReportSubmitterTest.java
MaciejG604: I tried doing so, but that's a huge pain in the neck when refactoring as the throwable remembers on which line it was created and so the URI will change everytime somebody edits the code/adds an import. Apart from that, encoding the `Throwable` into the URI is just String operations so I wouldn't expect any problems with that other than #1008

PawelLipski (approved):

frontend/ui/impl/src/test/java/com/virtuslab/gitmachete/frontend/ui/impl/GitMacheteErrorReportSubmitterTest.java
PawelLipski: Hmmm won't the test fail anyway if an exception is thrown inside `getReportBody`? 🤔

MaciejG604: No, as it's caught inside the method, LOG.error(...) is called inside the catch block to inform the outside world about the exception.

PR #1051, opened 2022-08-11
Make codebase build successfully under JDK 17

PawelLipski:


LukaszKontowski (changes requested):
Please, update docs as well. Tell the people that they should use 11 or 17. At least the https://github.com/VirtusLab/git-machete-intellij-plugin/blob/master/CONTRIBUTING.md file needs update: (maybe somewhere else too?)
"For running ./gradlew from the command line, make sure that java and javac are in PATH and point to Java 11."

LukaszKontowski (approved):

PR #1059, opened 2022-08-17
Remove branch word from action ids and action class names

mkondratek:


frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/backgroundables/GitCommandUpdatingCurrentBackgroundable.java
```diff
@@ -54,12 +54,12 @@
...
 @ExtensionMethod(GitMacheteBundle.class)
 @CustomLog
-public abstract class GitCommandUpdatingCurrentBranchBackgroundable extends Task.Backgroundable {
+public abstract class GitCommandUpdatingCurrentBackgroundable extends Task.Backgroundable {
 
   protected final Project project;
   protected final GitRepository gitRepository;
 
-  public GitCommandUpdatingCurrentBranchBackgroundable(
+  public GitCommandUpdatingCurrentBackgroundable(
```
PawelLipski, line 62: Huh I'd consider leaving it here... as kinda exception (or we can assume the rule that `Branch` is only implied in **action class** names... even to be documented in CONTRIBUTING.md?)

src/uiTest/resources/project.rhino.js
```diff
@@ -380,7 +380,7 @@ function Project(underlyingProject) {
   };
 
   this.checkoutBranch = function (branchName) {
-    invokeActionAndWait('GitMachete.CheckoutSelectedBranchAction', ACTION_PLACE_CONTEXT_MENU, { SELECTED_BRANCH_NAME: branchName });
+    invokeActionAndWait('GitMachete.CheckoutSelectedAction', ACTION_PLACE_CONTEXT_MENU, { SELECTED_BRANCH_NAME: branchName });
```
PawelLipski, line 383: Wow so that's the only action that's explicitly called from UI tests 😯

mkondratek: yup, and I think it's reasonable. we could use context menu action instead but this method is called so frequently (within the tests) so I think it would significantly extend the time of the tests run

PawelLipski (approved):

PR #1062, opened 2022-08-18
Simplify the Dockerfile used for CI

PawelLipski:


docker/Dockerfile
```diff
@@ -31,32 +19,28 @@ RUN \
 # But in our setup, it's sorted out by ide-probe instead (org.virtuslab.ideprobe.ide.intellij.IntellijPrivacyPolicy).
 
 # Tools necessary to run non-headless UI tests in the screen-less environment of CI
-RUN \
-  apt-get update; \
-  apt-get install --no-install-recommends -y libxtst6 xvfb xauth x11-apps imagemagick zip; \
-  `# libraries needed by JCEF` \
```
PawelLipski, line 37: @LukaszKontowski as discussed, I've dropped support for [Java Chromium Embedded Framework](https://github.com/chromiumembedded/java-cef)

PawelLipski: (as we already for some time no longer use a JetBrains Runtime with JCEF for UI tests)

docker/Dockerfile
```diff
@@ -31,32 +19,26 @@ RUN \
...
+RUN apt-get install --no-install-recommends -y binutils jq sudo unzip xxd
 
 # Markdown validation utilities
 RUN \
-  npm install --location=global remark-cli remark-lint-no-dead-urls remark-validate-links; \
-  npm uninstall --location=global npm
+  curl -sL https://deb.nodesource.com/setup_18.x | bash -; \
+  apt-get install -y nodejs; \
+  npm install --global remark-cli remark-lint-no-dead-urls remark-validate-links; \
+  remark --help
```
mkondratek, line 41: we are calling a `remark --help` command here, am I right? for what reason is that?

mkondratek (approved):
two (probably) minor comments. nice code removal 👍

docker/Dockerfile
```diff
@@ -1,25 +1,13 @@
 
-# Not using openjdk:11-jdk-slim-... due to https://github.com/AdoptOpenJDK/openjdk-docker/issues/75
-FROM debian:11-slim
+# Use JDK and not just JRE to provide `javadoc` executable
+# Jammy = Ubuntu 22.04 LTS
+FROM eclipse-temurin:11-jdk-jammy
```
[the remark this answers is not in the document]
PawelLipski: Mostly cons... as long as we support **2022.1 or earlier**, I want to make sure that everything we publish is 11-compatible... and the best way to ensure that is to build on JDK 11 and not later 😅

docker/Dockerfile
PawelLipski: Oh! I had a hard time getting this section to work... mostly using this call to `remark` as an instant "unit test" whether it installed successfully ;) I'd leave it here for good, for that purpose

PR #1067, opened 2022-08-19
Restore JCEF in CI Docker image

PawelLipski:


docker/Dockerfile
```diff
@@ -40,6 +40,12 @@ RUN \
   npm install --global remark-cli remark-lint-no-dead-urls remark-validate-links; \
   remark --help
 
+# TODO (#825): remove this `RUN`
+# Required for UI tests of 2021.1.
```
PawelLipski, line 44: Weird, no earlier nor later version requires that

mkondratek (approved):

PR #1073, opened 2022-08-21
Pin slf4j-simple at 1.x.y + add project properties to control test behavior + rectify CONTRIBUTING.md

PawelLipski:


CONTRIBUTING.md
```diff
@@ -156,18 +166,6 @@ docker run --rm -e UID=$(id -u) -e GID=$(id -g) \
 ```
 
 
-
-## Update version catalog
```
PawelLipski, line 160: No need to document this explicitly, this is no longer meant to be run by a human 🤖

gradle/libs.versions.toml
```diff
@@ -26,6 +26,12 @@ powerMock-moduleJunit4 = { module = "org.powermock:powermock-module-junit4", ver
 reflections = "org.reflections:reflections:0.10.2"
 scala-library = "org.scala-lang:scala-library:2.13.8"
 slf4j-lambda = "kr.pe.kwonnam.slf4j-lambda:slf4j-lambda-core:0.1"
+# @pin because slf4j-simple 2.0.0 isn't compatible with slf4j-api v1.x.y, which is still pulled in by our dependencies.
```
PawelLipski, line 29: Probably one of the longest-awaited releases in the JVM ecosystem, ever 😅
Basically:
* adds lambda API: `log.info(() -> "lolxd")` (we still can't use it tho, as long as IntelliJ itself isn't migrated to slf4j v2.0.0)
* adds fluent API: `log.atInfo().log("lolxd")`
* changes detection of implementation from the somewhat primitive-yet-effective static binding, to service provider (that's a breaking change)

CONTRIBUTING.md
```diff
@@ -217,16 +215,16 @@ Other coding conventions include:
...
+  Some exceptions are allowed (e.g. the backgroundable task classes).
 
 ## UI conventions
 
 So far created UI conventions:
 
 * Add `…` (ellipsis, `\u2026`)  at the end of an action name if it is not executed immediately after clicking e.g. `Sync to Parent by Rebase…` (after this operation the interactive rebase window opens)
 * Toolbar name texts of a **toolbar** actions that refer to a branch should indicate the branch under action with the word `Current`.
-  On the other hand, **context-menu** actions text names should be kept short (no `This`/`Selected`).
-* Avoid `Branch` word in action class names and action ids to keep them shorter.
```
mkondratek, line 228: y removed?

CONTRIBUTING.md
mkondratek: what will we do tomorrow then? 😨

CONTRIBUTING.md
```diff
@@ -67,7 +67,7 @@ This is because one may not be emulating bash environment in any way or doing it
 #### macOS
```
mkondratek, line 67: u added an empty line after `####` section below but here not. be consistent

CONTRIBUTING.md
mkondratek: oh, I see now (section above). good idea!

mkondratek (approved):

CONTRIBUTING.md
PawelLipski: Idk... maybe _bombelki_? 👶🏻👶🏻👶🏻

CONTRIBUTING.md
PawelLipski: Ha! nope, it's just that at the end of each `##` group there are two empty lines

PR #1078, opened 2022-08-22
Provide coverage in UI tests

PawelLipski:


src/uiTest/resources/ideprobe.conf
```diff
@@ -5,7 +5,10 @@ probe {
   }
 
   driver {
-    vmOptions = ["-Xmx1G"]
+    vmOptions = [
+      "-Xmx1G",
+      "-javaagent:/home/plipski/git-machete-intellij-plugin/build/tmp/expandedArchives/org.jacoco.agent-0.8.8.jar_a33b649e552c51298e5a242c2f0d0e3c/jacocoagent.jar=destfile=/home/plipski/git-machete-intellij-plugin/build/jacoco/uiTest.exec,append=true,inclnolocationclasses=false,dumponexit=true,output=file,jmx=false"
```
PawelLipski, line 10: I'm using path of some temporary file that showed up after running `jacocoTestReport`, this will ofc differ depending on machine/jacoco version.

The agent runs correctly... but the problem is that coverage data isn't emitted anyway.

My suspicion is that since plugin is dynamically loaded, the agent doesn't run on plugin's classes (?). TBD

PR #1085, opened 2022-08-25
Add a note about plugin versions that remain available after bumping earliestSupportedMajor (closes #1084)

### LINKED ISSUE #1084: Once a version which does NOT support IntelliJ X.Y is released, is it still possible to install an old version in IntelliJ X.Y?
In particular, is it still possible to install 1.x.y version of this plugin in e.g. IntelliJ 2019.3? 

This question is pretty important wrt. publishing 3.0.0: if after publishing 3.0.0 (which is only going to support IntelliJs 2022.1 onwards) it's still possible to install 2.x.y on 2021.3 and earlier, then maybe we shouldn't postpone the release of 3.0.0 so much 🤔

mkondratek:


buildSrc/src/main/kotlin/com/virtuslab/gitmachete/buildsrc/IntellijVersions.kt
```diff
@@ -10,6 +10,11 @@ object IntellijVersions {
   // When this value is updated, remember to update:
   // 1. the minimum required IDEA version in README.md,
   // 2. version of Gradle Kotlin plugin in gradle/libs.versions.toml
+  // Note that after bumping X.Y to A.B (A.B is later) the released plugin versions supporting X.Y remain available.
```
PawelLipski, line 13:
```suggestion
  // Note that after bumping A.B to C.D (C.D is later) the released plugin versions supporting A.B remain available in JetBrains Marketplace.
```

PawelLipski (approved):

buildSrc/src/main/kotlin/com/virtuslab/gitmachete/buildsrc/IntellijVersions.kt
```diff
@@ -10,7 +10,8 @@ object IntellijVersions {
   // When this value is updated, remember to update:
   // 1. the minimum required IDEA version in README.md,
   // 2. version of Gradle Kotlin plugin in gradle/libs.versions.toml
-  // Note that after bumping X.Y to A.B (A.B is later) the released plugin versions supporting X.Y remain available.
+  // Note that after bumping A.B to C.D (C.D is later)
```
PawelLipski, line 13:
```suggestion
  // Note that after bumping `earliestSupportedMajor` from A.B to C.D (C.D is later)
```

buildSrc/src/main/kotlin/com/virtuslab/gitmachete/buildsrc/IntellijVersions.kt
```diff
@@ -10,7 +10,8 @@ object IntellijVersions {
...
+  // the released plugin versions supporting A.B remain available in JetBrains Marketplace.
   // Dropping a support for an intellij version is less painful then,
```
PawelLipski, line 15:
```suggestion
  // Dropping a support for an IntelliJ version is less painful then,
```

PR #1086, opened 2022-08-25
Replace `ui-test.intellij.version` and `ui-test.plugin.path` with env vars

PawelLipski:
~Blocked on https://github.com/VirtusLab/ide-probe/issues/265~

### DIFF
--- a/buildSrc/src/main/kotlin/com/virtuslab/gitmachete/buildsrc/Dependencies.kt
+++ b/buildSrc/src/main/kotlin/com/virtuslab/gitmachete/buildsrc/Dependencies.kt
@@ -14,11 +14,12 @@ import org.gradle.kotlin.dsl.*
 import org.jetbrains.intellij.IntelliJPlugin
 import org.jetbrains.intellij.IntelliJPluginExtension
 import org.jetbrains.intellij.tasks.ClasspathIndexCleanupTask
+import org.jetbrains.kotlin.gradle.plugin.KotlinPluginWrapper
 import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 import java.net.URI
 
 fun Project.applyKotlinConfig() {
-  apply(plugin = "org.jetbrains.kotlin.jvm")
+  apply<KotlinPluginWrapper>()
 
   tasks.withType<KotlinCompile> {
     kotlinOptions {
@@ -105,12 +106,6 @@ fun Project.ideProbe() {
   dependencies {
     "uiTestImplementation"(testFixtures(project(":testCommon")))
     "uiTestImplementation"(bundle("ideProbe"))
-
-    // This is technically redundant (both since ide-probe pulls in scala-library anyway,
-    // and since ide-probe is meant to use in src/uiTest code, not src/test code),
-    // but apparently needed for IntelliJ to detect Scala SDK version in the project (it's
-    // probably https://youtrack.jetbrains.com/issue/SCL-14310).
-    "testImplementation"(lib("scala.library"))
   }
 }
 
--- a/buildSrc/src/main/kotlin/com/virtuslab/gitmachete/buildsrc/UiTests.kt
+++ b/buildSrc/src/main/kotlin/com/virtuslab/gitmachete/buildsrc/UiTests.kt
@@ -28,11 +28,10 @@ fun Project.configureUiTests() {
       testClassesDirs = uiTest.output.classesDirs
       classpath = configurations["uiTestRuntimeClasspath"] + uiTest.output
 
-      dependsOn(":buildPlugin")
       val buildPlugin = tasks.findByPath(":buildPlugin")!!
-
-      systemProperty("ui-test.intellij.version", version)
-      systemProperty("ui-test.plugin.path", buildPlugin.outputs.files.first().path)
+      dependsOn(buildPlugin)
+      environment("IDEPROBE_INTELLIJ_PLUGIN_URI", buildPlugin.outputs.files.first().path)
+      environment("IDEPROBE_INTELLIJ_VERSION_BUILD", version)
 
       // TODO (#945): caching of UI test results doesn't work in the CI anyway
       if (!isCI) {
--- a/frontend/ui/impl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/EnhancedGraphTable.java
+++ b/frontend/ui/impl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/EnhancedGraphTable.java
@@ -167,7 +167,6 @@ public final class EnhancedGraphTable extends BaseEnhancedGraphTable
     gitRepositorySelectionProvider.addSelectionChangeObserver(() -> queueRepositoryUpdateAndModelRefresh());
   }
 
-  // TODO (#620): rework callbacks into futures
   @UIEffect
   private void refreshModel(
       GitRepository gitRepository,
--- a/src/uiTest/resources/ide.rhino.js
+++ b/src/uiTest/resources/ide.rhino.js
@@ -27,10 +27,6 @@ function Ide() {
     }
   };
 
-  this.getMajorVersion = function() {
-    return ApplicationInfo.getInstance().getMajorVersion();
-  };
-
   this.soleOpenedProject = function () {
     const openProjects = ProjectUtil.getOpenProjects();
     return openProjects.length === 1 ? new Project(openProjects[0]) : null;
--- a/src/uiTest/resources/ideprobe.conf
+++ b/src/uiTest/resources/ideprobe.conf
@@ -27,10 +27,13 @@ probe {
       { uri = ${IDEPROBE_INTELLIJ_PLUGIN_URI} }
     ]
 
-    # We're cheating here a bit since IDEPROBE_INTELLIJ_VERSION_BUILD might be either a build number or a release number,
-    # while we're always treating it as a build number.
-    # Still, as of ide-probe 0.42.0, even when release number like `2020.3` is passed as `version.build`, UI tests work just fine.
-    version.build = ${IDEPROBE_INTELLIJ_VERSION_BUILD}
+    version {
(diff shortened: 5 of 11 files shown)


buildSrc/src/main/kotlin/com/virtuslab/gitmachete/buildsrc/Dependencies.kt
```diff
@@ -106,12 +107,6 @@ fun Project.ideProbe() {
   dependencies {
     "uiTestImplementation"(testFixtures(project(":testCommon")))
     "uiTestImplementation"(bundle("ideProbe"))
-
-    // This is technically redundant (both since ide-probe pulls in scala-library anyway,
-    // and since ide-probe is meant to use in src/uiTest code, not src/test code),
-    // but apparently needed for IntelliJ to detect Scala SDK version in the project (it's
```
PawelLipski, line 112: Meh it doesn't help, let's remove

buildSrc/src/main/kotlin/com/virtuslab/gitmachete/buildsrc/Dependencies.kt
```diff
@@ -14,11 +14,12 @@ import org.gradle.kotlin.dsl.*
 import org.jetbrains.intellij.IntelliJPlugin
 import org.jetbrains.intellij.IntelliJPluginExtension
 import org.jetbrains.intellij.tasks.ClasspathIndexCleanupTask
+import org.jetbrains.kotlin.gradle.plugin.KotlinPluginWrapper
 import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 import java.net.URI
 
 fun Project.applyKotlinConfig() {
-  apply(plugin = "org.jetbrains.kotlin.jvm")
+  apply<KotlinPluginWrapper>()
```
PawelLipski, line 22: Unrel, for consistency

buildSrc/src/main/kotlin/com/virtuslab/gitmachete/buildsrc/UiTests.kt
```diff
@@ -28,11 +28,10 @@ fun Project.configureUiTests() {
       testClassesDirs = uiTest.output.classesDirs
       classpath = configurations["uiTestRuntimeClasspath"] + uiTest.output
 
-      dependsOn(":buildPlugin")
       val buildPlugin = tasks.findByPath(":buildPlugin")!!
-
-      systemProperty("ui-test.intellij.version", version)
```
PawelLipski, line 34: System properties could as well be used here... but Typesafe config doesn't support them for some reason

frontend/ui/impl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/EnhancedGraphTable.java
```diff
@@ -167,7 +167,6 @@ private void subscribeToSelectedGitRepositoryChange() {
     gitRepositorySelectionProvider.addSelectionChangeObserver(() -> queueRepositoryUpdateAndModelRefresh());
   }
 
-  // TODO (#620): rework callbacks into futures
```
PawelLipski, line 170: Task abandoned long ago, turned out impractical

PawelLipski: Class above removed coz unused

src/uiTest/resources/ide.rhino.js
```diff
@@ -27,10 +27,6 @@ function Ide() {
     }
   };
 
-  this.getMajorVersion = function() {
```
PawelLipski, line 30: Unuuuuused

src/uiTest/scala/com/virtuslab/gitmachete/uitest/RunningIntelliJFixtureExtension.scala
```diff
@@ -6,22 +6,10 @@ import org.intellij.lang.annotations.Language
...
 
-  private val machetePlugin: Plugin = {
-    val pluginPath = sys.props
-      .get("ui-test.plugin.path")
-      .filterNot(_.isEmpty)
-      .getOrElse(throw new Exception("Plugin path is not provided"))
-    Plugin.Direct(Paths.get(pluginPath).toUri)
-  }
-
-  registerFixtureTransformer(_.withPlugin(machetePlugin))
```
PawelLipski, line 22: Removed, done via config

src/uiTest/scala/com/virtuslab/gitmachete/uitest/RunningIntelliJFixtureExtension.scala
```diff
@@ -6,22 +6,10 @@ import org.intellij.lang.annotations.Language
...
-  registerFixtureTransformer(_.withAfterIntelliJStartup((_, intelliJ) => intelliJ.ide.configure()))
```
PawelLipski, line 23: Can't be done via config; moved to `UISuite` tho, as this trait doesn't look like a good place for side-effecting statements

src/uiTest/scala/com/virtuslab/gitmachete/uitest/UISuite.scala
```diff
@@ -1,31 +1,18 @@
 package com.virtuslab.gitmachete.uitest
 
-import org.virtuslab.ideprobe.dependencies.IntelliJVersion
 import org.virtuslab.ideprobe.junit4.RunningIntelliJPerSuite
 import org.virtuslab.ideprobe.{IdeProbeFixture, IntelliJFixture}
 
 trait UISuite extends RunningIntelliJPerSuite with IdeProbeFixture with RunningIntelliJFixtureExtension {
 
-  lazy val intelliJVersion: IntelliJVersion = {
```
PawelLipski, line 9: Done cleanly via config

src/uiTest/resources/ideprobe.conf
```diff
@@ -22,6 +22,17 @@ probe {
     }
   }
 
+  intellij {
+    plugins = [
+      { uri = ${IDEPROBE_INTELLIJ_PLUGIN_URI} }
```
PawelLipski, line 27: Both env vars are mandatory now (`${...}`, not `${?...}`)

mkondratek (approved):
lgtm

src/uiTest/resources/ideprobe.conf
```diff
@@ -22,6 +22,17 @@ probe {
...
+    ]
+
+    # We're cheating here a bit since IDEPROBE_INTELLIJ_VERSION_BUILD might be either a build number or a release number,
+    # while we're always treating it as a build number.
+    # Still, as of ide-probe 0.42.0, even when release number like `2020.3` is passed as `version.build`, UI tests work just fine.
+    version.build = ${IDEPROBE_INTELLIJ_VERSION_BUILD}
```
LukaszKontowski, line 33: It works for ide-probe `0.42.0` but... After latest changes from https://github.com/VirtusLab/ide-probe/pull/252 - which should be released as `0.43.0` - it probably will not work well. The previous solution on `git-machete-intellij-plugin` used:
`fixtureFromConfig("ideprobe.conf").withVersion(intelliJVersion)`
which overrides the `IntelliJVersion` **after parsing configuration**. And this previous solution created an object of type `IntellijVersion` with `build` field specified correctly and `release` was a `None`. So, `ide-probe` used only the `build` parameter which worked just fine. New solution does not use that _trick_.

In effect, newer versions of`ide-probe` will get `version.build` from git-machete `ideprobe.conf` and `version.release` from the default ide-probe `reference.conf`. And that would result in issues like downloading not the version we specified here. If we want to go only with `ideprobe.conf` without the _withVersion_ tricks - we have to specify both `version.build` and `version.release` - so that they would match properly.

Anyway - did you test it against a snapshot version of `ide-probe` with changes from https://github.com/VirtusLab/ide-probe/pull/252 included?

LukaszKontowski: These change will be OK if we want to stick with `0.42.0` version of ide-probe. But for newer releases - this will not work well. The problem is also related to removing the default values from scala code of the ide-probe, which is connected with the way our configs are passed here: https://github.com/VirtusLab/ide-probe/blob/master/core/driver/sources/src/main/scala/org/virtuslab/ideprobe/ide/intellij/IntelliJProvider.scala#L184

LukaszKontowski (changes requested):

src/uiTest/resources/ideprobe.conf
LukaszKontowski: Alternatively - this changes could work well if we make a small change in the default `reference.conf` in ide-probe:
change from:
```
intellij {
    plugins = []
    version {
      build = "212.5080.55"
      release = "2021.2.1"
    }
```

to

```
intellij {
    plugins = []
    version {
      build = "212.5080.55"
      # release = ""
    }
```
so that `release` will be `None` by default

PawelLipski: Whooooops good spot, I didn't take it into account ☹️

Huh let's go deeper for a second... why is even this **double representation** needed by ide-probe? 🤔 I mean, if someone specifies `build` and `release` which point to different versions... what's gonna be the effect? Lemme open an issue for that in ide-probe...

PawelLipski: https://github.com/VirtusLab/ide-probe/issues/265, let's consider it a blocker for this PR then... changing to draft

PawelLipski: Ha! actually, as for now I think I have a simple workaround that would work even on 0.43.0+:

```
probe.intellij.version.release = null
```

See the newly-pushed commit. Un-drafting the PR

PawelLipski: Lemme still check it against a snapshot version of ide-probe, however...

LukaszKontowski: Looks like it works fine now!

LukaszKontowski (approved):
Should work fine both for ide-probe `0.42.0` and for newer releases, gr8 :)

PR #1090, opened 2022-08-27
Remove false negatives from unused classes checks by parsing plugin.xml

PawelLipski:


src/test/java/com/virtuslab/archunit/ClassStructureTestSuite.java
```diff
@@ -44,19 +48,42 @@ public void check(JavaClass javaClass, ConditionEvents events) {
...
+      val node = nodeList.item(i);
+      val attributes = node.getAttributes();
+      for (int j = 0; j < attributes.getLength(); j++) {
+        val attribute = attributes.item(j);
+        val maybeFqcn = attribute.getNodeValue();
+        try {
+          val clazz = Class.forName(maybeFqcn, /* initialize */ false, classLoader);
+          result.add(clazz);
+        } catch (ClassNotFoundException e) {
+          /* ignore */
```
mkondratek, line 68: don't we need any handling of this case? or is it handled in some other way? if so it may be worth to leave some comment here why we ignore it

src/test/java/com/virtuslab/archunit/ClassStructureTestSuite.java
```diff
@@ -44,19 +48,42 @@ public void check(JavaClass javaClass, ConditionEvents events) {
...
-            // Classes in frontend:file may be referenced from plugin.xml
-            "com.virtuslab.gitmachete.frontend.file..")
-        // Actions may be referenced from plugin.xml
-        .and().haveSimpleNameNotEndingWith("Action")
-        // For some reason, references to SlideInDialogKt aren't detected (Kotlin class?)
+            // For some reason, ArchUnit (com.tngtech.archunit.core.domain.JavaClass.getAccessesFromSelf)
+            // doesn't see accesses to static fields
+            "com.virtuslab.gitmachete.frontend.defs")
+        .and().doNotBelongToAnyOf(classesReferencedFromPluginXmlAttributes)
+        // For some reason, ArchUnit doesn't see access to the static field GIT_REF_PROTOTYPE_VALUE
```
mkondratek, line 86: hmmm. y dont we bring it into the class? probably into the companion object. would it solve that problem?

If this one remains pls use some {@link classreference}  or smth (if possible)

src/test/java/com/virtuslab/archunit/ClassStructureTestSuite.java
```diff
@@ -44,19 +48,42 @@ public void check(JavaClass javaClass, ConditionEvents events) {
     }
   }
 
+  @SneakyThrows
+  private Set<Class<?>> extractAllClassesReferencedFromPluginXmlAttributes() {
```
mkondratek, line 52: nice

mkondratek (approved):

src/test/java/com/virtuslab/archunit/ClassStructureTestSuite.java
PawelLipski: Inlined the constant completely

src/test/java/com/virtuslab/archunit/ClassStructureTestSuite.java
PawelLipski: Inlined the constant completely

PR #1097, opened 2022-08-30
Release v3.0.0

PawelLipski:


.circleci/config.yml
```diff
@@ -196,8 +227,7 @@ jobs:
                 command: |
                   source scripts/utils.sh
                   tag=v$(extract_version_from_current_wd)
-                  # Newlines (<br/>) are necessary for better layout on Marketplace, but make the formatting worse on GitHub
-                  change_notes=$(sed '/^<br.>$/d' CHANGE-NOTES.html)
+                  change_notes=$(./gradlew getChangelog --no-header -q)
```
PawelLipski, line 230: Actually @MaciejG604 can it be simplified even more? `--no-header` skipped here, and just `$change_notes` with no `$tag` prefix passed as `$message`?

CHANGE-NOTES.md
```diff
@@ -0,0 +1,165 @@
+# Changelog
+
+## Unreleased
```
PawelLipski, line 3: To be bumped on my pre-release PR

.circleci/config.yml
MaciejG604: Unfortunately no, this is what gets created: [Draft release](https://github.com/VirtusLab/git-machete-intellij-plugin/releases)

PawelLipski: ☹️

PR #1100, opened 2022-08-31
Shorten report url supressed exceptions

MaciejG604:


frontend/ui/impl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/GitMacheteErrorReportSubmitter.java
```diff
@@ -122,17 +122,19 @@ private java.util.Map<String, String> getTemplateVariables(
     return templateVariables;
   }
 
-  private String shortenWrappedExceptionsStack(String stackTrace) {
+  private String shortenExceptionsStack(String stackTrace) {
     val nl = System.lineSeparator();
-    val rootCauseIndex = stackTrace.lastIndexOf("Caused by:");
+    val rootCauseIndex = Math.max(
+        stackTrace.lastIndexOf("Caused by:"),
+        stackTrace.lastIndexOf("\tSuppressed:"));
```
PawelLipski, line 129: Nit: `\t` included here but not in `contains` below... intentional?

PawelLipski (approved):

frontend/ui/impl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/GitMacheteErrorReportSubmitter.java
MaciejG604: Yes, intentional. contains() just tests if line matches, but here i wanted to keep the TAB in the rootCauseStackTrace to get correct formatting. (Will still look wrong with multiply-nested wrapped exceptions)

PR #1102, opened 2022-08-31
Bump io.freefair.gradle:aspectj-plugin:6.4.3.1->6.5.1

mkondratek:


PawelLipski:
Huh maybe just apply plugin by name, not by class ne

PawelLipski:
Meh, this:

```kotlin
apply(plugin = "org.jetbrains.kotlin.jvm")
```

doesn't seem to be a fix either :/

```
* What went wrong:
Failed to apply plugin 'org.jetbrains.kotlin.jvm'.
> Could not create an instance of type org.jetbrains.kotlin.gradle.dsl.KotlinJvmProjectExtension.
   > Companion
```

PawelLipski:
Huh weird... anyway, let's check if https://github.com/VirtusLab/git-machete-intellij-plugin/pull/1200 passes (passed on my machine)

PR #1103, opened 2022-09-02
Correct status displayed on graph during ongoing git repository operation

MaciejG604:


MaciejG604:
When repository state is APPLYING (e.g. after `git am`) you can freely checkout other branches, but the ongoing operation is not bound to one branch, rather it can always be applied on what is `HEAD` at the moment. This is reflected in the change.

backend/impl/src/main/java/com/virtuslab/gitmachete/backend/impl/GitMacheteRepository.java
```diff
@@ -343,10 +344,23 @@ IGitMacheteRepositorySnapshot createSnapshot(IBranchLayout branchLayout) throws
...
+
       return new GitMacheteRepositorySnapshot(List.narrow(rootBranches), branchLayout, currentBranchIfManaged,
-          managedBranchByName, duplicatedBranchNames, skippedBranchNames, gitCoreRepository.deriveRebasedBranch(),
-          preRebaseHookExecutor,
-          ongoingOperation);
+          managedBranchByName, duplicatedBranchNames, skippedBranchNames, preRebaseHookExecutor,
+          ongoingOperation, operationsBaseBranchName);
+    }
+
+    @SneakyThrows
```
PawelLipski, line 354: Hmm I'd avoid `@SneakyThrows` in production (non-test) code unless there's a good reason too... what's the exception that can be thrown from this method's body?

backend/impl/src/main/java/com/virtuslab/gitmachete/backend/impl/GitMacheteRepositorySnapshot.java
```diff
@@ -36,14 +36,14 @@ public class GitMacheteRepositorySnapshot implements IGitMacheteRepositorySnapsh
   @Getter
   private final Set<String> skippedBranchNames;
 
-  @Getter
-  private final Option<String> rebasedBranchName;
-
   private final PreRebaseHookExecutor preRebaseHookExecutor;
 
   @Getter
   private final OngoingRepositoryOperation ongoingRepositoryOperation;
```
PawelLipski, line 42: Actually... would it make sense to group the enum defining the operation (currently: `OngoingRepositoryOperation`) and the affected branch name into a single data class with 2 fields? 🤔

PawelLipski: Esp. that AFAICS the operation type and branch name are almost always derived&used together

backend/impl/src/main/java/com/virtuslab/gitmachete/backend/impl/GitMacheteRepository.java
PawelLipski: Generally I'd rather wrap the exception into `GitMacheteException`, as done in other methods

backend/impl/src/main/java/com/virtuslab/gitmachete/backend/impl/GitMacheteRepositorySnapshot.java
MaciejG604: Sure, why not. Also will be more verbose than fiveWordLongNames

backend/api/src/main/java/com/virtuslab/gitmachete/backend/api/IGitMacheteRepositorySnapshot.java
```diff
@@ -29,7 +29,10 @@ public interface IGitMacheteRepositorySnapshot {
   Option<IExecutionResult> executeMachetePreRebaseHookIfPresent(IGitRebaseParameters gitRebaseParameters)
       throws GitMacheteException;
 
-  OngoingRepositoryOperation getOngoingRepositoryOperation();
+  interface IOngoingRepositoryOperationInfo {
```
PawelLipski, line 32: Doesn't make much sense to separate interface and implementation for a basic data class like that, just declaring `OngoingRepositoryOperationInfo` in `backend:api` Gradle module is okay. The boundary since when it makes is sense to separate intf&impl is somewhat fuzzy (not to say arbitrary)... but here I'd opt for simply just using concrete class even in the API module

frontend/ui/impl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/DemoGitMacheteRepositorySnapshot.java
```diff
@@ -133,13 +133,18 @@ public Option<IExecutionResult> executeMachetePreRebaseHookIfPresent(IGitRebaseP
...
   @Override
-  public OngoingRepositoryOperation getOngoingRepositoryOperation() {
-    return OngoingRepositoryOperation.NO_OPERATION;
-  }
-
-  @Override
-  public Option<String> getOngoingOperationsBaseBranchName() {
-    return Option.none();
+  public IOngoingRepositoryOperationInfo getOngoingRepositoryOperationInfo() {
+    return new IOngoingRepositoryOperationInfo() {
```
PawelLipski, line 137: Also, I'd name the data class `OngoingRepositoryOperation`, and the enum `OngoingRepositoryOperationType`

backend/api/src/main/java/com/virtuslab/gitmachete/backend/api/IGitMacheteRepositorySnapshot.java
PawelLipski: Or actually, it isn't so fuzzy: the question you need to ask yourself is whether keeping the impl class inside api module **would expose any underlying details about the underlying implementation in the api**? In our case, would you need to leak details about gitCore:jGit to backend:api? We strive to keep backend:api INdependent from gitCore:jGit. Or in general, we keep all api modules independent from all impl modules.

PawelLipski: And in case of this particular class, you won't really expose anything about backend:impl's deps (gitCore:jGit) by declaring it as concrete class in backend:api

backend/api/src/main/java/com/virtuslab/gitmachete/backend/api/IGitMacheteRepositorySnapshot.java
```diff
@@ -29,10 +29,10 @@ public interface IGitMacheteRepositorySnapshot {
   Option<IExecutionResult> executeMachetePreRebaseHookIfPresent(IGitRebaseParameters gitRebaseParameters)
       throws GitMacheteException;
 
-  interface IOngoingRepositoryOperationInfo {
-    OngoingRepositoryOperation getOperationType();
+  interface IOngoingRepositoryOperation {
```
PawelLipski, line 32: For the problem with Interning Checker, look for `lombok-generated equals` string across the project

backend/api/src/main/java/com/virtuslab/gitmachete/backend/api/OngoingRepositoryOperationType.java
```diff
@@ -1,5 +1,5 @@
 package com.virtuslab.gitmachete.backend.api;
 
-public enum OngoingRepositoryOperation {
+public enum OngoingRepositoryOperationType {
```
PawelLipski, line 3: Rename the file accordingly

backend/api/src/main/java/com/virtuslab/gitmachete/backend/api/IGitMacheteRepositorySnapshot.java
MaciejG604: It actually does not complain here.

CONTRIBUTING.md
```diff
@@ -49,6 +49,9 @@ Use IntelliJ IDEA Community Edition/Ultimate.
    It can be significantly useful while working with UI components (or tests).
    To investigate the UI you may want to use `Tools > Internal Actions > UI > UI Inspector`.
 
+10. Consider setting `Class count before import with '*'` and `Names count to use static import with '*'` to a very high number
```
PawelLipski, line 52: Nit: specify the exact number (like 100 or whatnot)

backend/api/src/main/java/com/virtuslab/gitmachete/backend/api/IGitMacheteRepositorySnapshot.java
MaciejG604: Okay it does. 😆

CONTRIBUTING.md
```diff
@@ -49,6 +49,9 @@ Use IntelliJ IDEA Community Edition/Ultimate.
    It can be significantly useful while working with UI components (or tests).
    To investigate the UI you may want to use `Tools > Internal Actions > UI > UI Inspector`.
 
+10. Consider setting `Class count before import with '*'` and `Names count to use static import with '*'` to a very high number (e.q. 500)
```
PawelLipski, line 52:
```suggestion
10. Consider setting `Class count before import with '*'` and `Names count to use static import with '*'` to a very high number (e.g. 500)
```

_exempli gratia_ 🤓

CONTRIBUTING.md
```diff
@@ -49,6 +49,9 @@ Use IntelliJ IDEA Community Edition/Ultimate.
...
+to avoid problems with CheckStyle when editing code. To do that go to `Preferences > Code Style > Java/Kotlin/Scala`.
```
PawelLipski, line 53: Huh do you have `Preferences` in your IntelliJ, could you show the screenshot? For me it's sth like `File > Settings > Editor > Code Style > Java > Imports`

PawelLipski (approved):

backend/api/src/main/java/com/virtuslab/gitmachete/backend/api/IGitMacheteRepositorySnapshot.java
```diff
@@ -29,7 +30,14 @@ public interface IGitMacheteRepositorySnapshot {
   Option<IExecutionResult> executeMachetePreRebaseHookIfPresent(IGitRebaseParameters gitRebaseParameters)
       throws GitMacheteException;
 
-  OngoingRepositoryOperation getOngoingRepositoryOperation();
+  @Data
+  // So that Interning Checker doesn't complain about enum comparison (by `equals` and not by `==`) in Lombok-generated `equals`
+  @SuppressWarnings("interning:not.interned")
```
PawelLipski, line 35:
```suggestion
  @SuppressWarnings("interning:unnecessary.equals")
```
(or both `not.interned` and `unnecessary.equals`? sth like that)

CONTRIBUTING.md
MaciejG604: image

"You can use ⌘, (macOS), or Ctrl+Alt+S (Windows/Linux) to display the Preferences/Settings dialog in IntelliJ IDEA."

[IntelliJ IDEA Guide](https://www.jetbrains.com/idea/guide/tutorials/configuring-intellij-idea/configuring-preferences-settings/)

PR #1106, opened 2022-09-05
Contribution.md: info on bundled plugins and gradle term of service

vzmerr:
- added information about checking the bundled plugins in the Installed tab

- added information about accepting gradle term of service in the first build


PawelLipski (approved):

PawelLipski:
@vzmerr merge once you're back? in principle, in this project it's the PR author that does the merging

PR #1117, opened 2022-09-10
Relax the ban on inner classes

PawelLipski:


src/test/java/com/virtuslab/archunit/ClassStructureTestSuite.java
```diff
@@ -104,20 +106,32 @@ public void all_classes_should_be_referenced() {
...
 
   @Test
-  public void inner_classes_should_not_be_used() {
+  public void inner_classes_should_not_be_instantiated_from_constructor_of_enclosing_class() {
     noClasses()
-        .that().haveNameNotMatching("^.*\\$[0-9]+$") // inner classes autogenerated by the compiler
+        .that()
+        .doNotHaveFullyQualifiedName(com.virtuslab.gitmachete.frontend.actions.dialogs.GitPushDialog.class.getName())
         .and()
-        .haveNameNotMatching("^" + com.virtuslab.gitmachete.frontend.actions.dialogs.GitPushDialog.class.getName() + "\\$.*$")
```
PawelLipski, line 111: This check is in a way more strict then it used to be, as we now also check **lambdas** instantiated in c'tors.
Previously, we were checking **all expect for lambdas**, but in all methods (not only c'tors).

MaciejG604 (approved):

PR #1118, opened 2022-09-12
Add break lines in possibly too long tooltip text

MaciejG604:


MaciejG604:
I thought about using a function to split tooltip text in half either on a `'.'` or `' '`, but then there is a possibility to split a branch name that contains dots and that could bother the user much more than long one-liners.

frontend/ui/impl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/cell/BranchOrCommitCellRendererComponent.java
```diff
@@ -313,14 +313,15 @@ private static String getSyncToParentStatusBasedToolTipText(INonRootManagedBranc
                 .format(currentBranchName, parentBranchName)),
```
PawelLipski, line 313: Nit: I'd make `sync-to-parent-status-tooltip.in-sync` string a HTML one as well for consistency (even if `<br>` isn't needed)

PawelLipski (approved):

PawelLipski (approved):

PawelLipski (approved):

PR #1124, opened 2022-09-15
close milestone on release

### LINKED ISSUE #1025: After opening a Release on GitHub, close the corresponding milestone
See https://github.com/VirtusLab/git-machete/issues/608

vzmerr:
resolves #1025 

It is using [update milestone on release](https://github.com/mhutchie/update-milestone-on-release)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ version: 2.1
 executors:
   docker_executor:
     docker:
-      - image: 'gitmachete/intellij-plugin-ci:8.0.0'
+      - image: 'gitmachete/intellij-plugin-ci:8.1.0'
     working_directory: ~/git-machete-intellij-plugin
     resource_class: large
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,4 +1,6 @@
 version: 2.1
+orbs:
+  github-cli: circleci/github-cli@2.1.0
 
 executors:
   docker_executor:
@@ -232,7 +234,9 @@ jobs:
                     --attach build/distributions/git-machete-intellij-plugin-*.zip \
                     --message "$tag"$'\n\n'"$change_notes" \
                     $tag
-
+            - run:
+                name: Close the Release Milestone
+                command: scripts/close-milestone v$(extract_version_from_current_wd)
 
   checkForUpdates:
     executor: docker_executor
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -227,16 +227,15 @@ jobs:
                 # Implicit usage of ${GITHUB_TOKEN} by hub
                 # language=sh
                 command: |
+                  corresponding_milestone_title=v$(extract_version_from_current_wd)
                   source scripts/utils.sh
-                  tag=v$(extract_version_from_current_wd)
+                  tag=$corresponding_milestone_title
                   change_notes=$(./gradlew getChangelog --no-header -q)
                   hub release create \
                     --attach build/distributions/git-machete-intellij-plugin-*.zip \
                     --message "$tag"$'\n\n'"$change_notes" \
                     $tag
-            - run:
-                name: Close the Release Milestone
-                command: scripts/close-milestone v$(extract_version_from_current_wd)
+                  scripts/close-milestone $corresponding_milestone_title
 
   checkForUpdates:
     executor: docker_executor
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,4 @@
 version: 2.1
-orbs:
-  github-cli: circleci/github-cli@2.1.0
 
 executors:
   docker_executor:
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -7,7 +7,7 @@ SHELL ["/bin/bash", "-euxo", "pipefail", "-c"]
 
 RUN \
   apt-get update; \
-  apt-get install --no-install-recommends -y git hub openssh-client \
+  apt-get install --no-install-recommends -y git hub openssh-client
 
 #Install GitHub Cli
 Run \
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -7,7 +7,15 @@ SHELL ["/bin/bash", "-euxo", "pipefail", "-c"]
 
 RUN \
   apt-get update; \
-  apt-get install --no-install-recommends -y git hub openssh-client
+  apt-get install --no-install-recommends -y git hub openssh-client \
+
+#Install GitHub Cli
+Run \
+  curl -fsSL https://cli.github.com/packages/githubcli-archive-keyring.gpg | dd of=/usr/share/keyrings/githubcli-archive-keyring.gpg \
+  && chmod go+r /usr/share/keyrings/githubcli-archive-keyring.gpg \
+  && echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/githubcli-archive-keyring.gpg] https://cli.github.com/packages stable main" | tee /etc/apt/sources.list.d/github-cli.list > /dev/null \
+  && apt-get update \
+  && apt-get install gh -y
 
 # Disable IntelliJ data sharing
 RUN \
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -12,7 +12,6 @@ RUN \
 #Install GitHub Cli
 Run \
   curl -fsSL https://cli.github.com/packages/githubcli-archive-keyring.gpg | dd of=/usr/share/keyrings/githubcli-archive-keyring.gpg \
-  && chmod go+r /usr/share/keyrings/githubcli-archive-keyring.gpg \
   && echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/githubcli-archive-keyring.gpg] https://cli.github.com/packages stable main" > /etc/apt/sources.list.d/github-cli.list \
   && apt-get update \
   && apt-get install gh -y
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -13,7 +13,7 @@ RUN \
 Run \
   curl -fsSL https://cli.github.com/packages/githubcli-archive-keyring.gpg | dd of=/usr/share/keyrings/githubcli-archive-keyring.gpg \
   && chmod go+r /usr/share/keyrings/githubcli-archive-keyring.gpg \
-  && echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/githubcli-archive-keyring.gpg] https://cli.github.com/packages stable main" | tee /etc/apt/sources.list.d/github-cli.list > /dev/null \
+  && echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/githubcli-archive-keyring.gpg] https://cli.github.com/packages stable main" > /etc/apt/sources.list.d/github-cli.list \
   && apt-get update \
   && apt-get install gh -y
 
--- a/scripts/close-milestone
+++ b/scripts/close-milestone
@@ -1,0 +1,13 @@
+#!/usr/bin/env bash
+
+set -x -e -o pipefail -u
+
+milestone_title=$1
+
+indented_milestones_json_array=$(gh api -H "Accept: application/vnd.github+json" /repos/VirtusLab/git-machete-intellij-plugin/milestones)
+
+number=$( echo $indented_milestones_json_array | jq -c --arg TITLE $milestone_title '.[] | select(.title | contains($TITLE)) | .number')
+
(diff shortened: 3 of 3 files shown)


PawelLipski:
Whoops I'd stick to CircleCI completely, it's hard to manage 2 concurrent CIs (very prone to race conditions etc. in general)

scripts/close-milestone.sh
```diff
@@ -0,0 +1,8 @@
+echo "'.[] | select(.title | contains("$1"))'" | read jqMilestoneSearchCommand
```
PawelLipski, line 1: Whoops shebang + `set -e -o pipefail -u`, as in the other scripts (https://virtuslab.github.io/tips/#shell/strict-mode)

PawelLipski: Also, `read -r`

PawelLipski: Or just `foo=$(...)` instead of `... | read foo`, more natural

vzmerr: Thank you for this. read was as well removing the whitespace which was causing error in jq. Let me check and see.

scripts/close-milestone.sh
```diff
@@ -0,0 +1,14 @@
+set -e -o pipefail -u
+jqMilestoneSearchCommand=$(echo "'.[] | select(.title | contains("$1"))'")
```
PawelLipski, line 2: Also, pls use `lower_snake_case` for shell variables

vzmerr:
I wanted to test as well by an InvalidRelease, but I was wondering if it would cause other issues.

.circleci/config.yml
```diff
@@ -232,7 +232,9 @@ jobs:
                     --attach build/distributions/git-machete-intellij-plugin-*.zip \
                     --message "$tag"$'\n\n'"$change_notes" \
                     $tag
-
+            - run:
+                name: Close the Release Milestone
+                command: scripts/close-milestone.sh v$(extract_version_from_current_wd)
```
PawelLipski, line 237: Hmm is `extract_version_from_current_wd` available in this context? Pls test (for the tests, move this step out of the `when` block)

scripts/close-milestone.sh
```diff
@@ -0,0 +1,11 @@
+set -x -e -o pipefail -u
+
+indented_milestones_json_array=$(gh api -H "Accept: application/vnd.github+json" /repos/VirtusLab/git-machete-intellij-plugin/milestones)
```
PawelLipski, line 3: Same concern, is `gh` available in the CI?

scripts/close-milestone.sh
```diff
@@ -0,0 +1,11 @@
...
+
+desired_milestone=$( echo $indented_milestones_json_array | jq -c --arg TITLE $1 '.[] | select(.title | contains($TITLE))')
```
PawelLipski, line 5: Don't use vanilla `$1` inside the script like that; name the param first like e.g. in https://github.com/VirtusLab/git-machete-intellij-plugin/blob/master/scripts/git-hooks/machete-status-branch#L9

scripts/close-milestone.sh
```diff
@@ -0,0 +1,11 @@
...
+
+number=$(echo $desired_milestone | jq '.number')
```
PawelLipski, line 7: Nit: just append `| .number` to the jq expression in the preceding line, no need for a separate statement (as `desired_milestone` is otherwise unused)

scripts/close-milestone.sh
```diff
@@ -0,0 +1,11 @@
+set -x -e -o pipefail -u
```
PawelLipski, line 1: Pls add shebang

scripts/close-milestone.sh
PawelLipski: (see `docker/Dockerfile` + `docker_executor` in CircleCI config)

scripts/close-milestone.sh
```diff
@@ -0,0 +1,11 @@
...
+
+indented_milestones_json_array=$(gh api -H "Accept: application/vnd.github+json" /repos/VirtusLab/git-machete-intellij-plugin/milestones)
+
+desired_milestone=$( echo $indented_milestones_json_array | jq -c --arg TITLE $1 '.[] | select(.title | contains($TITLE))')
+
+number=$(echo $desired_milestone | jq '.number')
+
+milestone_url=$(echo "/repos/VirtusLab/git-machete-intellij-plugin/milestones/$number")
```
PawelLipski, line 9: Just
```suggestion
milestone_url="/repos/VirtusLab/git-machete-intellij-plugin/milestones/$number"
```

scripts/close-milestone.sh
PawelLipski: Also... `indented_milestones_...`? 🤔  it's rather `all_milestones_...` or sth like that

vzmerr: I noticed that the `gh` api returns the pretty printed json, instead of the compacted one. `jq` seemed to fail in parsing  in that case.  both assigning the result json to a variable, or using `read` would remove the excess whitespace.

PawelLipski: Hmmm `jq` definitely shouldn't fail depending just on whitespace... could you cook up a repro for that? I.e. a sample such pretty-printed JSON that `jq` would fail on

vzmerr: is it a good approach to introduce this?
https://circleci.com/developer/orbs/orb/circleci/github-cli

.circleci/config.yml
vzmerr: Thanks for bringing this up. I was thinking of assigning the value of `extract_version_from_current_wd` before the release to a value in the release script, and move the logic to close the associated milestone to the end of the release script. What do you think?

PawelLipski: Okay, let's give it a shot 👌🏻

scripts/close-milestone.sh
PawelLipski: Hmmm in general it'd make sense... although for the sake of consistency I'd rather have `gh` installed in the Docker image, as is the case for all the other software required in the CI.

See CONTRIBUTING.md for the CI image regeneration instructions... the tag should be 8.1.0 according to semver (a non-breaking change). 

Pls provide your DockerHub login (create an account if you don't have one), I'll add you with Write (push) permissions

scripts/close-milestone
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+
+set -x -e -o pipefail -u
+
+milestone_title=$1
+
+indented_milestones_json_array=$(gh api -H "Accept: application/vnd.github+json" /repos/VirtusLab/git-machete-intellij-plugin/milestones)
+
+number=$( echo $indented_milestones_json_array | jq -c --arg TITLE $milestone_title '.[] | select(.title | contains($TITLE)) | .number')
```
PawelLipski, line 9: To prevent quoting issues:
```suggestion
number=$(echo "$indented_milestones_json_array" | jq -c --arg TITLE $milestone_title '.[] | select(.title | contains($TITLE)) | .number')
```

scripts/close-milestone.sh
vzmerr: DockerHub login is vzmerr

vzmerr: It seems that it is [not a good practice](https://github.com/cli/cli/discussions/3698#discussioncomment-785838) to install `gh` on a Docker container. What do you think?

PawelLipski: > DockerHub login is vzmerr

Pls check now (https://hub.docker.com/orgs/gitmachete/repositories)

PawelLipski: > It seems that it is https://github.com/cli/cli/discussions/3698#discussioncomment-785838 to install gh on a Docker container.

I'd say the thread generally speaks in favor of dockerized `gh`, it's quite useful in some cases:
* as per https://github.com/cli/cli/discussions/3698#discussioncomment-3472925, for ease of use in setups like kubernetes
* as per https://github.com/cli/cli/discussions/3698#discussioncomment-3327598, for the use of `gh` in CI pipelines, 

Note that by we do **not** need to use a standalone Docker image with `gh`, it's more of matter of installing `gh` into https://hub.docker.com/repository/docker/gitmachete/intellij-plugin-ci, as `hub` already is (see line 10 in Dockerfile)

scripts/close-milestone
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+
+set -x -e -o pipefail -u
+
+milestone_title=$1
+
+indented_milestones_json_array=$(gh api -H "Accept: application/vnd.github+json" /repos/VirtusLab/git-machete-intellij-plugin/milestones)
+
+number=$( echo "$indented_milestones_json_array" | jq -c --arg TITLE $milestone_title '.[] | select(.title | contains($TITLE)) | .number')
+
```
PawelLipski, line 10: I'd also check for whether there're still any open PRs/issues in the milestone, and only close the milestone if there aren't any

docker/Dockerfile
```diff
@@ -7,7 +7,15 @@ SHELL ["/bin/bash", "-euxo", "pipefail", "-c"]
 
 RUN \
   apt-get update; \
-  apt-get install --no-install-recommends -y git hub openssh-client
+  apt-get install --no-install-recommends -y git hub openssh-client \
+
+#Install GitHub Cli
+Run \
+  curl -fsSL https://cli.github.com/packages/githubcli-archive-keyring.gpg | dd of=/usr/share/keyrings/githubcli-archive-keyring.gpg \
+  && chmod go+r /usr/share/keyrings/githubcli-archive-keyring.gpg \
```
PawelLipski, line 15: `chmod` is unlikely to be required; you're running all commands as **root** anyway

docker/Dockerfile
```diff
@@ -7,7 +7,15 @@ SHELL ["/bin/bash", "-euxo", "pipefail", "-c"]
...
+  && echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/githubcli-archive-keyring.gpg] https://cli.github.com/packages stable main" | tee /etc/apt/sources.list.d/github-cli.list > /dev/null \
```
PawelLipski, line 16: Or just
```suggestion
  && echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/githubcli-archive-keyring.gpg] https://cli.github.com/packages stable main" > /etc/apt/sources.list.d/github-cli.list \
```

scripts/close-milestone
```diff
@@ -0,0 +1,21 @@
...
+milestone_title=$1
+
+milestones_json_array=$(gh api -H "Accept: application/vnd.github+json" /repos/VirtusLab/git-machete-intellij-plugin/milestones)
+
+desired_milestone=$(echo "$milestones_json_array" | jq -c --arg TITLE $milestone_title '.[] | select(.title | contains($TITLE))')
+
+number=$(echo "$desired_milestone" | jq '.number')
+
+#open issues and PRs are bundled together in the same field
+open_issues=$(echo "$desired_milestone" | jq '.open_issues')
```
PawelLipski, line 14: Oh so `open_issues` are a **number** field (count of the issues/PRs), not the **list** of issues/PRs right?

.circleci/config.yml
```diff
@@ -225,14 +225,15 @@ jobs:
...
                   source scripts/utils.sh
-                  tag=v$(extract_version_from_current_wd)
+                  tag=$corresponding_milestone_title
                   change_notes=$(./gradlew getChangelog --no-header -q)
                   hub release create \
                     --attach build/distributions/git-machete-intellij-plugin-*.zip \
                     --message "$tag"$'\n\n'"$change_notes" \
                     $tag
-
+                  scripts/close-milestone $corresponding_milestone_title
```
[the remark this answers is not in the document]
PawelLipski: Pls test if `gh` is available on a step outside of the `when:` block

scripts/close-milestone
vzmerr: right.

.circleci/config.yml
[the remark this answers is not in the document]
vzmerr: >You're still using the same Docker image (see docker_executor): gitmachete/intellij-plugin-ci:8.0.0, which does not contain gh :/

Right +1 So there is a need to put the new version `8.0.1`, also.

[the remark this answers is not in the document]
PawelLipski: 8.1.0, as from the perspective of the CI pipeline that uses the image, it's a **change** (so MAJOR or MINOR bump), but a backward-compatible one (so MINOR).

Additional non-trivia on semver: https://virtuslab.github.io/tips/#versioning/semver-vs-forward-compatibility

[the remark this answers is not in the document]
PawelLipski: And https://virtuslab.github.io/tips/#versioning/semantic-versioning

PawelLipski (approved):

PR #1133, opened 2022-09-19
Add hyperlink to warning dialog

MaciejG604:


scripts/prohibit-properties-unused-keys
```diff
@@ -10,7 +10,7 @@ bundle=frontend/resourcebundles/src/main/resources/GitMacheteBundle.properties
 
 # Let's exclude the keys that are used by IntelliJ implicitly as action descriptions (see `<override-text>` tags in plugin.xml).
 all_expected_keys=$(
-  grep --only-matching --perl-regexp '.*(?==)' -- $bundle \
+  grep --only-matching --perl-regexp '.*(?==)(?<!href)' -- $bundle \
```
PawelLipski, line 13: Maybe sth like
```suggestion
  grep --only-matching --perl-regexp '[^=]*(?==)' -- $bundle \
```

PawelLipski (approved):

scripts/prohibit-properties-unused-keys
MaciejG604: An equal sign that is not preceded by another equal sign, right?

MaciejG604: This is incorrect, I'm trying to figure out why as this way of handling an equal signs in a string is much more elegant and robust.

PR #1136, opened 2022-09-19
Check if all local branches are managed before suggesting rediscovery

MaciejG604:


frontend/resourcebundles/src/main/resources/GitMacheteBundle.properties
```diff
@@ -309,7 +309,7 @@ action.GitMachete.EnhancedGraphTable.branch-layout-write-failure=Writing new bra
 
 string.GitMachete.RediscoverSuggester.dialog.title=Git Machete: Rediscover Suggestion
 string.GitMachete.RediscoverSuggester.dialog.question=It looks that you have not modified your git machete file for a while. Would you like to rediscover the branch layout?
-
+string.GitMachete.RediscoverSuggester.backgroundable-check-task.title=Check if (re)discovery should be performed
```
mkondratek, line 312: Sounds a bit too internal 🤔 what about smth like "Git Machete: Checking for repository update..."?

frontend/ui/impl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/RediscoverSuggester.java
```diff
@@ -71,4 +80,32 @@ private long daysDiffTillNow(long lastModifiedTimeMillis) {
...
+  private boolean areAllLocalBranchesManaged() {
+    val localBranches = gitRepository.getBranches().getLocalBranches();
+    val branchLayoutReader = RuntimeBinding.instantiateSoleImplementingClass(IBranchLayoutReader.class);
+    try {
+      val branchLayout = branchLayoutReader.read(gitRepository.getMacheteFilePath());
+      val areLocalBranchesManagedSeq = localBranches.stream().map(GitReference::getName)
+          .map(branchLayout::findEntryByName)
+          .collect(Collectors.toList());
+
+      return Option.sequence(areLocalBranchesManagedSeq).isDefined();
```
mkondratek, line 106: nice 👍

frontend/ui/impl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/RediscoverSuggester.java
```diff
@@ -71,4 +80,32 @@ private long daysDiffTillNow(long lastModifiedTimeMillis) {
     val millisDiff = currentTimeMillis - lastModifiedTimeMillis;
     return millisDiff / (24 * 60 * 60 * 1000);
   }
+
+  public void enqueueChecksThenMaybePerform() {
```
mkondratek, line 84: nit: I do not like this name xd "maybe" sounds a bit random XD `checkAndPerformIfApplicable` maybe?

MaciejG604: True 😆

mkondratek (approved):
lgtm

frontend/ui/impl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/RediscoverSuggester.java
PawelLipski: Hmmm are you checking here whether **all** local branches are managed, or whether **any** local branch is managed? 🤔

frontend/ui/impl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/root/GitMachetePanel.java
```diff
@@ -79,8 +79,9 @@ public void ancestorAdded(AncestorEvent event) {
           val macheteFilePath = gitRepository.getMacheteFilePath();
           Runnable queueDiscoverOperation = () -> graphTable.queueDiscover(macheteFilePath, () -> {});
           val rediscoverSuggester = new RediscoverSuggester(gitRepository, queueDiscoverOperation);
-          graphTable.queueRepositoryUpdateAndModelRefresh(() -> rediscoverSuggester.perform());
+          graphTable.queueRepositoryUpdateAndModelRefresh(rediscoverSuggester::enqueueChecksAndPerformIfApplicable);
         }
+        removeAncestorListener(this);
```
PawelLipski, line 84: Hmmm is this change related to the PR? 🤔

PawelLipski: Anyway, I'd leave a comment in code on why it's needed as it's quite non-trivial

frontend/ui/impl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/RediscoverSuggester.java
```diff
@@ -71,4 +80,32 @@ private long daysDiffTillNow(long lastModifiedTimeMillis) {
...
+  }
+
+  private boolean areAllLocalBranchesManaged() {
+    val localBranches = gitRepository.getBranches().getLocalBranches();
+    val branchLayoutReader = RuntimeBinding.instantiateSoleImplementingClass(IBranchLayoutReader.class);
+    try {
+      val branchLayout = branchLayoutReader.read(gitRepository.getMacheteFilePath());
+      val areLocalBranchesManagedSeq = localBranches.stream().map(GitReference::getName)
+          .map(branchLayout::findEntryByName)
+          .collect(Collectors.toList());
```
PawelLipski, line 104: Somewhat simpler (with `io.vavr.collection.List`):
```java
      val managedLocalBranches = List.ofAll(localBranches)
            .map(GitReference::getName)
            .flatMap(branchLayout::findEntryByName);
```

PawelLipski:
Moving to v3.0.2 to avoid bloating the scope of v3.0.1

PawelLipski (changes requested):
As in the comments.

frontend/ui/impl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/RediscoverSuggester.java
MaciejG604: `Option.sequence()` will be `None` if any element in the sequence are `None`.

PawelLipski: > Option.sequence() will be None if any element in the sequence are None.

Whoops that's very non-obvious... I wouldn't rely on that, pls make the check more explicit.

In fact I'm now not even sure if [Cats](https://typelevel.org/cats/) which is probably the source of inspiration for Vavr wrt. `sequence`/`traverse` would behave the same way (probably it would, but again, I wouldn't rely on that in code that's supposed to future-proof/maintainable)

frontend/ui/impl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/root/GitMachetePanel.java
```diff
@@ -71,6 +71,8 @@ public void after(java.util.List<? extends VFileEvent> events) {
     setContent(ScrollPaneFactory.createScrollPane(graphTable));
 
     // The following listener executes on each opening of the Git Machete tab.
+    // Since we don't want to constantly ask the user about rediscovery the listener is removed after first execution.
+    // Final effect: execution only when user opens Git Machete tab for the first time during an IntelliJ session.
```
PawelLipski, line 75: Hmmm one more concern... what if there are multiple git repositories within the project (see lines 79-80)? won't the rediscovery just trigger for one of the git repositories &mdash; the one that happens to be selected when the Git Machete tab is first opened? or even for **no** repository at all, if none is selected (I don't remember now if it's realistically possible, nvm)

CC @mkondratek who implemented most of multi-git-repository handling

frontend/ui/impl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/root/GitMachetePanel.java
MaciejG604: Yes, it's related. Before this PR the rediscovery suggestion was not proposed because the last edit time of the .git/machete file was set. Now we have more operations before checking the date of the file and this is to ensure that we don't do them constantly.

MaciejG604: Okay, now we do the modification date checking first then check if all local branches are managed.

frontend/ui/impl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/root/GitMachetePanel.java
MaciejG604: After changing the order of checking operations to `Modification Date -> Are local branches managed` I also went back to the previous way reacting to the every opening of Git Machete tab. Now We can state that a suggestion has already been made or is not applicable by changing the modification date of the `.git/machete` file to `System.currentTimeMillis()`. The listener will react everytime so this should cover the multi-git-repo scenario now.

frontend/ui/impl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/RediscoverSuggester.java
```diff
@@ -62,13 +71,49 @@ private void queueSuggestion(Path macheteFilePath) {
...
+  }
+
+  private boolean areAllLocalBranchesManaged(Path macheteFilePath) {
+    val localBranches = gitRepository.getBranches().getLocalBranches();
+    val branchLayoutReader = RuntimeBinding.instantiateSoleImplementingClass(IBranchLayoutReader.class);
+    try {
+      val branchLayout = branchLayoutReader.read(macheteFilePath);
+      val managedLocalBranches = List.ofAll(localBranches)
+          .map(GitReference::getName)
+          .map(branchLayout::findEntryByName);
```
PawelLipski, line 112: One more thing, as a matter of code style: I'd add a method like `boolean hasEntryWithName(String)` to `IBranchLayout`/`BranchLayout`, and use it here via `forAll` instead of `map`+`contains`

frontend/ui/impl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/RediscoverSuggester.java
```diff
@@ -62,13 +71,49 @@ private void queueSuggestion(Path macheteFilePath) {
       queueDiscoverOperation.run();
     } else { // closing dialog goes here too
       LOG.info("Rediscover declined from dialog");
-      macheteFilePath.setFileModificationDate(System.currentTimeMillis());
+      setMacheteFileModificationDate(macheteFilePath, System.currentTimeMillis());
     }
   }
 
+  @UIEffect
+  private void setMacheteFileModificationDate(Path macheteFilePath, long timeMillis) {
```
PawelLipski, line 79: Doesn't need to be named `...Machete...`, this method doesn't need to know that the file being updated is machete file; also, the method can be static, plus the `timeMillis` param is always `System.currentTimeMillis()`

frontend/ui/impl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/RediscoverSuggester.java
```diff
@@ -62,13 +71,49 @@ private void queueSuggestion(Path macheteFilePath) {
...
+    new Task.Backgroundable(
+        gitRepository.getProject(),
+        getString("string.GitMachete.RediscoverSuggester.backgroundable-check-task.title")) {
+      @Override
+      public void run(ProgressIndicator indicator) {
+        if (!areAllLocalBranchesManaged(macheteFilePath)) {
+          ModalityUiUtil.invokeLaterIfNeeded(NON_MODAL, () -> queueSuggestion(macheteFilePath));
+        } else {
+          ModalityUiUtil.invokeLaterIfNeeded(NON_MODAL,
+              () -> setMacheteFileModificationDate(macheteFilePath, System.currentTimeMillis()));
```
PawelLipski, line 99: Oh, so you basically defer the next check until `DAYS_AFTER_WHICH_TO_SUGGEST_DISCOVER` pass right? Pls add a comment in code

PawelLipski (approved):

PawelLipski:
Whoops that was supposed to go in v3.0.2 😅 but okay, let's include in v3.0.1 as well, that's rather a low-risk change

PR #1147, opened 2022-09-20
Feature/dont suggest rediscovery with same result layout

MaciejG604:


branchLayout/api/src/main/java/com/virtuslab/branchlayout/api/IBranchLayoutEntry.java
```diff
@@ -15,4 +18,23 @@ public interface IBranchLayoutEntry {
 
   @Nullable
   String getCustomAnnotation();
+
+  default boolean equals(IBranchLayoutEntry other) {
```
PawelLipski, line 22: Remove `@UsesObjectEquals` from `BranchLayout` and `BranchLayoutEntry`

branchLayout/impl/src/test/java/com/virtuslab/branchlayout/unit/BranchLayoutTestSuite.java
```diff
@@ -190,4 +190,67 @@ public void withBranchSlideOut_givenNonExistingBranch_noExceptionThrown() {
     // then no exception thrown
     Assert.assertTrue(result.getRootEntries().isEmpty());
   }
+
+  @Test
+  public void equalsGivesCorrectResults() {
```
PawelLipski, line 195: Coz it's hard to deduce by a quick look... is there a case where two layouts differ just by **order of children** of certain entry (they should still be equal)?

frontend/ui/impl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/RediscoverSuggester.java
```diff
@@ -108,15 +113,35 @@ public void run(ProgressIndicator indicator) {
...
+    Path rootDirPath = gitRepository.getRootDirectoryPath().toAbsolutePath();
+    Path mainGitDirPath = gitRepository.getMainGitDirectoryPath().toAbsolutePath();
+    Path worktreeGitDirPath = gitRepository.getWorktreeGitDirectoryPath().toAbsolutePath();
+
+    val discoverRunResult = Try.of(() -> RuntimeBinding.instantiateSoleImplementingClass(IGitMacheteRepositoryCache.class)
+        .getInstance(rootDirPath, mainGitDirPath, worktreeGitDirPath).discoverLayoutAndCreateSnapshot());
+
+    if (discoverRunResult.isSuccess()) {
+      try {
+        val managedBranchLayout = branchLayoutReader.read(macheteFilePath);
```
PawelLipski, line 137: I'd call it 
```suggestion
        val currentBranchLayout = branchLayoutReader.read(macheteFilePath);
```
The word `managed` makes sense for _branch_, but not necessarily for the entire layout

branchLayout/impl/src/test/java/com/virtuslab/branchlayout/unit/BranchLayoutTestSuite.java
MaciejG604: Should they be equal? Right now I've defined `IBranchLayoutEntry` equality as name, customAnnotation and children (recursively) are ALL equal. And so for `IBranchLayout` equality means that all rootEntries are equal (in a sense of `IBranchLayoutEntry` equality).

So to answer, if one of the entries of a branch layout will have an extra child branch(es) then it will not be equal to a branch layout that does not have this child branch.

This branch layout: 

```
A
  AA
B
  BA
  BB
```

is not equal to this one:

```
A
  AA
B
  BA
```

PawelLipski: Oh! I don't mean the difference in **structure**, just the difference in **child order**:
```
A
  B
  C
```
vs
```
A
   C
   B
```
AFAICS your implementation of `equals` considers the two layouts equal thanks to sorting the children before comparison right?

MaciejG604: Oh, sorry. Yeah it sorts them by name so order should not be a problem, but I will add a test case for that.

branchLayout/api/src/main/java/com/virtuslab/branchlayout/api/IBranchLayoutEntry.java
```diff
@@ -1,10 +1,15 @@
...
+
 import io.vavr.collection.List;
+import lombok.val;
 import org.checkerframework.checker.nullness.qual.Nullable;
 
 /**
- * The only criterion for equality of any instances of any class implementing this interface is reference equality
+ *  Class that encapsulates a single branch and a list of its children. <br>
+ *  Two IBranchLayoutEntry objects are equal when their names, their custom annotations
+ *  and their children are ALL equal (recursively checked for children).
```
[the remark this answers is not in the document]
MaciejG604: Order is not important, since the equals algorithm first sorts the list of chaild branches by name.

PawelLipski (approved):

PR #1149, opened 2022-09-21
Change property name from headless to virtualDisplay

MaciejG604:


mkondratek (commented):
docker/Dockerfile:
`# Tools necessary to run non-headless UI tests in the screen-less environment of CI`

any chance that we want to reword this one btw?

MaciejG604:
If I understand [THIS](https://virtuslab.github.io/tips/#testing/dummy-linux-display) correctly then it actually IS non-headless, but screen-less. Do you mean reword screen-less to virtual display to keep it consistent?

mkondratek:
yup

PawelLipski:
> Do you mean reword screen-less to virtual display to keep it consistent?
+1, I'd avoid using the word _screen-less_ as well as it's somewhat confusing (if _screen_ == _physical display_ then it's okay, but this "==" isn't obvious)

PawelLipski (approved):

PR #1151, opened 2022-09-21
Use only present tense in change notes

mkondratek:


PawelLipski:
Huh I wouldn't modify the existing change notes tho 🤔 rather won't break anything, but still...

PawelLipski:
As a primitive remedy, I'd ban `ed`-ending words in CHANGE-NOTES since the current release with a script 😅

PawelLipski:
As a primitive remedy, I'd ban `ed`-ending words in CHANGE-NOTES since the current release with a script 😅

mkondratek:
We still have things like 
> Allow checkout and rebase on detach**ed** state
I do not think a script could manage that (well, a simple script ofc)

PawelLipski:
> We still have things like

Meh in fact I'd leave past tense as it is, for the sake of sticking to _local conventions_

PR #1162, opened 2022-09-23
Made where-to-find-the-plugin-tab in README obvious

### LINKED ISSUE #1161: Make `where-to-find-the-plugin-tab` part of README more obvious
For now we have:
> Git Machete IntelliJ Plugin is available under the Git tool window in the Git Machete tab.

Which is true, but could be misunderstood by a user new to IntelliJ IDE plugins. For example (happened to me), one could by mistake think that they should search for the `Git Machete` under `Git` tab on the top of the screen - not on the down of the screen. We can make this part of README more obvious so that new users with similar problem would not get angry.

LukaszKontowski:
Closes #1161

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,8 @@ In that case, just click `Restart IDE` and confirm that action in a message box.
 
 ## Where to find the plugin tab
 
-Git Machete IntelliJ Plugin is available under the `Git` tool window in the `Git Machete` tab. <br/>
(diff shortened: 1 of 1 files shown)


PawelLipski:
For the sake of easier issue/PR management, pls always fill up milestone (v3.0.2 currently, as per `version.gradle.kts`) and labels on issues/PRs

PawelLipski:
Pls link the PR to its issue in Development section

PawelLipski (approved):

PR #1167, opened 2022-09-25
Move to previous next parent

### LINKED ISSUE #1120: Add actions to move to parent/next/previous branch
Similar spirit to `git machete go up/next/prev`.

They need to be available via keyboard shortcuts (now that #645 turned out to be resolved).

This will make it possible to simulate `git machete traverse` entirely from keyboard (with other shortcuts assigned to Rebase, Push etc. actions).

vzmerr:
Resolves #1120

This PR is a draft and for receiving feedback.

The goal is to make it possible to checkout to the up, previous, next branches from the keyboard

Currently, it only has implementation for going to the next branch.
Checking out action is refactored to implement the functionality for the desired branch.

### DIFF
--- a/CHANGE-NOTES.md
+++ b/CHANGE-NOTES.md
@@ -5,6 +5,7 @@
 ## v3.1.0
 - Added support for IntelliJ 2022.3.
 - Fixed spurious `NullPointerException` thrown when opening Machete file, as reported by @oksana-cherniavskaia.
+- Added branch navigation allowing to checking out the next branch with `alt + down-arrow` or `alt + right-arrow`, previous branch with `alt + up-arrow`, and parent branch with `alt + left-arrow` in the machete layout.
 
 ## v3.0.3
 - Enabled Ctrl + left-click to work as a right-click for macOS users, for displaying the machete actions on a branch.
--- a/CHANGE-NOTES.md
+++ b/CHANGE-NOTES.md
@@ -5,7 +5,12 @@
 ## v3.1.0
 - Added support for IntelliJ 2022.3.
 - Fixed spurious `NullPointerException` thrown when opening Machete file, as reported by @oksana-cherniavskaia.
-- Added branch navigation allowing to check out the next branch with `alt + down-arrow` or `alt + right-arrow`, previous branch with `alt + up-arrow`, and parent branch with `alt + left-arrow` when Git Machete tab is opened, and has focus. Note that `option` key is an `alt` key counterpart in macOS.
+- Added branch navigation allowing to check out:
+  * previous branch with `alt + up-arrow`,
+  * next branch with `alt + down-arrow`,
+  * parent branch with `alt + left-arrow`,
+  * first child branch `alt + right-arrow`
+  when Git Machete tab is opened, and has focus. Note that `option` key is an `alt` key counterpart in macOS.
 
 ## v3.0.3
 - Enabled Ctrl + left-click to work as a right-click for macOS users, for displaying the machete actions on a branch.
--- a/CHANGE-NOTES.md
+++ b/CHANGE-NOTES.md
@@ -5,7 +5,7 @@
 ## v3.1.0
 - Added support for IntelliJ 2022.3.
 - Fixed spurious `NullPointerException` thrown when opening Machete file, as reported by @oksana-cherniavskaia.
-- Added branch navigation allowing to checking out the next branch with `alt + down-arrow` or `alt + right-arrow`, previous branch with `alt + up-arrow`, and parent branch with `alt + left-arrow` in the machete layout.
+- Added branch navigation allowing to check out the next branch with `alt + down-arrow` or `alt + right-arrow`, previous branch with `alt + up-arrow`, and parent branch with `alt + left-arrow` when Git Machete tab is opened, and has focus.
 
 ## v3.0.3
 - Enabled Ctrl + left-click to work as a right-click for macOS users, for displaying the machete actions on a branch.
--- a/CHANGE-NOTES.md
+++ b/CHANGE-NOTES.md
@@ -4,13 +4,15 @@
 
 ## v3.1.0
 - Added support for IntelliJ 2022.3.
-- Fixed spurious `NullPointerException` thrown when opening Machete file, as reported by @oksana-cherniavskaia.
+- Fixed spurious `NullPointerException` thrown when opening machete file via toolbar, as reported by @oksana-cherniavskaia.
 - Added branch navigation allowing to check out:
   * previous branch with `alt + up-arrow`,
   * next branch with `alt + down-arrow`,
   * parent branch with `alt + left-arrow`,
-  * first child branch `alt + right-arrow`
-  when Git Machete tab is opened, and has focus. Note that `option` key is an `alt` key counterpart in macOS.
+  * first child branch with `alt + right-arrow`
+  when Git Machete tab is opened, and has focus.
+
+  Note that `option` key is the equivalent of `alt` in macOS.
 
 ## v3.0.3
 - Enabled Ctrl + left-click to work as a right-click for macOS users, for displaying the machete actions on a branch.
--- a/CHANGE-NOTES.md
+++ b/CHANGE-NOTES.md
@@ -5,7 +5,7 @@
 ## v3.1.0
 - Added support for IntelliJ 2022.3.
 - Fixed spurious `NullPointerException` thrown when opening Machete file, as reported by @oksana-cherniavskaia.
-- Added branch navigation allowing to check out the next branch with `alt + down-arrow` or `alt + right-arrow`, previous branch with `alt + up-arrow`, and parent branch with `alt + left-arrow` when Git Machete tab is opened, and has focus.
+- Added branch navigation allowing to check out the next branch with `alt + down-arrow` or `alt + right-arrow`, previous branch with `alt + up-arrow`, and parent branch with `alt + left-arrow` when Git Machete tab is opened, and has focus. You would need to use the `option` key instead of `alt` in MacOS.
 
 ## v3.0.3
 - Enabled Ctrl + left-click to work as a right-click for macOS users, for displaying the machete actions on a branch.
--- a/CHANGE-NOTES.md
+++ b/CHANGE-NOTES.md
@@ -5,7 +5,7 @@
 ## v3.1.0
 - Added support for IntelliJ 2022.3.
 - Fixed spurious `NullPointerException` thrown when opening Machete file, as reported by @oksana-cherniavskaia.
-- Added branch navigation allowing to check out the next branch with `alt + down-arrow` or `alt + right-arrow`, previous branch with `alt + up-arrow`, and parent branch with `alt + left-arrow` when Git Machete tab is opened, and has focus. You would need to use the `option` key instead of `alt` in MacOS.
+- Added branch navigation allowing to check out the next branch with `alt + down-arrow` or `alt + right-arrow`, previous branch with `alt + up-arrow`, and parent branch with `alt + left-arrow` when Git Machete tab is opened, and has focus. Note that `option` key is an `alt` key counterpart in macOS.
 
 ## v3.0.3
 - Enabled Ctrl + left-click to work as a right-click for macOS users, for displaying the machete actions on a branch.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -226,7 +226,7 @@ Other coding conventions include:
 * Properties in `GitMacheteBundle.properties` that use HTML should be wrapped in tags `<html>` ... `</html>`.
   Additionally, their keys should have a `.HTML` suffix.
 * `@Tainted` and `@Untainted` annotations are used in the context of method parameters that may or may not use HTML. Those annotated with `@Untainted` should not contain HTML tags, whereas values annotated with
-  `@Tainted` can contain HTML (but they don't have to).
+  `@Tainted` can contain HTML (but they don't have to). Please note that you need to use `org.checkerframework.checker.tainting.qual.Tainted` or `org.checkerframework.checker.tainting.qual.Untainted` annotations, rather than the `javax.annotation` ones.
 * Avoid `Branch` word in action class names and action ids to keep them shorter.
   Some exceptions are allowed (e.g. the backgroundable task classes).
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -113,12 +113,13 @@ Local (non-CI) builds by default skip most of [Checker Framework's](https://chec
 To make local builds more aligned with CI builds (at the expense of ~2x longer compilation from scratch),
 set `runAllCheckers` Gradle project property (e.g. `./gradlew -PrunAllCheckers build`).
 
-In case of spurious cache-related issues with Gradle build, try one of the following:
-* `./gradlew --stop` to shut down gradle daemon
+In case of spurious cache-related issues with Gradle build, try the following remedies (in the order from the least intrusive to the most):
+* `./gradlew --stop` to shut down Gradle daemon
+* `pkill -e -9 -f '.*Gradle.*'` to kill all Gradle processes
 * `./gradlew clean` and re-run the failing `./gradlew` command with `--no-build-cache`
 * remove .gradle/ directory in the project directory
 * remove ~/.gradle/caches/ (or even the entire ~/.gradle/) directory
-* reinstall gradle AND remove the entire ~/.gradle/ directory
+* reinstall Gradle AND remove the entire ~/.gradle/ directory
 
 
 ## Run
--- a/docs/features.md
+++ b/docs/features.md
@@ -34,8 +34,14 @@ The current branch is underlined in a branch layout.
 
 ![](checkout.gif)
 
-On the Git Machete tab, you can checkout the next branch with `alt + down-arrow` or `alt + right-arrow`, previous branch with `alt + up-arrow`, and parent branch with `alt + left-arrow` in the machete layout.
+If Git Machete tab is opened & focused, branch navigation with keyboard shortcuts is available.
+Check out:
+* previous branch with `alt + up-arrow`,
+* next branch with `alt + down-arrow`,
+* parent branch with `alt + left-arrow`,
+* first child branch with `alt + right-arrow`.
 
+Note that `option` key is the equivalent of `alt` in macOS.
 
 ## Toggle listing commits
 
--- a/docs/features.md
+++ b/docs/features.md
@@ -34,6 +34,8 @@ The current branch is underlined in a branch layout.
 
 ![](checkout.gif)
 
+On the Git Machete tab, you can checkout the next branch with `alt + down-arrow` or `alt + right-arrow`, previous branch with `alt + up-arrow`, and parent branch with `alt + left-arrow` in the machete layout.
+
 
 ## Toggle listing commits
 
--- a/frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseCheckoutAction.java
+++ b/frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseCheckoutAction.java
@@ -1,0 +1,110 @@
+package com.virtuslab.gitmachete.frontend.actions.base;
+
+import static com.virtuslab.gitmachete.frontend.resourcebundles.GitMacheteBundle.getNonHtmlString;
+import static com.virtuslab.gitmachete.frontend.resourcebundles.GitMacheteBundle.getString;
+
+import java.util.Collections;
+
+import com.intellij.openapi.actionSystem.AnActionEvent;
+import com.intellij.openapi.progress.ProgressIndicator;
+import com.intellij.openapi.progress.Task;
+import com.intellij.openapi.project.Project;
+import git4idea.branch.GitBranchUiHandlerImpl;
+import git4idea.branch.GitBranchWorker;
+import git4idea.commands.Git;
+import git4idea.repo.GitRepository;
+import kr.pe.kwonnam.slf4jlambda.LambdaLogger;
+import lombok.CustomLog;
+import lombok.experimental.ExtensionMethod;
+import lombok.val;
+import org.checkerframework.checker.guieffect.qual.UIEffect;
+import org.checkerframework.checker.nullness.qual.Nullable;
+import org.checkerframework.checker.tainting.qual.Untainted;
+
+import com.virtuslab.gitmachete.frontend.actions.expectedkeys.IExpectsKeySelectedBranchName;
+import com.virtuslab.gitmachete.frontend.resourcebundles.GitMacheteBundle;
+import com.virtuslab.gitmachete.frontend.vfsutils.GitVfsUtils;
+import com.virtuslab.qual.guieffect.UIThreadUnsafe;
+
+@ExtensionMethod({GitVfsUtils.class, GitMacheteBundle.class})
+@CustomLog
+public abstract class BaseCheckoutAction extends BaseGitMacheteRepositoryReadyAction
+    implements
+      IExpectsKeySelectedBranchName {
+
+  @Override
+  public LambdaLogger log() {
+    return LOG;
+  }
+
+  protected abstract @Nullable String getTargetBranchName(AnActionEvent anActionEvent);
+
+  protected abstract @Untainted String getNonExistentBranchMessage(AnActionEvent anActionEvent);
+
+  @Override
+  @UIEffect
+  protected void onUpdate(AnActionEvent anActionEvent) {
+    super.onUpdate(anActionEvent);
+
+    val presentation = anActionEvent.getPresentation();
+    if (!presentation.isEnabledAndVisible()) {
+      return;
+    }
+
+    val targetBranchName = getTargetBranchName(anActionEvent);
+
+    // It's very unlikely that targetBranchName is empty at this point since it's assigned directly before invoking this
+    // action in EnhancedGraphTable.EnhancedGraphTableMouseAdapter#mouseClicked; still, it's better to be safe.
+    if (targetBranchName == null || targetBranchName.isEmpty()) {
+      presentation.setEnabled(false);
+      presentation.setDescription(getNonExistentBranchMessage(anActionEvent));
+      return;
+    }
+
+    val currentBranchName = getCurrentBranchNameIfManaged(anActionEvent);
+
+    if (currentBranchName != null && currentBranchName.equals(targetBranchName)) {
+      presentation.setEnabled(false);
+      presentation.setDescription(
+          getNonHtmlString("action.GitMachete.BaseCheckoutAction.description.disabled.currently-checked-out")
+              .format(targetBranchName));
+
+    } else {
+      presentation.setDescription(
+          getNonHtmlString("action.GitMachete.BaseCheckoutAction.description.precise")
+              .format(targetBranchName));
+    }
+  }
+
+  @Override
+  @UIEffect
+  public void actionPerformed(AnActionEvent anActionEvent) {
+    val targetBranchName = getTargetBranchName(anActionEvent);
+    if (targetBranchName == null || targetBranchName.isEmpty()) {
+      return;
+    }
+
+    val project = getProject(anActionEvent);
+    val gitRepository = getSelectedGitRepository(anActionEvent);
+
+    if (gitRepository != null) {
+
+      log().debug(() -> "Queuing '${targetBranchName}' branch checkout background task");
+      new Task.Backgroundable(project, getString("action.GitMachete.BaseCheckoutAction.task-title")) {
+        @Override
+        @UIThreadUnsafe
+        public void run(ProgressIndicator indicator) {
+          doCheckout(project, indicator, targetBranchName, gitRepository);
+        }
+      }.queue();
+    }
+  }
+
+  @UIThreadUnsafe
+  public static void doCheckout(Project project, ProgressIndicator indicator, String branchToCheckoutName,
+      GitRepository gitRepository) {
+    val uiHandler = new GitBranchUiHandlerImpl(project, indicator);
+    new GitBranchWorker(project, Git.getInstance(), uiHandler)
+        .checkout(branchToCheckoutName, /* detach */ false, Collections.singletonList(gitRepository));
+  }
+}
--- a/frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/navigation/CheckoutPreviousAction.java
+++ b/frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/navigation/CheckoutPreviousAction.java
@@ -1,0 +1,39 @@
+package com.virtuslab.gitmachete.frontend.actions.navigation;
+
+import static com.virtuslab.gitmachete.frontend.resourcebundles.GitMacheteBundle.getNonHtmlString;
+
+import com.intellij.openapi.actionSystem.AnActionEvent;
+import lombok.experimental.ExtensionMethod;
+import lombok.val;
+import org.checkerframework.checker.nullness.qual.Nullable;
+import org.checkerframework.checker.tainting.qual.Untainted;
+
+import com.virtuslab.gitmachete.frontend.actions.base.BaseCheckoutAction;
+import com.virtuslab.gitmachete.frontend.actions.expectedkeys.IExpectsKeySelectedBranchName;
+import com.virtuslab.gitmachete.frontend.resourcebundles.GitMacheteBundle;
+
+@ExtensionMethod(GitMacheteBundle.class)
+public class CheckoutPreviousAction extends BaseCheckoutAction
+    implements
+      IExpectsKeySelectedBranchName {
+
+  @Override
+  protected @Untainted String getNonExistentBranchMessage(AnActionEvent anActionEvent) {
+    val currentBranchName = getCurrentBranchNameIfManaged(anActionEvent);
+    return currentBranchName != null
+        ? getNonHtmlString("action.GitMachete.CheckoutPreviousAction.undefined.branch-name").format(currentBranchName)
+        : getNonHtmlString("action.GitMachete.BaseCheckoutAction.undefined.current-branch");
+  }
+
+  @Override
+  protected @Nullable String getTargetBranchName(AnActionEvent anActionEvent) {
+    val currentBranchName = getCurrentBranchNameIfManaged(anActionEvent);
+    val branchLayout = getBranchLayout(anActionEvent);
+    if (branchLayout != null && currentBranchName != null) {
+      val previousEntry = branchLayout.findPreviousEntry(currentBranchName);
+      return previousEntry != null ? previousEntry.getName() : null;
+    }
+    return null;
+  }
+
+}
--- a/frontend/resourcebundles/src/main/resources/GitMacheteBundle.properties
+++ b/frontend/resourcebundles/src/main/resources/GitMacheteBundle.properties
@@ -441,3 +441,7 @@ action.GitMachete.CheckoutPreviousAction.undefined.branch-name=No branches befor
 action.GitMachete.CheckoutParentAction.text=Git Machete: Checkout Parent Branch
 action.GitMachete.CheckoutParentAction.description=Checkout parent branch (''{0}'')
 action.GitMachete.CheckoutParentAction.undefined.branch-name=Branch (''{0}'') does not have a parent in the branch layout
+
+action.GitMachete.CheckoutFirstChildAction.text=Git Machete: Checkout First Child Branch
+action.GitMachete.CheckoutFirstChildAction.description=Checkout first child branch (''{0}'')
+action.GitMachete.CheckoutFirstChildAction.undefined.branch-name=Branch (''{0}'') does not have any children in the branch layout
--- a/frontend/resourcebundles/src/main/resources/GitMacheteBundle.properties
+++ b/frontend/resourcebundles/src/main/resources/GitMacheteBundle.properties
@@ -194,12 +194,12 @@ action.GitMachete.ISyncToRemoteStatusDependentAction.description.enabled={0} bra
 
 action.GitMachete.CheckoutSelectedAction.text=Git Machete: Checkout Selected Branch
 action.GitMachete.CheckoutSelectedAction.GitMacheteContextMenu.text=_Checkout
-action.GitMachete.CheckoutSelectedAction.description=Checkout this branch
 action.GitMachete.CheckoutSelectedAction.undefined.branch-name=Checkout disabled due to undefined branch name
 
-action.GitMachete.CheckoutSelectedAction.description.disabled.currently-checked-out=Branch ''{0}'' is currently checked out
-action.GitMachete.CheckoutSelectedAction.description.precise=Checkout branch ''{0}''
-action.GitMachete.CheckoutSelectedAction.task-title=Checking out
+action.GitMachete.BaseCheckoutAction.description.disabled.currently-checked-out=Branch ''{0}'' is currently checked out
+action.GitMachete.BaseCheckoutAction.description.precise=Checkout branch ''{0}''
+action.GitMachete.BaseCheckoutAction.task-title=Checking out
+action.GitMachete.BaseCheckoutAction.undefined.current-branch=There is no current branch; the repository is probably in detached HEAD state
 
 
 action.GitMachete.FastForwardMergeSelectedToParentAction.text=Git Machete: Fast-forward Merge into Parent
@@ -427,3 +427,17 @@ action.GitMachete.OpenMacheteTabAction.description=Open Git Machete tab
 
 action.GitMachete.OpenMacheteTabAction.notification.title.could-not-open-tab.HTML=<html><b>Could not open Git Machete tab</b></html>
 action.GitMachete.OpenMacheteTabAction.notification.message.no-git=Ensure that Git VCS is enabled in the project and there is at least one local branch present.
+
+# Branch Navigation
+
+action.GitMachete.CheckoutNextAction.text=Git Machete: Checkout Next Branch
+action.GitMachete.CheckoutNextAction.description=Checkout next branch (''{0}'')
+action.GitMachete.CheckoutNextAction.undefined.branch-name=No branches after (''{0}'') in the branch layout
+
+action.GitMachete.CheckoutPreviousAction.text=Git Machete: Checkout Previous Branch
+action.GitMachete.CheckoutPreviousAction.description=Checkout previous branch (''{0}'')
+action.GitMachete.CheckoutPreviousAction.undefined.branch-name=No branches before (''{0}'') in the branch layout
+
+action.GitMachete.CheckoutParentAction.text=Git Machete: Checkout Parent Branch
+action.GitMachete.CheckoutParentAction.description=Checkout parent branch (''{0}'')
+action.GitMachete.CheckoutParentAction.undefined.branch-name=Branch (''{0}'') does not have a parent in the branch layout
--- a/src/main/resources/META-INF/plugin.xml
+++ b/src/main/resources/META-INF/plugin.xml
@@ -220,7 +220,6 @@
         <action id="GitMachete.CheckoutNextAction"
                 class="com.virtuslab.gitmachete.frontend.actions.navigation.CheckoutNextAction">
             <keyboard-shortcut keymap="$default" first-keystroke="alt DOWN" />
-            <keyboard-shortcut keymap="$default" first-keystroke="alt RIGHT" />
         </action>
 
         <action id="GitMachete.CheckoutPreviousAction"
@@ -228,9 +227,13 @@
             <keyboard-shortcut keymap="$default" first-keystroke="alt UP" />
         </action>
 
-        <action id="GitMachete.ParentBranchCheckoutAction"
+        <action id="GitMachete.CheckoutParentAction"
                 class="com.virtuslab.gitmachete.frontend.actions.navigation.CheckoutParentAction">
             <keyboard-shortcut keymap="$default" first-keystroke="alt LEFT" />
+        </action>
+        <action id="GitMachete.CheckoutFirstChildAction"
+                class="com.virtuslab.gitmachete.frontend.actions.navigation.CheckoutFirstChildAction">
+            <keyboard-shortcut keymap="$default" first-keystroke="alt RIGHT" />
         </action>
     </actions>
 
--- a/src/main/resources/META-INF/plugin.xml
+++ b/src/main/resources/META-INF/plugin.xml
@@ -216,6 +216,22 @@
             <!-- In $default keymap `control` key (Windows/Linux) corresponds to `cmd` key on macOS -->
             <keyboard-shortcut keymap="$default" first-keystroke="control alt shift M" />
         </action>
+
+        <action id="GitMachete.CheckoutNextAction"
+                class="com.virtuslab.gitmachete.frontend.actions.navigation.CheckoutNextAction">
+            <keyboard-shortcut keymap="$default" first-keystroke="alt DOWN" />
+            <keyboard-shortcut keymap="$default" first-keystroke="alt RIGHT" />
+        </action>
+
+        <action id="GitMachete.CheckoutPreviousAction"
+                class="com.virtuslab.gitmachete.frontend.actions.navigation.CheckoutPreviousAction">
+            <keyboard-shortcut keymap="$default" first-keystroke="alt UP" />
+        </action>
+
+        <action id="GitMachete.ParentBranchCheckoutAction"
+                class="com.virtuslab.gitmachete.frontend.actions.navigation.CheckoutParentAction">
+            <keyboard-shortcut keymap="$default" first-keystroke="alt LEFT" />
+        </action>
     </actions>
 
 </idea-plugin>
--- a/src/uiTest/resources/project.rhino.js
+++ b/src/uiTest/resources/project.rhino.js
@@ -119,6 +119,7 @@ function Project(underlyingProject) {
 
   const ACTION_PLACE_TOOLBAR = 'GitMacheteToolbar';
   const ACTION_PLACE_CONTEXT_MENU = 'GitMacheteContextMenu';
+  const ACTION_PLACE_EMPTY = '';
   const SHOW_RESET_INFO = 'git-machete.reset.info.show';
   const SHOW_MERGE_WARNING = 'git-machete.merge.warning.show';
 
@@ -400,6 +401,18 @@ function Project(underlyingProject) {
 
   this.checkoutBranch = function (branchName) {
     invokeActionAndWait('GitMachete.CheckoutSelectedAction', ACTION_PLACE_CONTEXT_MENU, { SELECTED_BRANCH_NAME: branchName });
+  };
+
+  this.checkoutNextBranch = function () {
+    invokeActionAndWait('GitMachete.CheckoutNextAction', ACTION_PLACE_EMPTY, {});
+  };
+
+  this.checkoutPreviousBranch = function () {
+    invokeActionAndWait('GitMachete.CheckoutPreviousAction', ACTION_PLACE_EMPTY, {});
+  };
+
+  this.checkoutParentBranch = function () {
+    invokeActionAndWait('GitMachete.ParentBranchCheckoutAction', ACTION_PLACE_EMPTY, {});
   };
 
   this.fastForwardMergeSelectedToParent = function (branchName) {
--- a/src/uiTest/resources/project.rhino.js
+++ b/src/uiTest/resources/project.rhino.js
@@ -403,6 +403,10 @@ function Project(underlyingProject) {
     invokeActionAndWait('GitMachete.CheckoutSelectedAction', ACTION_PLACE_CONTEXT_MENU, { SELECTED_BRANCH_NAME: branchName });
   };
 
+  this.checkoutFirstChildBranch = function () {
+    invokeActionAndWait('GitMachete.CheckoutFirstChildAction', ACTION_PLACE_EMPTY, {});
+  };
+
   this.checkoutNextBranch = function () {
     invokeActionAndWait('GitMachete.CheckoutNextAction', ACTION_PLACE_EMPTY, {});
   };
--- a/src/uiTest/resources/project.rhino.js
+++ b/src/uiTest/resources/project.rhino.js
@@ -412,7 +412,7 @@ function Project(underlyingProject) {
   };
 
   this.checkoutParentBranch = function () {
-    invokeActionAndWait('GitMachete.ParentBranchCheckoutAction', ACTION_PLACE_EMPTY, {});
+    invokeActionAndWait('GitMachete.CheckoutParentAction', ACTION_PLACE_EMPTY, {});
   };
 
   this.fastForwardMergeSelectedToParent = function (branchName) {
--- a/src/uiTest/scala/com/virtuslab/gitmachete/uitest/RunningIntelliJFixtureExtension.scala
+++ b/src/uiTest/scala/com/virtuslab/gitmachete/uitest/RunningIntelliJFixtureExtension.scala
@@ -131,6 +131,10 @@ trait RunningIntelliJFixtureExtension extends RobotPluginExtension { this: IdePr
         runJs(s"project.checkoutBranch('$branch')")
       }
 
+      def checkoutFirstChildBranch(): Unit = doAndAwait {
+        runJs("project.checkoutFirstChildBranch()")
+      }
+
       def checkoutNextBranch(): Unit = doAndAwait {
         runJs("project.checkoutNextBranch()")
       }
--- a/src/uiTest/scala/com/virtuslab/gitmachete/uitest/RunningIntelliJFixtureExtension.scala
+++ b/src/uiTest/scala/com/virtuslab/gitmachete/uitest/RunningIntelliJFixtureExtension.scala
@@ -131,6 +131,18 @@ trait RunningIntelliJFixtureExtension extends RobotPluginExtension { this: IdePr
         runJs(s"project.checkoutBranch('$branch')")
       }
 
+      def checkoutNextBranch(): Unit = doAndAwait {
+        runJs("project.checkoutNextBranch()")
+      }
+
+      def checkoutPreviousBranch(): Unit = doAndAwait {
+        runJs("project.checkoutPreviousBranch()")
+      }
+
+      def checkoutParentBranch(): Unit = doAndAwait {
+        runJs("project.checkoutParentBranch()")
+      }
+
       def configure(): Unit = {
         runJs("project.configure()")
       }
--- a/src/uiTest/scala/com/virtuslab/gitmachete/uitest/UITestSuite.scala
+++ b/src/uiTest/scala/com/virtuslab/gitmachete/uitest/UITestSuite.scala
@@ -91,6 +91,14 @@ class UITestSuite extends BaseGitRepositoryBackedIntegrationTestSuite(SETUP_WITH
     // 7 branch rows + 11 commit rows
     Assert.assertEquals(18, branchAndCommitRowsCount)
 
+    project.checkoutBranch("build-chain")
+    project.checkoutPreviousBranch()
+    Assert.assertEquals(project.getCurrentBranchName(), "update-icons")
+    project.checkoutNextBranch()
+    Assert.assertEquals(project.getCurrentBranchName(), "build-chain")
+    project.checkoutParentBranch()
+    Assert.assertEquals(project.getCurrentBranchName(), "allow-ownership-link")
+
     // Let's slide out a root branch now
     project.slideOutSelected("develop")
     project.acceptBranchDeletionOnSlideOut()
--- a/src/uiTest/scala/com/virtuslab/gitmachete/uitest/UITestSuite.scala
+++ b/src/uiTest/scala/com/virtuslab/gitmachete/uitest/UITestSuite.scala
@@ -40,7 +40,7 @@ class UITestSuite extends BaseGitRepositoryBackedIntegrationTestSuite(SETUP_WITH
     waitAndCloseProject()
   }
 
-  val _saveThreadDumpWhenTestFailed = new TestWatcher() {
+  val _saveThreadDumpWhenTestFailed: TestWatcher = new TestWatcher() {
     override protected def failed(e: Throwable, description: Description): Unit = {
       val pid = probe.pid()
       probe.screenshot("exception")
--- a/src/uiTest/scala/com/virtuslab/gitmachete/uitest/UITestSuite.scala
+++ b/src/uiTest/scala/com/virtuslab/gitmachete/uitest/UITestSuite.scala
@@ -105,14 +105,13 @@ class UITestSuite extends BaseGitRepositoryBackedIntegrationTestSuite(SETUP_WITH
     project.slideOutSelected("develop")
     project.acceptBranchDeletionOnSlideOut()
     branchAndCommitRowsCount = project.refreshModelAndGetRowCount()
-    // 5 branch rows (`develop` is no longer there) + 7 commit rows
+    // 6 branch rows (`develop` is no longer there) + 7 commit rows
     // (1 commit of `allow-ownership-link` and 3 commits of `call-ws` are all gone)
     Assert.assertEquals(13, branchAndCommitRowsCount)
 
     project.checkoutBranch("master")
     project.slideOutSelected("call-ws")
     project.rejectBranchDeletionOnSlideOut()
-    branchAndCommitRowsCount = project.refreshModelAndGetRowCount()
     val managedBranchesAfterSlideOut = project.refreshModelAndGetManagedBranches()
     // Non-existent branches should be skipped while causing no error (only a low-severity notification).
     Assert.assertEquals(
@@ -125,7 +124,8 @@ class UITestSuite extends BaseGitRepositoryBackedIntegrationTestSuite(SETUP_WITH
       ),
       managedBranchesAfterSlideOut.toSeq.sorted
     )
-    // 4 branch rows (`call-ws` is also no longer there) + 8 commit rows
+    branchAndCommitRowsCount = project.refreshModelAndGetRowCount()
+    // 5 branch rows (`call-ws` is also no longer there) + 7 commit rows
     Assert.assertEquals(12, branchAndCommitRowsCount)
   }
 
--- a/src/uiTest/scala/com/virtuslab/gitmachete/uitest/UITestSuite.scala
+++ b/src/uiTest/scala/com/virtuslab/gitmachete/uitest/UITestSuite.scala
@@ -59,7 +59,6 @@ class UITestSuite extends BaseGitRepositoryBackedIntegrationTestSuite(SETUP_WITH
   def saveThreadDumpWhenTestFailed: TestWatcher = _saveThreadDumpWhenTestFailed
 
   @Test def skipNonExistentBranches_toggleListingCommits_slideOutRoot(): Unit = {
-    project.openGitMacheteTab()
     overwriteMacheteFile(
       """develop
         |  non-existent
@@ -72,6 +71,7 @@ class UITestSuite extends BaseGitRepositoryBackedIntegrationTestSuite(SETUP_WITH
         |master
         |  hotfix/add-trigger""".stripMargin
     )
+    project.openGitMacheteTab()
     val managedBranches = project.refreshModelAndGetManagedBranches()
     // Non-existent branches should be skipped while causing no error (only a low-severity notification).
     Assert.assertEquals(
--- a/src/uiTest/scala/com/virtuslab/gitmachete/uitest/UITestSuite.scala
+++ b/src/uiTest/scala/com/virtuslab/gitmachete/uitest/UITestSuite.scala
@@ -64,8 +64,8 @@ class UITestSuite extends BaseGitRepositoryBackedIntegrationTestSuite(SETUP_WITH
       """develop
         |  non-existent
         |    allow-ownership-link
-        |      update-icons
         |      build-chain
+        |      update-icons
         |    non-existent-leaf
         |  call-ws
         |non-existent-root
@@ -91,10 +91,12 @@ class UITestSuite extends BaseGitRepositoryBackedIntegrationTestSuite(SETUP_WITH
     // 7 branch rows + 11 commit rows
     Assert.assertEquals(18, branchAndCommitRowsCount)
 
-    project.checkoutBranch("build-chain")
-    project.checkoutPreviousBranch()
-    Assert.assertEquals(project.getCurrentBranchName(), "update-icons")
+    project.checkoutBranch("allow-ownership-link")
+    project.checkoutFirstChildBranch()
+    Assert.assertEquals(project.getCurrentBranchName(), "build-chain")
     project.checkoutNextBranch()
+    Assert.assertEquals(project.getCurrentBranchName(), "update-icons")
+    project.checkoutPreviousBranch()
     Assert.assertEquals(project.getCurrentBranchName(), "build-chain")
     project.checkoutParentBranch()
     Assert.assertEquals(project.getCurrentBranchName(), "allow-ownership-link")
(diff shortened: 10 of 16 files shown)


src/main/resources/META-INF/plugin.xml
```diff
@@ -36,6 +36,12 @@
                 <override-text place="GitMacheteContextMenu"/>
             </action>
 
+            <action id="GitMachete.NextBranchCheckoutAction"
+                    class="com.virtuslab.gitmachete.frontend.actions.contextmenu.NextBranchCheckoutAction">
+                <override-text place="GitMacheteContextMenu"/>
```
PawelLipski, line 41: TBD whether these actions should even be in the context menu... actually it'd only make sense if they have predefined keyboard shortcuts, and then the menu will serve as kinda-tip for the users that such checkout actions and their keyboard shortcuts are available at all

PawelLipski: CC @mkondratek

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/contextmenu/NextBranchCheckoutAction.java
```diff
@@ -0,0 +1,38 @@
...
+import com.virtuslab.gitmachete.frontend.actions.expectedkeys.IExpectsKeySelectedBranchName;
+import com.virtuslab.gitmachete.frontend.resourcebundles.GitMacheteBundle;
+
+@ExtensionMethod(GitMacheteBundle.class)
+public class NextBranchCheckoutAction extends BaseCheckoutAction
+    implements
+      IExpectsKeySelectedBranchName {
+
+  @Override
+  protected @Nullable String getSubjectBranchName(AnActionEvent anActionEvent) {
```
PawelLipski, line 18: Maybe 
```suggestion
  protected @Nullable String getTargetBranchName(AnActionEvent anActionEvent) {
```

src/main/resources/META-INF/plugin.xml
PawelLipski: Note that as for now we've got just one predefined keyboard shortcut: `Ctrl+Alt+Shift+M` for opening the Git Machete tab

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/contextmenu/NextBranchCheckoutAction.java
```diff
@@ -0,0 +1,38 @@
...
+    val gitRepository = getSelectedGitRepository(anActionEvent);
+    val currentBranchName = getCurrentBranchNameIfManaged(anActionEvent);
+    if (gitRepository != null) {
+      val localBranches = gitRepository.getBranches().getLocalBranches().iterator();
+      while (localBranches.hasNext()) {
+        val nextBranch = localBranches.next();
+        if (nextBranch.getName() == currentBranchName) {
```
PawelLipski, line 25: Whoops `equals`!

PawelLipski: Oh, Interning Checker captured that 💪🏻

PawelLipski: Interning Checker is a bit PITA (with a all that esoteric annotations like `@UsesObjectEquals` but captures incorrectly used `==` which is awesome

src/main/resources/META-INF/plugin.xml
vzmerr: Thank you Pawel, so if this is the desired functionality, I would implement `previous` and `up` too.

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/contextmenu/NextBranchCheckoutAction.java
PawelLipski: Note that it doesn't trigger locally: https://github.com/VirtusLab/git-machete-intellij-plugin/blob/master/CONTRIBUTING.md#build -> `-PrunAllCheckers`

src/main/resources/META-INF/plugin.xml
PawelLipski: Yes, the general direction is okay, details TBD

PawelLipski: Pls think over the possible combinations of keyboard shortcuts 🤔 for Mac/non-Mac

vzmerr: Finding key combinations on Mac is very difficult. Whatever key combination I use, it says, it is already taken. except shift+Letter sometimes.

PawelLipski: Note that many shortcuts are taken by somewhat niche actions (limited to the scope of a specific plugin etc.)... ofc we shouldn't try taking over any well-known, global shortcuts (like, `Ctrl+Shift+Alt+N` for _Next branch_ would be a very bad idea 😅)

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseCheckoutAction.java
```diff
@@ -0,0 +1,108 @@
...
+    super.onUpdate(anActionEvent);
+
+    val presentation = anActionEvent.getPresentation();
+    if (!presentation.isEnabledAndVisible()) {
+      return;
+    }
+
+    val toBeCheckedOutBranchName = getTargetBranchName(anActionEvent);
+
+    // It's very unlikely that selectedBranchName is empty at this point since it's assigned directly before invoking this
```
vzmerr, line 53:
```suggestion
    // It's very unlikely that targetBranchName is empty at this point since it's assigned directly before invoking this
```

src/main/resources/META-INF/plugin.xml
vzmerr: Ctrl+shift+option+M does nothing for me :(

PawelLipski: On Mac it's ⌘⌥⇧M: https://github.com/VirtusLab/git-machete-intellij-plugin/#where-to-find-the-plugin-tab

PawelLipski:
As discussed, context menu is not a good place for keeping this actions in general (as they always act relative to the **current** branch, not to the **selected** branch).

To keep this PR self-contained before we proceed with PoC of Traverse, let's remove these actions from context menu, and assign them some keyboard shortcuts. A minor mention in docs/features.md + update to CHANGE-NOTES.md would also be useful.

src/main/resources/META-INF/plugin.xml
```diff
@@ -227,6 +212,27 @@
             <!-- In $default keymap `control` key (Windows/Linux) corresponds to `cmd` key on macOS -->
             <keyboard-shortcut keymap="$default" first-keystroke="control alt shift M" />
         </action>
+
+        <action id="GitMachete.NextBranchCheckoutAction"
+                class="com.virtuslab.gitmachete.frontend.actions.traverse.NextBranchCheckoutAction">
+            <add-to-group group-id="GitRepositoryActions" anchor="last"/>
```
PawelLipski, line 218: Hmmm is `GitRepositoryActions` some group that we control? or is it some existing group from git4idea? 🤔

src/main/resources/META-INF/plugin.xml
```diff
@@ -227,6 +212,27 @@
...
+            <keyboard-shortcut keymap="$default" first-keystroke="control alt shift T" second-keystroke="control shift L" />
+            <!--  currently, the second key stroke is not getting triggered. suggestions welcome-->
+        </action>
+
+        <action id="GitMachete.PreviousBranchCheckoutAction"
+                class="com.virtuslab.gitmachete.frontend.actions.traverse.PreviousBranchCheckoutAction">
+            <add-to-group group-id="GitRepositoryActions" anchor="last"/>
+            <keyboard-shortcut keymap="$default" first-keystroke="control alt shift T" second-keystroke="shift B" />
```
PawelLipski, line 226: Huh `Ctrl Alt Shift T` is somewhat unfortunate as it opens PRs on Linux 😅 what's the Mac equivalent? (you need to be logged in to GitHub in IntelliJ probably to see that tool window tho)

PawelLipski: How about `Ctrl Alt Shift G` however? Doesn't seem to correspond to anything important (`Tx Isolation` in `Database` menu xD). 

Note the comment in line 212 ofc (`<!-- In $default keymap `control` key (Windows/Linux) corresponds to `cmd` key on macOS -->`)

src/main/resources/META-INF/plugin.xml
vzmerr: The `OpenGitMacheteTab` actions is as well added to that. Don't we need it here?

src/main/resources/META-INF/plugin.xml
```diff
@@ -227,6 +212,27 @@
             <!-- In $default keymap `control` key (Windows/Linux) corresponds to `cmd` key on macOS -->
             <keyboard-shortcut keymap="$default" first-keystroke="control alt shift M" />
         </action>
+
+        <action id="GitMachete.NextBranchCheckoutAction"
+                class="com.virtuslab.gitmachete.frontend.actions.traverse.NextBranchCheckoutAction">
+            <add-to-group group-id="GitRepositoryActions" anchor="last"/>
+            <keyboard-shortcut keymap="$default" first-keystroke="control alt shift T" second-keystroke="control shift L" />
```
PawelLipski, line 219: I'd make the second stroke more consistent... what are the limitations actually? are you free to use anything for the second stroke, or are there sequence that are gonna be captured as _first_ stroke anyway? 🤔 like, if you use `F1` for second stroke, will it be captured as second stroke, or first stroke of single-stroke F1 shortcut (Help, by default)?

vzmerr: The only limitation is that, the second stroke should not overlap with any existing key shortcut. such as ctrl+C

src/main/resources/META-INF/plugin.xml
vzmerr: Perhaps, for a more convenient use, we can put, `Ctrl+alt+Shift+G+N` for next `Ctrl+alt+Shift+G+P` for Previous
`Ctrl+alt+Shift+G+U` for up?

vzmerr: Unfortunately, not allowed :(

src/main/resources/META-INF/plugin.xml
vzmerr: with `G` it was not getting triggered

src/main/resources/META-INF/plugin.xml
mkondratek: we can try vim direction keys (jk) 😆 

image

PawelLipski: > we can try vim direction keys (jk) laughing

Huh nope 😅 I mean, even ASDF would be better ;D knowledge of vim among the developers (not counting the devops/sysadmin crowd) is rather low... for most users, this choice will look rather random/arbitrary ☹️

PawelLipski: TBH I myself despite using `vim` quite extensively back in early 2010s, I've never got used to HJKL... always the arrows

src/main/resources/META-INF/plugin.xml
PawelLipski: Huh @mkondratek do you remember what's the actual effect of having a `add-to-group` for `GitRepositoryActions`? I.e. where is the action displayed (that it otherwise wouldn't be)? 🤔

src/main/resources/META-INF/plugin.xml
PawelLipski: > Perhaps, for a more convenient use, we can put, Ctrl+alt+Shift+G+N for next Ctrl+alt+Shift+G+P for Previous
Ctrl+alt+Shift+G+U for up?

That would be awesome... okay so what was specifically not working? Is `Ctrl+Alt+Shift+G` somehow forbidden as first keystroke of double-stroke shortcut? In general, if you could research & post what combinations are allowed in double-stroke shortcuts (maybe it's documented somewhere by JetBrains?)

src/main/resources/META-INF/plugin.xml
PawelLipski: Aww okay I see now hmmm

src/main/resources/META-INF/plugin.xml
PawelLipski: Ok, seen the response below, nvm ;D

src/main/resources/META-INF/plugin.xml
PawelLipski: Okay so assuming the second stroke overlaps with an existing shortcut... what specifically happens? The existing single-stroke shortcut takes precedence (and e.g. Copy action is triggered instead of our double-stroke action)?

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseCheckoutAction.java
```diff
@@ -0,0 +1,108 @@
...
+    if (gitRepository != null) {
+
+      log().debug(() -> "Queuing '${toBeCheckedOutBranchName}' branch checkout background task");
+      new Task.Backgroundable(project, getString("action.GitMachete.CheckoutSelectedAction.task-title")) {
+        @Override
+        @UIThreadUnsafe
+        public void run(ProgressIndicator indicator) {
+          doCheckout(project, indicator, toBeCheckedOutBranchName, gitRepository);
+        }
+        // TODO (#95): on success, refresh only indication of the current branch
```
PawelLipski, line 96: Remove the TODO while you're here, the issue is closed as `wontfix` already

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/traverse/NextBranchCheckoutAction.java
```diff
@@ -0,0 +1,38 @@
...
+      IExpectsKeySelectedBranchName {
+
+  @Override
+  protected @Nullable String getTargetBranchName(AnActionEvent anActionEvent) {
+    val gitRepository = getSelectedGitRepository(anActionEvent);
+    val currentBranchName = getCurrentBranchNameIfManaged(anActionEvent);
+    if (gitRepository != null) {
+      val localBranches = gitRepository.getBranches().getLocalBranches().iterator();
+      while (localBranches.hasNext()) {
+        val nextBranch = localBranches.next();
```
PawelLipski, line 24: Whoops it should check out the next branch in the list of branches returned by `gitRepository.getBranches().getLocalBranches()`... but what we aim for is to check out the next branch in the **branch layout** as managed by git-machete!

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/traverse/PreviousBranchCheckoutAction.java
```diff
@@ -0,0 +1,38 @@
...
+public class PreviousBranchCheckoutAction extends BaseCheckoutAction
+    implements
+      IExpectsKeySelectedBranchName {
+
+  @Override
+  protected @Nullable String getTargetBranchName(AnActionEvent anActionEvent) {
+    val gitRepository = getSelectedGitRepository(anActionEvent);
+    val currentBranchName = getCurrentBranchNameIfManaged(anActionEvent);
+    if (gitRepository != null) {
+      val localBranches = gitRepository.getBranches().getLocalBranches().iterator();
```
PawelLipski, line 22: Same problem as with `NextBranchCheckoutAction` ☹️

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/traverse/NextBranchCheckoutAction.java
PawelLipski: Actually! that's a good moment to cover these new actions with some UI tests. I'd replace some calls to `project.checkoutBranch(...)` with `project.checkout{Next,Previous,Parent}Branch()` which would call your newly-added actions. `project.checkout...Branch()` method would need to be added in both Scala and Rhino JS side.

Before the work on UI tests, pls make sure you've got all the useful IntelliJ plugins enabled/installed, as per CONTRIBUTING.md

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseCheckoutAction.java
vzmerr, line 96:
```suggestion
```

src/main/resources/META-INF/plugin.xml
vzmerr: > Copy action is triggered on Ctrl+C instead of our action

Yes.

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseCheckoutAction.java
```diff
@@ -0,0 +1,131 @@
...
+  @UIEffect
+  protected void onUpdate(AnActionEvent anActionEvent) {
+    super.onUpdate(anActionEvent);
+
+    val presentation = anActionEvent.getPresentation();
+    if (!presentation.isEnabledAndVisible()) {
+      return;
+    }
+
+    val toBeCheckedOutBranchName = getTargetBranchName(anActionEvent);
```
PawelLipski, line 58: For consistency
```suggestion
    val targetBranchName = getTargetBranchName(anActionEvent);
```

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseCheckoutAction.java
```diff
@@ -0,0 +1,131 @@
...
+
+    // It's very unlikely that selectedBranchName is empty at this point since it's assigned directly before invoking this
```
PawelLipski, line 60:
```suggestion
    // It's very unlikely that targetBranchName is empty at this point since it's assigned directly before invoking this
```

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseCheckoutAction.java
```diff
@@ -0,0 +1,131 @@
...
+    // action in GitMacheteGraphTable.GitMacheteGraphTableMouseAdapter.mouseClicked; still, it's better to be safe.
```
PawelLipski, line 61:
```suggestion
    // action in EnhancedGraphTable.EnhancedGraphTableMouseAdapter#mouseClicked; still, it's better to be safe.
```

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseCheckoutAction.java
```diff
@@ -0,0 +1,131 @@
...
+    if (toBeCheckedOutBranchName == null || toBeCheckedOutBranchName.isEmpty()) {
+      presentation.setEnabled(false);
+      presentation.setDescription(getNonHtmlString("action.GitMachete.CheckoutSelectedAction.undefined.branch-name"));
```
PawelLipski, line 64: Pls update the class name in property keys accordingly

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseCheckoutAction.java
```diff
@@ -0,0 +1,131 @@
...
+      presentation.setDescription(
+          getNonHtmlString("action.GitMachete.CheckoutSelectedAction.description.precise")
+              .format(toBeCheckedOutBranchName));
+    }
+  }
+
+  @Override
+  @UIEffect
+  public void actionPerformed(AnActionEvent anActionEvent) {
+    val toBeCheckedOutBranchName = getTargetBranchName(anActionEvent);
```
PawelLipski, line 86:
```suggestion
    val targetBranchName = getTargetBranchName(anActionEvent);
```

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseCheckoutAction.java
```diff
@@ -0,0 +1,131 @@
...
+    new GitBranchWorker(project, Git.getInstance(), uiHandler)
+        .checkout(branchToCheckoutName, /* detach */ false, Collections.singletonList(gitRepository));
+  }
+
+  protected Stream<String> deriveLayoutBranches(AnActionEvent anActionEvent) {
+    val gitRepository = getSelectedGitRepository(anActionEvent);
+    if (gitRepository != null) {
+      val macheteFilePath = gitRepository.getMacheteFilePath();
+      try {
+        val macheteFileContent = Files.readString(macheteFilePath, StandardCharsets.UTF_8);
```
PawelLipski, line 120: I wouldn't operate directly on the file contents... pls use `BranchLayoutReader`, see e.g. `com.virtuslab.gitmachete.frontend.ui.impl.RediscoverSuggester`

src/uiTest/resources/project.rhino.js
```diff
@@ -402,6 +402,18 @@ function Project(underlyingProject) {
     invokeActionAndWait('GitMachete.CheckoutSelectedAction', ACTION_PLACE_CONTEXT_MENU, { SELECTED_BRANCH_NAME: branchName });
   };
 
+  this.checkoutNextBranch = function() {
+      invokeActionAndWait('GitMachete.NextBranchCheckoutAction');
```
PawelLipski, line 406: Nit: pls use two spaces for indent

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseCheckoutAction.java
vzmerr: It produces the branchLayout. I tried that initially. The algorithm for jumping to the next branch using the branchLayout is recursive and longer, so I was unsure of keeping it.

1. see if the branch has children
2. if so go to the first child,
3. else see if the current branch is nonRoot,
4. if so, take the children of the parent,
5. find the next sibling,
6. if there is no sibling next, 
7. see if the parent is nonRoot
8. if so get the grandparent
9. get the children of the grandparent
10. find the next sibling of the parent
11. if there is no sibling for the parent, repeat from 7 for grandparent.

src/main/resources/META-INF/plugin.xml
mkondratek: oh, this one is interesting - gimme a moment 🕵️

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseCheckoutAction.java
vzmerr: The next alternative I would say is to add a method: `getNextBranch(branch)` to the branchLayout, and actually in the  branchLayout, either keep an ordered list of all branches in layout, or keep their order.

src/main/resources/META-INF/plugin.xml
mkondratek: ok, I see know. `GitRepositoryActions` used to be the action group available from menu `Git`. 
**Now, OpenMacheteTabAction is NOT available in that menu.**
git4idea must have change it in the meantime. We even was expecting that but forgot it; we have the comment 
```
```

I ve tested it, and the following:
```
 <add-to-group group-id="VcsGroups" anchor="last"/>
```
works fine now - as expected.

I am opening a separate issue for this one as the lack of our action in Git menu is a bug.

vzmerr: So I guess this implies that I can remove `addToGroup` from the actions of this PR.

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseCheckoutAction.java
vzmerr: It's just that in each update, we have to also update the list, which I think won't be that much of a deal.

PawelLipski: Hmmm lemme think... actually, the recursive algorithm will probably be cleaner than keeping extra piece of state in each node, just for the purpose of these actions.

I'd suggest the following:
* put this PR on hold for a moment,
* implement the algorithm of `getNext` and `getPrevious` in `BranchLayout` according to https://github.com/VirtusLab/git-machete-intellij-plugin/pull/1167#discussion_r982266997 on a separate PR, together with tests
* use the new methods to cleanly implement `{Previous,Next}BranchCheckoutAction`

WDYT?

vzmerr: > implement the algorithm of getNext and getPrevious in BranchLayout 

using the recursive algorithm or the list of layout branches?

vzmerr: > the recursive algorithm will probably be cleaner than keeping extra piece of state in each node

ok, got it. so I would go with recursive.

src/main/resources/META-INF/plugin.xml
```diff
@@ -211,6 +211,21 @@
...
+
+        <action id="GitMachete.PreviousBranchCheckoutAction"
+                class="com.virtuslab.gitmachete.frontend.actions.traverse.PreviousBranchCheckoutAction">
+            <keyboard-shortcut keymap="$default" first-keystroke="control alt shift B" />
+        </action>
+
+        <action id="GitMachete.ParentBranchCheckoutAction"
+                class="com.virtuslab.gitmachete.frontend.actions.traverse.ParentBranchCheckoutAction">
+            <keyboard-shortcut keymap="$default" first-keystroke="control alt shift U" />
+        </action>
```
vzmerr, line 228: Meanwhile, can you please do me a favour by trying these key combinations on Linux or Windows?

PawelLipski: Lemme check, as I'm probably the sole non-Mac (Linux) user around 😅

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseCheckoutAction.java
```diff
@@ -0,0 +1,131 @@
+package com.virtuslab.gitmachete.frontend.actions.base;
```
PawelLipski, line 1: Btw, pls also add an entry to CHANGE-NOTES.md and bump version.gradle.kts to 3.1.0 (as this constitutes a new user-facing feature, hence a MINOR bump is necessary)

src/main/resources/META-INF/plugin.xml
PawelLipski: They work like a charm 🥰 still, TBD what shortcuts to use... actually, how about the F<number> buttons? they seem extremely underused in IntelliJ...  other than the standard Alt+F4, F1 some F7-F10 for build/run/debug, and maybe Alt+F12 for listing class members, I can't think of any

src/uiTest/resources/project.rhino.js
```diff
@@ -402,6 +402,18 @@ function Project(underlyingProject) {
     invokeActionAndWait('GitMachete.CheckoutSelectedAction', ACTION_PLACE_CONTEXT_MENU, { SELECTED_BRANCH_NAME: branchName });
   };
 
+  this.checkoutNextBranch = function() {
+    invokeActionAndWait('GitMachete.NextBranchCheckoutAction', ACTION_PLACE_CONTEXT_MENU,{});
```
PawelLipski, line 406: To be checked if e.g. an empty string can be used as `actionPlace` (as `null` is forbidden, the underlying IntelliJ API method has a `@NonNull` param)

src/main/resources/META-INF/plugin.xml
vzmerr: In Mac, the function keys do a lot of code navigation. Option + Function keys types strange characters.

vzmerr: In Mac, Option is Alt

PawelLipski: @mkondratek any ideas here?

src/uiTest/resources/project.rhino.js
```diff
@@ -402,6 +402,18 @@ function Project(underlyingProject) {
     invokeActionAndWait('GitMachete.CheckoutSelectedAction', ACTION_PLACE_CONTEXT_MENU, { SELECTED_BRANCH_NAME: branchName });
   };
 
+  this.checkoutNextBranch = function() {
+    invokeActionAndWait('GitMachete.NextBranchCheckoutAction', "",{});
```
PawelLipski, line 406: Nit
```suggestion
    invokeActionAndWait('GitMachete.NextBranchCheckoutAction', /* actionPlace */ "", {});
```
Here and in the others

src/uiTest/scala/com/virtuslab/gitmachete/uitest/RunningIntelliJFixtureExtension.scala
```diff
@@ -131,6 +131,18 @@ trait RunningIntelliJFixtureExtension extends RobotPluginExtension { this: IdePr
         runJs(s"project.checkoutBranch('$branch')")
       }
 
+      def checkoutNextBranch(): Unit = doAndAwait {
+        runJs(s"project.checkoutNextBranch()")
```
PawelLipski, line 135: No need for `s"..."`

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseCheckoutAction.java
```diff
@@ -0,0 +1,131 @@
...
+      return;
+    }
+
+    val project = getProject(anActionEvent);
+    val gitRepository = getSelectedGitRepository(anActionEvent);
+
+    if (gitRepository != null) {
+
+      log().debug(() -> "Queuing '${targetBranchName}' branch checkout background task");
+      new Task.Backgroundable(project, getString("action.GitMachete.CheckoutSelectedAction.task-title")) {
```
PawelLipski, line 97: There are more propertykeys that are to be renamed (not specific to `CheckoutSelectedAction`, but shared b/w all Checkout actions)

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/traverse/NextBranchCheckoutAction.java
```diff
@@ -0,0 +1,27 @@
...
+import lombok.experimental.ExtensionMethod;
+import lombok.val;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+import com.virtuslab.gitmachete.frontend.actions.base.BaseCheckoutAction;
+import com.virtuslab.gitmachete.frontend.actions.expectedkeys.IExpectsKeySelectedBranchName;
+import com.virtuslab.gitmachete.frontend.resourcebundles.GitMacheteBundle;
+
+@ExtensionMethod(GitMacheteBundle.class)
+public class NextBranchCheckoutAction extends BaseCheckoutAction
```
PawelLipski, line 14: For consistency:
```suggestion
public class CheckoutNextAction extends BaseCheckoutAction
```
We skip `Branch` from action names, https://github.com/VirtusLab/git-machete-intellij-plugin/blob/master/CONTRIBUTING.md#coding-conventions -> last bullet

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/traverse/PreviousBranchCheckoutAction.java
```diff
@@ -0,0 +1,28 @@
...
+import lombok.experimental.ExtensionMethod;
+import lombok.val;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+import com.virtuslab.gitmachete.frontend.actions.base.BaseCheckoutAction;
+import com.virtuslab.gitmachete.frontend.actions.expectedkeys.IExpectsKeySelectedBranchName;
+import com.virtuslab.gitmachete.frontend.resourcebundles.GitMacheteBundle;
+
+@ExtensionMethod(GitMacheteBundle.class)
+public class PreviousBranchCheckoutAction extends BaseCheckoutAction
```
PawelLipski, line 14: Ditto, rename

src/main/resources/META-INF/plugin.xml
mkondratek: what's wrong with arrow keys?

vzmerr: they navigate in the text instead, even with ctrl+alt+shift

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseCheckoutAction.java
```diff
@@ -0,0 +1,131 @@
...
+        @UIThreadUnsafe
+        public void run(ProgressIndicator indicator) {
+          doCheckout(project, indicator, targetBranchName, gitRepository);
+        }
+      }.queue();
+    }
+  }
+
+  @UIThreadUnsafe
+  public static void doCheckout(Project project, @NonNull ProgressIndicator indicator, String branchToCheckoutName,
```
mkondratek, line 108: `@NonNull` - do we need it?

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/traverse/PreviousBranchCheckoutAction.java
```diff
@@ -0,0 +1,28 @@
...
+    implements
+      IExpectsKeySelectedBranchName {
+
+  @Override
+  protected @Nullable String getTargetBranchName(AnActionEvent anActionEvent) {
+    val currentBranchName = getCurrentBranchNameIfManaged(anActionEvent);
+    return Option
```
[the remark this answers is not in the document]
mkondratek: Do we have some alternatives? Singleton set/list/smth?

mkondratek (changes requested):

src/main/resources/META-INF/plugin.xml
mkondratek: well yes, but no

mkondratek: I think we are overthinking this one. let's go for a simple **and known** solution - the arrow. 

They navigate through the editor - that's true unless the focus is elsewhere. 
It can be in Project overview - where we can navigate b/w files.
It can be in Git Log - which is almost exactly the application that we need. 
Try arrow navigation in Git Log for yourself - it's extremely easy and clear.

Ofc we will lose the potential possibility to navigate branches from anywhere (probably without git machete tab opened) but it makes things way more complicated and less readable. Let's stick to good and simple solution.

mkondratek: for me, the more complicated key combination is, the less likely I will use it. 

I cannot remember using the shortcut for opening git machete tab for purposes other than testing this shortcut - this key, that key, stand on one leg, touch your nose, use different nose on Mac... who remembers that? 😆

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseCheckoutAction.java
```diff
@@ -0,0 +1,131 @@
...
+          getNonHtmlString("action.GitMachete.BaseCheckoutAction.description.precise")
+              .format(targetBranchName));
+    }
+  }
+
+  @Override
+  @UIEffect
+  public void actionPerformed(AnActionEvent anActionEvent) {
+    val targetBranchName = getTargetBranchName(anActionEvent);
+    if (targetBranchName == null || targetBranchName.isEmpty()) {
```
mkondratek, line 87: I think we need proper descriptions in this case - to handle situations when there is no next/prev/parent

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseCheckoutAction.java
```diff
@@ -0,0 +1,131 @@
...
+      return;
+    }
+
+    val targetBranchName = getTargetBranchName(anActionEvent);
+
+    // It's very unlikely that targetBranchName is empty at this point since it's assigned directly before invoking this
+    // action in EnhancedGraphTable.EnhancedGraphTableMouseAdapter#mouseClicked; still, it's better to be safe.
+    if (targetBranchName == null || targetBranchName.isEmpty()) {
+      presentation.setEnabled(false);
+      presentation.setDescription(getNonHtmlString("action.GitMachete.BaseCheckoutAction.undefined.branch-name"));
```
mkondratek, line 64: I think we need proper descriptions in this case - to handle situations when there is no next/prev/parent. I mean, we should precisely say
`Branch 'master' has no parent branch`, etc

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/traverse/PreviousBranchCheckoutAction.java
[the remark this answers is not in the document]
vzmerr: there is indeed another PR for getting next/prev branches in the BranchLayout. Once that get merged, we are going to use that instead of this.

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseCheckoutAction.java
vzmerr: I guess this was the preexisting code, I refactored from our `CheckoutSelectedAction`. I would try to see if the `@NonNull` is still needed.

src/uiTest/scala/com/virtuslab/gitmachete/uitest/UITestSuite.scala
```diff
@@ -98,8 +98,8 @@ class UITestSuite extends BaseGitRepositoryBackedIntegrationTestSuite(SETUP_WITH
     // 5 branch rows (`develop` is no longer there) + 7 commit rows
     // (1 commit of `allow-ownership-link` and 3 commits of `call-ws` are all gone)
     Assert.assertEquals(13, branchAndCommitRowsCount)
-
-    project.checkoutBranch("master")
+    project.checkoutBranch("allow-ownership-link")
+    project.checkoutParentBranch()
```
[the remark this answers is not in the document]
PawelLipski: 'twas my idea 😅 but okay, let's keep all changes within `skipNonExistentBranches_toggleListingCommits_slideOutRoot` indeed

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseCheckoutAction.java
PawelLipski: Almost surely not needed, non-nullness is always assumed by default by Checker

src/main/resources/META-INF/plugin.xml
PawelLipski: > Ofc we will lose the potential possibility to navigate branches from anywhere (probably without git machete tab opened) but it makes things way more complicated and less readable

Wise advice indeed, +1

Just some research is needed if it's easy/doable at all to make Git Machete tab capture the arrows  🤔 

> I cannot remember using the shortcut for opening git machete tab for purposes other than testing this shortcut - this key, that key, stand on one leg, touch your nose, use a different nose on Mac... who remembers that?

I do, all the time :trollface: but I'm keyboard shortcut addict basically (I've even set up a shortcut to install a plugin from disk ;p)

frontend/base/src/main/java/com/virtuslab/gitmachete/frontend/defs/ActionPlaces.java
```diff
@@ -6,4 +6,6 @@ private ActionPlaces() {}
   public static final String ACTION_PLACE_CONTEXT_MENU = "GitMacheteContextMenu";
   public static final String ACTION_PLACE_TOOLBAR = "GitMacheteToolbar";
   public static final String ACTION_PLACE_VCS_NOTIFICATION = "Vcs.Notification";
+
+  public static final String ACTION_PLACE_EMPTY = "";
```
PawelLipski, line 10: Seems unused, can be removed (UI tests have their own constant)

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/navigation/CheckoutPreviousAction.java
```diff
@@ -0,0 +1,28 @@
...
+    implements
+      IExpectsKeySelectedBranchName {
+
+  @Override
+  protected @Nullable String getTargetBranchName(AnActionEvent anActionEvent) {
+    val currentBranchName = getCurrentBranchNameIfManaged(anActionEvent);
+    return Option
+        .ofOptional(deriveLayoutBranches(anActionEvent)
+            .takeWhile(branchName -> !branchName.equals(currentBranchName))
+            .reduce((first, second) -> second))
```
PawelLipski, line 24: Isn't there some method like `last()`? 🤔

vzmerr: no, unfortunately. But this is going to get removed when the get-next-prev-parent gets merged.

frontend/base/src/main/java/com/virtuslab/gitmachete/frontend/defs/ActionPlaces.java
vzmerr, line 10:
```suggestion
```

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseCheckoutAction.java
vzmerr: There are just some issues with the `@Untainted` requirement.

frontend/resourcebundles/src/main/resources/GitMacheteBundle.properties
```diff
@@ -433,9 +434,12 @@ action.GitMachete.OpenMacheteTabAction.notification.message.no-git=Ensure that G
 
 action.GitMachete.CheckoutNextAction.text=Git Machete: Checkout Next Branch
 action.GitMachete.CheckoutNextAction.description=Checkout next branch (''{0}'')
+action.GitMachete.CheckoutNextAction.undefined.branch-name=No branches after (''{0}'') in the machete layout
```
PawelLipski, line 437: We call it 
```suggestion
action.GitMachete.CheckoutNextAction.undefined.branch-name=No branches after (''{0}'') in the branch layout
```
`machete layout` is clear as well, but let's be consistent

src/main/resources/META-INF/plugin.xml
vzmerr: well, I was wondering if someone has git-machete tab open, wouldn't it be much handier to just click on the needed branch for checking out?

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseCheckoutAction.java
PawelLipski: Hmmm they might justified... what's spefically the erorr? we use `@(Un)Tainted` annotations to manage HTML formatting in properties (as not all places in IntelliJ UIs allow for HTML tags)

frontend/resourcebundles/src/main/resources/GitMacheteBundle.properties
```diff
@@ -192,14 +192,16 @@ action.GitMachete.ISyncToRemoteStatusDependentAction.description.enabled={0} bra
 
 # Context menu
 
+
```
PawelLipski, line 195: Remove

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseCheckoutAction.java
```diff
@@ -0,0 +1,132 @@
...
+      IExpectsKeySelectedBranchName {
+
+  @Override
+  public LambdaLogger log() {
+    return LOG;
+  }
+
+  protected abstract @Nullable String getTargetBranchName(AnActionEvent anActionEvent);
+
+  protected abstract String getNonExistingBranchMessage(AnActionEvent anActionEvent);
```
vzmerr, line 47: @PawelLipski the result of this is identified as `@Tainted`

vzmerr: I think it might be due to the `.format` method or due to method call.

frontend/resourcebundles/src/main/resources/GitMacheteBundle.properties
```diff
@@ -192,14 +192,16 @@ action.GitMachete.ISyncToRemoteStatusDependentAction.description.enabled={0} bra
...
+
```
vzmerr, line 196:
```suggestion
```

src/main/resources/META-INF/plugin.xml
PawelLipski: In general: yes, that's long implemented (double click checks out the branch).
ut
The aim of this entire task however, is to provide a gradual base for traverse-like feature: not sure if the (semi)-automatic traversal can reasonably be implemented in this plugin, and how it's gonna look like... but providing a way to jump b/w branches without clicking looks like a first good step to figuring out the best UX for traversal.

Hence, I'm willing to approve this PR even if **there's no built-in way to access the checkout {next,prev,parent}** actions. It still has the research value for traverse. However, if an easy&natural way to trigger these actions can be figured on this PR (like arrows?)... then I'm more than happy to take it :)

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseCheckoutAction.java
PawelLipski: Okay... and then the value is passed is to a place where an `@Untainted` value is expected?
In our interpretation:
* `@Tainted` = maybe has HTML
* `@Untainted` = surely does **not** have HTML

src/main/resources/META-INF/plugin.xml
vzmerr: So let me see what I can do with the arrows 👍

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseCheckoutAction.java
vzmerr: > the value is passed is to a place where an @Untainted value is expected?

Yes, the `presentation.setDescription(value)`

Now, all the current implementations of the method only return `nonHtmlString`

PawelLipski: Now as a remedy, pls mark the return type of `getNonExistingBranchMessage` as `@Untainted` (non-HTML) as well (might be needed in both base class, and in concrete impls).

Post the compiler (Checker) errors in case of problems.

vzmerr: Yes, and I tried adding `@Untainted` tag to the return type of `getNonExistingBranchMessage` and all the implementations, to further assist the compile, process. But still, the same problem.

PawelLipski: Note there is a short mention of _`@Tainted` and `@Untainted` annotations_ in CONTRIBUTING.md. If you think there's anything you can add there which might be useful for the future devs, feel free to include such change in this PR

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseCheckoutAction.java
```diff
@@ -10,6 +10,8 @@
 import java.util.Collections;
 import java.util.stream.Stream;
 
+import javax.annotation.Untainted;
```
PawelLipski, line 13: Use checkerframework annotations. I'm not sure if Taintedness Checker recognizes these ones (Nullness Checker surely recognizes a lot of `@Nullable`/`@NonNull` annotations from other libraries)

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseCheckoutAction.java
PawelLipski: > Yes, and I tried adding @Untainted tag to the return type of getNonExistingBranchMessage and all the implementations, to further assist the compile, process. But still, the same problem.

Probably wrong import (from `javax`). Lmk if the problem persists

CONTRIBUTING.md
```diff
@@ -223,7 +223,7 @@ Other coding conventions include:
 * Properties in `GitMacheteBundle.properties` that use HTML should be wrapped in tags `<html>` ... `</html>`.
   Additionally, their keys should have a `.HTML` suffix.
 * `@Tainted` and `@Untainted` annotations are used in the context of method parameters that may or may not use HTML. Those annotated with `@Untainted` should not contain HTML tags, whereas values annotated with
-  `@Tainted` can contain HTML (but they don't have to).
+  `@Tainted` can contain HTML (but they don't have to). Please note that if needed, you might need to use the` @Tainted` or `@Untainted` tags from the Checker Framework, rather than the Java Annotations ones.
```
PawelLipski, line 226:
```suggestion
  `@Tainted` can contain HTML (but they don't have to). Please note that you need to use the` @Tainted` or `@Untainted` annotations from the Checker Framework, rather than the Java Annotations ones.
```

PawelLipski: Actually, pls use FQCNs for clarity

src/main/resources/META-INF/plugin.xml
vzmerr: The alt+up, alt+right, alt+left arrows are working for me on MacOS. Can you please do me a favour by trying them on a different operating system?

vzmerr: I mean `Option`.
Now if the machete tab is open and on focus, the arrow keys work for branch navigation.

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/navigation/CheckoutPreviousAction.java
PawelLipski: Yes, or even better, you can make this PR based on the get-next-prev-parent PR, for example using Git Machete 😁

src/main/resources/META-INF/plugin.xml
PawelLipski: Lemme check on Linux; Mikołaj pls double-verify on MacOS

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/navigation/CheckoutPreviousAction.java
PawelLipski: Yes! exactly

branchLayout/api/src/main/java/com/virtuslab/branchlayout/api/BranchLayout.java
```diff
@@ -29,6 +30,28 @@ private static List<IBranchLayoutEntry> collectEntriesRecursively(IBranchLayoutE
     return entryByName.get(branchName).getOrNull();
   }
 
+  @Override
+  public @Nullable IBranchLayoutEntry findNextEntry(String branchName) {
```
PawelLipski, line 34: Change the base of this PR to `get-next-prev-branch`. Then once you merge the PR #1184, this PR #1167 will automatically retarget to `develop`

src/main/resources/META-INF/plugin.xml
```diff
@@ -211,6 +211,21 @@
             <!-- In $default keymap `control` key (Windows/Linux) corresponds to `cmd` key on macOS -->
             <keyboard-shortcut keymap="$default" first-keystroke="control alt shift M" />
         </action>
+
+        <action id="GitMachete.CheckoutNextAction"
+                class="com.virtuslab.gitmachete.frontend.actions.navigation.CheckoutNextAction">
+            <keyboard-shortcut keymap="$default" first-keystroke="alt RIGHT" />
```
PawelLipski, line 217: Works for me on Linux! Albeit I'd use down/up for prev/next, and left for parent

PawelLipski: Right can also be for next, same as down

branchLayout/api/src/main/java/com/virtuslab/branchlayout/api/BranchLayout.java
PawelLipski: (there's also `git machete github retarget-pr` in CLI that sets the base of the PR in GitHub to what's in your local .git/machete)

CONTRIBUTING.md
```diff
@@ -223,7 +223,7 @@ Other coding conventions include:
 * Properties in `GitMacheteBundle.properties` that use HTML should be wrapped in tags `<html>` ... `</html>`.
   Additionally, their keys should have a `.HTML` suffix.
 * `@Tainted` and `@Untainted` annotations are used in the context of method parameters that may or may not use HTML. Those annotated with `@Untainted` should not contain HTML tags, whereas values annotated with
-  `@Tainted` can contain HTML (but they don't have to). Please note that you need to use `org.checkerframework.checker.tainting.qual.Untainted` or `org.checkerframework.checker.tainting.qual.Untainted` annotations, rather than the `java.annotations` ones.
+  `@Tainted` can contain HTML (but they don't have to). Please note that you need to use `org.checkerframework.checker.tainting.qual.Untainted` or `org.checkerframework.checker.tainting.qual.Untainted` annotations, rather than the `javax.annotations` ones.
```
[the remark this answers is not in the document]
vzmerr: Now both of them are `Untainted`

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseCheckoutAction.java
```diff
@@ -0,0 +1,133 @@
...
+      IExpectsKeySelectedBranchName {
+
+  @Override
+  public LambdaLogger log() {
+    return LOG;
+  }
+
+  protected abstract @Nullable String getTargetBranchName(AnActionEvent anActionEvent);
+
+  protected abstract @Untainted String getNonExistingBranchMessage(AnActionEvent anActionEvent);
```
PawelLipski, line 48: Nit: 
```suggestion
  protected abstract @Untainted String getNonExistentBranchMessage(AnActionEvent anActionEvent);
```

src/main/resources/META-INF/plugin.xml
vzmerr: Now because next in the layout is down, and prev is up, I set them as such to be intuitive.

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseCheckoutAction.java
```diff
@@ -0,0 +1,133 @@
...
+
+  @UIThreadUnsafe
+  public static void doCheckout(Project project, ProgressIndicator indicator, String branchToCheckoutName,
+      GitRepository gitRepository) {
+    val uiHandler = new GitBranchUiHandlerImpl(project, indicator);
+    new GitBranchWorker(project, Git.getInstance(), uiHandler)
+        .checkout(branchToCheckoutName, /* detach */ false, Collections.singletonList(gitRepository));
+  }
+
+  protected Stream<String> deriveLayoutBranches(AnActionEvent anActionEvent) {
```
PawelLipski, line 117: To be replaced with `BranchLayout` usage

src/main/resources/META-INF/plugin.xml
PawelLipski: Ouch, right, I meant _down/up for **next/prev**_, not _prev/next_

frontend/resourcebundles/src/main/resources/GitMacheteBundle.properties
```diff
@@ -194,12 +194,12 @@ action.GitMachete.ISyncToRemoteStatusDependentAction.description.enabled={0} bra
...
-action.GitMachete.CheckoutSelectedAction.description=Checkout this branch
 action.GitMachete.CheckoutSelectedAction.undefined.branch-name=Checkout disabled due to undefined branch name
 
-action.GitMachete.CheckoutSelectedAction.description.disabled.currently-checked-out=Branch ''{0}'' is currently checked out
-action.GitMachete.CheckoutSelectedAction.description.precise=Checkout branch ''{0}''
-action.GitMachete.CheckoutSelectedAction.task-title=Checking out
+action.GitMachete.BaseCheckoutAction.description.disabled.currently-checked-out=Branch ''{0}'' is currently checked out
+action.GitMachete.BaseCheckoutAction.description.precise=Checkout branch ''{0}''
+action.GitMachete.BaseCheckoutAction.task-title=Checking out
+action.GitMachete.BaseCheckoutAction.undefined.current-branch=There is no current branch
```
PawelLipski, line 202:
```suggestion
action.GitMachete.BaseCheckoutAction.undefined.current-branch=There is no current branch; the repository is probably in detached HEAD state
```

frontend/resourcebundles/src/main/resources/GitMacheteBundle.properties
```diff
@@ -427,3 +427,17 @@ action.GitMachete.OpenMacheteTabAction.description=Open Git Machete tab
...
+action.GitMachete.CheckoutNextAction.description=Checkout next branch (''{0}'')
+action.GitMachete.CheckoutNextAction.undefined.branch-name=No branches after (''{0}'') in the branch layout
+
+action.GitMachete.CheckoutPreviousAction.text=Git Machete: Checkout Previous Branch
+action.GitMachete.CheckoutPreviousAction.description=Checkout previous branch (''{0}'')
+action.GitMachete.CheckoutPreviousAction.undefined.branch-name=No branches before (''{0}'') in the branch layout
+
+action.GitMachete.CheckoutParentAction.text=Git Machete: Checkout Parent Branch
+action.GitMachete.CheckoutParentAction.description=Checkout parent branch (''{0}'')
+action.GitMachete.CheckoutParentAction.undefined.branch-name=branch (''{0}'') does not have a parent in the branch layout
```
PawelLipski, line 443: Initial cap?
```suggestion
action.GitMachete.CheckoutParentAction.undefined.branch-name=Branch (''{0}'') does not have a parent in the branch layout
```

CHANGE-NOTES.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+## v3.1.0
+- Added branch navigation allowing to checking out the next, previous, and parent branches in the machete layout, with `ctrl+alt+shift+G`, `ctrl+alt+shift+B`, and `ctrl+alt+shift+U`, respectively.
```
PawelLipski, line 6: Pls update accordingly

CHANGE-NOTES.md
```diff
@@ -2,6 +2,9 @@
...
+
```
PawelLipski, line 7: Pls also update  docs/features.md (no need to add screenshots, just text is ok)

src/main/resources/META-INF/plugin.xml
```diff
@@ -211,6 +211,21 @@
             <!-- In $default keymap `control` key (Windows/Linux) corresponds to `cmd` key on macOS -->
             <keyboard-shortcut keymap="$default" first-keystroke="control alt shift M" />
         </action>
+
+        <action id="GitMachete.CheckoutNextAction"
+                class="com.virtuslab.gitmachete.frontend.actions.navigation.CheckoutNextAction">
+            <keyboard-shortcut keymap="$default" first-keystroke="alt DOWN" />
```
PawelLipski, line 217: Also add `alt RIGHT` for this action? coz we don't want to end up in a situation where alt-right does something completely unrelated to Git Machete when Git tool window is focused

src/main/resources/META-INF/plugin.xml
```diff
@@ -211,6 +211,21 @@
...
+        </action>
+
```
PawelLipski, line 219: Also one more thing: when Git tool window is opened but Git Machete tab is not, do the shortcuts take effect or not?

src/uiTest/resources/project.rhino.js
```diff
@@ -402,6 +403,18 @@ function Project(underlyingProject) {
     invokeActionAndWait('GitMachete.CheckoutSelectedAction', ACTION_PLACE_CONTEXT_MENU, { SELECTED_BRANCH_NAME: branchName });
   };
 
+  this.checkoutNextBranch = function() {
+    invokeActionAndWait('GitMachete.CheckoutNextAction', ACTION_PLACE_EMPTY,{});
```
PawelLipski, line 407: Nit:
```suggestion
    invokeActionAndWait('GitMachete.CheckoutNextAction', ACTION_PLACE_EMPTY, {});
```

src/uiTest/scala/com/virtuslab/gitmachete/uitest/UITestSuite.scala
```diff
@@ -99,6 +99,28 @@ class UITestSuite extends BaseGitRepositoryBackedIntegrationTestSuite(SETUP_WITH
     // (1 commit of `allow-ownership-link` and 3 commits of `call-ws` are all gone)
     Assert.assertEquals(13, branchAndCommitRowsCount)
 
+    project.checkoutBranch("allow-ownership-link")
```
PawelLipski, line 102: Wow there's quite a lot of checking out here in this test... actually, since UI test execution timeout is a bottleneck in CI, I'd just leave a **single invocation** to each of `checkout{Next,Prev,Parent}Branch`

src/main/resources/META-INF/plugin.xml
PawelLipski: (works on Linux)

src/main/resources/META-INF/plugin.xml
vzmerr: No they don't.

PawelLipski: Okay, gr8

PawelLipski (approved):
:benissimo:, that was a long one! approving

PawelLipski:
@vzmerr pls refrain from merging to `develop` until v3.0.3 is released (Monday)

src/main/resources/META-INF/plugin.xml
```diff
@@ -211,6 +211,22 @@
             <!-- In $default keymap `control` key (Windows/Linux) corresponds to `cmd` key on macOS -->
             <keyboard-shortcut keymap="$default" first-keystroke="control alt shift M" />
         </action>
+
+        <action id="GitMachete.CheckoutNextAction"
+                class="com.virtuslab.gitmachete.frontend.actions.navigation.CheckoutNextAction">
+            <keyboard-shortcut keymap="$default" first-keystroke="alt DOWN" />
```
mkondratek, line 217: Why do we need Alt (option on macOS)? Cannot it be just arrows? Navigation in Git Log works with arrows only.

PawelLipski: Right, good point... if it arrows truly aren't available (and/or have some unintended side effects), let's put a comment here why.

I assume the problem might lie in that Git Machete tab isn't truly _focusable_ in the same way as Log tag is... but maybe I'm wrong.

Generally, we need to avoid at all cost a scenario where a branch is checked out **inadvertently**: a user thinks that the focus is on e.g. editor and clicks an arrow, and then suddenly a different branch checks out. Note that arrows in Log tab don't cause any **side effects**, unlike with Git Machete tab.

CHANGE-NOTES.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+## v3.1.0
+- Added branch navigation allowing to checking out the next branch with `alt + down-arrow` or `alt + right-arrow`, previous branch with `alt + up-arrow`, and parent branch with `alt + left-arrow` in the machete layout.
```
PawelLipski, line 6:
```suggestion
- Added branch navigation allowing to checking out the next branch with `alt + down-arrow` or `alt + right-arrow`, previous branch with `alt + up-arrow`, and parent branch with `alt + left-arrow` when Git Machete tab is opened.
```

src/main/resources/META-INF/plugin.xml
mkondratek: okay then. so I would make arrows navigate the layout without the checkout but checkout on `Enter` key. wdyt @vzmerr , @PawelLipski ?

vzmerr: what if the user does not notice that the focus is on the machete tab, and presses enter while a branch is highlighted, assuming it would add a new line in the editor?

mkondratek: ofc there is this risk 😕 creating a warning popup before the action will make the process of checkouts less convenient (unless we provide `dont show it again` option). 

on the other hand 
1. the selection indication is rather clear and hard to miss
2. if the checkout is not possible (uncommitted changes) if will fail anyway
3. if the checkout is possible one can always checkout the previous branch back with the same ease painlessly learning on their own mistake

PawelLipski: Huh TBH this is a pretty risky idea, exactly due to

> what if the user does not notice that the focus is on the machete tab, and presses enter while a branch is highlighted, assuming it would add a new line in the editor?

My suggestion would be to first check if there is even anything like _focus on machete tab_ 🤔 I suppose that we can currently only distinguish between _machete tab shown_ and _machete tab hidden_

mkondratek: > what if the user does not notice that the focus is on the machete tab, and presses enter while a branch is highlighted, assuming it would add a new line in the editor?

I agree there is a such risk but tell the cost is minimal

PawelLipski: >  it is more clearly indicated in Git Log where additionaly you have the row selection
Awww indeed I see now 🤦🏻 

Okay... arrows let it be then! @vzmerr pls update plugin.xml, change notes and the docs... and I think we can finally merge this PR to `develop` (and thus also meld v3.0.4 into v3.1.0, i.e. the next release will be v3.1.0).

vzmerr: gee, I just remembered that assigning any pre-existing key combination to an IntelliJ plugin action, with or without focus has no effect. This means that, for example, because ctrl+c is preassigned, despite the fact that it does nothing in the machete tab, assigning it, for example, to an action for copying a branch, would not have any effect. this is the same for the arrows. Sorry for not mentioning it earlier under this thread.

PawelLipski: Pls verify experimentally that indeed pure arrows (no Alt/Option) will have no effect even when Git Machete tab is verified

vzmerr: yes, alt + arrows are indeed assigned. So I would try the pure arrows on focus and let you know.

vzmerr: I confirm that using pure arrows for navigation has **no** effect, even if machete tab has the focus.

CHANGE-NOTES.md
```diff
@@ -4,6 +4,7 @@
 
 ## v3.1.0
 - Added support for IntelliJ 2022.3.
+- Added branch navigation allowing to check out the next branch with `alt + down-arrow` or `alt + right-arrow`, previous branch with `alt + up-arrow`, and parent branch with `alt + left-arrow` when Git Machete tab is opened, and has focus.
```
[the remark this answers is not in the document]
mkondratek: no yelling Xd I am just excited b/c we are so close to merge this one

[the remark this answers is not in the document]
vzmerr: ok, so doing it right now.

CHANGE-NOTES.md
```diff
@@ -4,6 +4,7 @@
 
 ## v3.1.0
 - Added support for IntelliJ 2022.3.
+- Added branch navigation allowing to check out the next branch with `alt + down-arrow` or `alt + right-arrow`, previous branch with `alt + up-arrow`, and parent branch with `alt + left-arrow` when Git Machete tab is opened, and has focus. You would need to use the `option` key instead of `alt` in MacOS.
```
vzmerr, line 7: @mkondratek please check 👍

mkondratek, line 7:
```suggestion
- Added branch navigation allowing to check out the next branch with `alt + down-arrow` or `alt + right-arrow`, previous branch with `alt + up-arrow`, and parent branch with `alt + left-arrow` when Git Machete tab is opened, and has focus. Note that `option` key is an `alt` key counterpart in macOS.
```

mkondratek (changes requested):
almost theeereeeee

mkondratek (approved):

mkondratek (changes requested):
hold on! Xd

src/main/resources/META-INF/plugin.xml
```diff
@@ -216,6 +216,22 @@
...
+            <keyboard-shortcut keymap="$default" first-keystroke="alt RIGHT" />
```
mkondratek, line 223: @vzmerr I hope you still have patience for this one 😆 

as discussed on slack (take a look to the channel if you have not seen it yet)...

Let's be consistent with git machete CLI - if I understood correctly it means that the right arrow should go to the next sibling (not to the child)

vzmerr: @PawelLipski suggests right arrow to go to the first child

PawelLipski: +1, right arrow to first child if such child exists, otherwise do nothing

PawelLipski:
Whoops some unclear JS error in UI tests... lemme see what it could be 🤔

CHANGE-NOTES.md
```diff
@@ -4,6 +4,7 @@
 
 ## v3.1.0
 - Added support for IntelliJ 2022.3.
+- Added branch navigation allowing to check out the next branch with `alt + down-arrow` or `alt + right-arrow`, previous branch with `alt + up-arrow`, and parent branch with `alt + left-arrow` when Git Machete tab is opened, and has focus. Note that `option` key is an `alt` key counterpart in macOS.
```
PawelLipski, line 7:
```suggestion
- Added branch navigation allowing to check out:
  * previous branch with `alt + up-arrow`,
  * next branch with `alt + down-arrow`,
  * parent branch with `alt + left-arrow`,
  * first child branch `alt + right-arrow`
  when Git Machete tab is opened, and has focus. Note that `option` key is an `alt` key counterpart in macOS.
```

PawelLipski (approved):
:benissimo:

src/uiTest/resources/project.rhino.js
```diff
@@ -402,6 +403,18 @@ function Project(underlyingProject) {
...
+  this.checkoutNextBranch = function () {
+    invokeActionAndWait('GitMachete.CheckoutNextAction', ACTION_PLACE_EMPTY, {});
+  };
+
+  this.checkoutPreviousBranch = function () {
+    invokeActionAndWait('GitMachete.CheckoutPreviousAction', ACTION_PLACE_EMPTY, {});
+  };
+
+  this.checkoutParentBranch = function () {
+    invokeActionAndWait('GitMachete.ParentBranchCheckoutAction', ACTION_PLACE_EMPTY, {});
```
PawelLipski, line 415:
```suggestion
    invokeActionAndWait('GitMachete.CheckoutParentAction', ACTION_PLACE_EMPTY, {});
```

src/uiTest/resources/project.rhino.js
```diff
@@ -402,6 +403,18 @@ function Project(underlyingProject) {
     invokeActionAndWait('GitMachete.CheckoutSelectedAction', ACTION_PLACE_CONTEXT_MENU, { SELECTED_BRANCH_NAME: branchName });
   };
 
+  this.checkoutNextBranch = function () {
```
PawelLipski, line 406: Pls also use `CheckoutFirstChildAction` in one place

docs/features.md
```diff
@@ -34,6 +34,8 @@ The current branch is underlined in a branch layout.
 
 ![](checkout.gif)
 
+On the Git Machete tab, you can checkout the next branch with `alt + down-arrow` or `alt + right-arrow`, previous branch with `alt + up-arrow`, and parent branch with `alt + left-arrow` in the machete layout.
```
mkondratek, line 37: whoops - I think this is not true now

mkondratek (changes requested):

mkondratek (approved):
!!!!

Issue #1030, opened 2022-08-02
`whileDescendantOf` config for fork point override isn't just redundant, it's probably erroneous


PawelLipski:
Port the change for https://github.com/VirtusLab/git-machete/issues/611 once released

PawelLipski:
Duplicate of #1580.

PR #1168, opened 2022-09-26
Support fork points with listed and unlisted commits

### LINKED ISSUE #449: Add support for overriding fork points - listed and unlisted commits
Issue #307 covers the following cases:
- overriding to parent
- overriding to inferred

***
## The scope of this task

Enhance the `Override Fork Point` Dialog with an option (aka radioButton) containing a combobox with a list of commits.
We should allow overriding fork point to one of the branch commits (listed) and of the older ones (not listed).

For UI see: https://jetbrains.org/intellij/sdk/docs/user_interface_components/kotlin_ui_dsl.html#combo-boxes

vzmerr:
It resolves #449 

Should we add a custom renderer to render each commit with its hash and message?

### DIFF
--- a/CHANGE-NOTES.md
+++ b/CHANGE-NOTES.md
@@ -1,13 +1,12 @@
 # Changelog
 
 ## v3.2.0
+- Added new `alt + enter` intention action to create a non-existing branch if user adds it to the machete file.
+- Added option to manually pick a commit for the fork point override.
 
 ## v3.1.1
 - Removed automatic rediscovery in case of empty machete file
 - Fixed indication of a repeated entry in machete file.
-
-## v3.2.0
-- Added new `alt + enter` intention action to create a non-existing branch if user adds it to the machete file.
 
 ## v3.1.0
 - Added support for IntelliJ 2022.3.
--- a/frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseOverrideForkPointAction.java
+++ b/frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseOverrideForkPointAction.java
@@ -66,8 +66,7 @@ public abstract class BaseOverrideForkPointAction extends BaseGitMacheteReposito
     }
 
     val nonRootBranch = branch.asNonRoot();
-    val selectedCommit = new OverrideForkPointDialog(project, nonRootBranch.getParent(), nonRootBranch)
-        .showAndGetSelectedCommit();
+    val selectedCommit = new OverrideForkPointDialog(project, nonRootBranch).showAndGetSelectedCommit();
     if (selectedCommit == null) {
       log().debug(
           "Commit selected to be the new fork point is null: most likely the action has been canceled from override-fork-point dialog");
--- a/frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseOverrideForkPointAction.java
+++ b/frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseOverrideForkPointAction.java
@@ -122,7 +122,7 @@ public abstract class BaseOverrideForkPointAction extends BaseGitMacheteReposito
     }
 
     try {
-      GitConfigUtil.setValue(project, root, whileDescendantOfKey, ancestorCommit.getHash());
+      GitConfigUtil.setValue(project, root, whileDescendantOfKey, forkPoint.getHash());
     } catch (VcsException e) {
       LOG.info("Attempt to get '${whileDescendantOf}' git config value failed: " + e.getMessage());
     }
--- a/frontend/actions/src/main/kotlin/com/virtuslab/gitmachete/frontend/actions/dialogs/OverrideForkPointDialog.kt
+++ b/frontend/actions/src/main/kotlin/com/virtuslab/gitmachete/frontend/actions/dialogs/OverrideForkPointDialog.kt
@@ -18,126 +18,105 @@ import javax.swing.JList
 import javax.swing.UIManager
 
 enum class OverrideOption {
-  PARENT,
-  INFERRED
+    PARENT,
+    INFERRED
 }
 
 class OverrideForkPointDialog(
-  project: Project,
-  private val parentBranch: IManagedBranchSnapshot,
-  private val branch: INonRootManagedBranchSnapshot
+    project: Project,
+    private val parentBranch: IManagedBranchSnapshot,
+    private val branch: INonRootManagedBranchSnapshot
 ) : DialogWrapper(project, /* canBeParent */ true) {
 
-  private var myOverrideOption = OverrideOption.PARENT
+    private var myOverrideOption = OverrideOption.PARENT
 
-  private var customCommit = branch.forkPoint
-  init {
-    title =
-      getString("action.GitMachete.BaseOverrideForkPointAction.dialog.override-fork-point.title")
-    setOKButtonMnemonic('O'.code)
-    super.init()
-  }
+    private var customCommit = branch.forkPoint
 
-  fun showAndGetSelectedCommit() =
-    if (showAndGet()) {
-      when (myOverrideOption) {
-        OverrideOption.PARENT -> parentBranch.pointedCommit
-        OverrideOption.INFERRED -> branch.forkPoint
-      }
-    } else {
-      customCommit
+    init {
+        title =
+            getString("action.GitMachete.BaseOverrideForkPointAction.dialog.override-fork-point.title")
+        setOKButtonMnemonic('O'.code)
+        super.init()
     }
 
-  override fun createCenterPanel() = panel {
-    row {
-      label(
-        format(
-          getString(
-            "action.GitMachete.BaseOverrideForkPointAction.dialog.override-fork-point.label.HTML"
-          ),
-          branch.name
-        )
-      )
-    }
-    buttonsGroup {
-      row {
-        radioButton(
-          format(
-            getString(
-              "action.GitMachete.BaseOverrideForkPointAction.dialog.override-fork-point.radio-button.parent"
-            ),
-            parentBranch.name
-          ),
-          OverrideOption.PARENT
-        )
-          .comment(parentBranch.pointedCommit.shortMessage)
-      }
-
-      var thisRowComment = "cannot resolve commit message"
-
-      var radioButtonComment = "cannot resolve commit hash"
-
-      if (branch.forkPoint != null) {
-        thisRowComment = branch.forkPoint!!.shortMessage
-        radioButtonComment = branch.forkPoint!!.shortHash
-      }
-
-      row {
-        radioButton(
-          format(
-            getString(
-              "action.GitMachete.BaseOverrideForkPointAction.dialog.override-fork-point.radio-button.inferred"
-            ),
-            radioButtonComment
-          ),
-          OverrideOption.INFERRED
-        )
-          .comment(
-            thisRowComment
-          )
-      }
-    }
-      .bind(
-        MutableProperty(::myOverrideOption) { myOverrideOption = it },
-        OverrideOption::class.java
-      )
-
-    row("The fork point commit:") {
-      comboBox(
-        branch.commits.toMutableList(),
-        object : DefaultListCellRenderer() {
-          private val defaultBackground = UIManager.get("List.background") as Color
-          override fun getListCellRendererComponent(
-            list: JList<*>?,
-            value: Any,
-            index: Int,
-            isSelected: Boolean,
-            cellHasFocus: Boolean
-          ): Component {
-            super.getListCellRendererComponent(list, value, index, isSelected, cellHasFocus)
-            val commit: ICommitOfManagedBranch = value as ICommitOfManagedBranch
-
-            var prefix = ""
-
-            if (parentBranch.pointedCommit.shortHash.equals(commit.shortHash)) {
-              prefix = "parent pointed "
-            } else if (branch.forkPoint?.shortHash.equals(commit.shortHash)) {
-              prefix = "branch fork point "
+    fun showAndGetSelectedCommit() =
+        if (showAndGet()) {
+            when (myOverrideOption) {
+                OverrideOption.PARENT -> parentBranch.pointedCommit
+                OverrideOption.INFERRED -> branch.forkPoint
             }
-            text = "$prefix[${commit.shortHash}] [${commit.shortMessage}]"
-
-            if (!isSelected) {
-              setBackground(if (index % 2 == 0) background else defaultBackground)
-            }
-            return this
-          }
+        } else {
+            customCommit
         }
 
-      ).bindItem(
-        MutableProperty(::customCommit) {
-          customCommit = it as IForkPointCommitOfManagedBranch?
+    override fun createCenterPanel() = panel {
+        row {
+            label(
+                format(
+                    getString(
+                        "action.GitMachete.BaseOverrideForkPointAction.dialog.override-fork-point.label.HTML"
+                    ),
+                    branch.name
+                )
+            )
         }
-      )
+        row {
+            label(
+                format(
+                    getString(
+                        "action.GitMachete.BaseOverrideForkPointAction.dialog.override-fork-point.radio-button.parent"
+                    ),
+                    parentBranch.name
+                )
+            )
+        }
+
+        row {
+            label(
+                format(
+                    getString(
+                        "action.GitMachete.BaseOverrideForkPointAction.dialog.override-fork-point.radio-button.inferred"
+                    )
+                )
+            )
+        }
+
+        row("The fork point commit:") {
+            comboBox(
+                branch.commits.toMutableList(),
+                object : DefaultListCellRenderer() {
+                    private val defaultBackground = UIManager.get("List.background") as Color
+                    override fun getListCellRendererComponent(
+                        list: JList<*>?,
+                        value: Any,
+                        index: Int,
+                        isSelected: Boolean,
+                        cellHasFocus: Boolean
+                    ): Component {
+                        super.getListCellRendererComponent(list, value, index, isSelected, cellHasFocus)
+                        val commit: ICommitOfManagedBranch = value as ICommitOfManagedBranch
+
+                        var prefix = ""
+
+                        if (parentBranch.pointedCommit.shortHash.equals(commit.shortHash)) {
+                            prefix = "parent pointed "
+                        } else if (branch.forkPoint != null && branch.forkPoint?.shortHash.equals(commit.shortHash)) {
+                            prefix = "branch fork point "
+                        }
+                        text = "$prefix[${commit.shortHash}] [${commit.shortMessage}]"
+
+                        if (!isSelected) {
+                            setBackground(if (index % 2 == 0) background else defaultBackground)
+                        }
+                        return this
+                    }
+                }
+
+            ).bindItem(
+                MutableProperty(::customCommit) {
+                    customCommit = it as IForkPointCommitOfManagedBranch?
+                }
+            )
+        }
     }
-  }
 }
--- a/frontend/actions/src/main/kotlin/com/virtuslab/gitmachete/frontend/actions/dialogs/OverrideForkPointDialog.kt
+++ b/frontend/actions/src/main/kotlin/com/virtuslab/gitmachete/frontend/actions/dialogs/OverrideForkPointDialog.kt
@@ -111,9 +111,6 @@ class OverrideForkPointDialog(
           ),
           OverrideOption.CUSTOM
         )
-          .comment(
-            thisRowComment
-          )
 
         comboBox<ICommitOfManagedBranch?>(
           (listOf(branch.forkPoint, parentBranch.pointedCommit) + branch.commits.toMutableList()).filterNotNull(),
@@ -155,26 +152,7 @@ class OverrideForkPointDialog(
               isEnabled = list.isEnabled
 
               text = if (commit != null) {
-                var prefix =
-                  if (parentBranch.pointedCommit.shortHash.equals(commit.shortHash)) {
-                    format(
-                      getString(
-                        "action.GitMachete.BaseOverrideForkPointAction.dialog.override-fork-point.radio-button.parent"
-                      ),
-                      parentBranch.name
-                    )
-                  } else if (branch.forkPoint?.shortHash.equals(commit.shortHash)) {
-                    format(
-                      getString(
-                        "action.GitMachete.BaseOverrideForkPointAction.dialog.override-fork-point.radio-button.inferred"
-                      ),
-                      branch.name
-                    )
-                  } else {
-                    ""
-                  }
-
-                "$prefix[${commit.shortHash}] [${commit.shortMessage}]"
+                "[${commit.shortHash}] ${commit.shortMessage}"
               } else {
                 ""
               }
--- a/frontend/actions/src/main/kotlin/com/virtuslab/gitmachete/frontend/actions/dialogs/OverrideForkPointDialog.kt
+++ b/frontend/actions/src/main/kotlin/com/virtuslab/gitmachete/frontend/actions/dialogs/OverrideForkPointDialog.kt
@@ -29,8 +29,6 @@ class OverrideForkPointDialog(
 
   private val parent = branch.parent
 
-  private lateinit var rb: Cell<JBRadioButton>
-
   init {
     title =
       getString("action.GitMachete.BaseOverrideForkPointAction.dialog.override-fork-point.title")
@@ -89,7 +87,7 @@ class OverrideForkPointDialog(
       }
 
       row {
-        rb = radioButton(
+        val rb: Cell<JBRadioButton> = radioButton(
           format(
             getString(
               "action.GitMachete.BaseOverrideForkPointAction.dialog.override-fork-point.radio-button.custom"
@@ -99,8 +97,8 @@ class OverrideForkPointDialog(
           OverrideOption.CUSTOM
         )
 
-        val list = branch.commitsUntilParent.reverse() + listOf(branch.forkPoint, parent.pointedCommit)
-        val items = list.filterNotNull().distinct()
+        val list = branch.commitsUntilParent + listOf(branch.forkPoint, parent.pointedCommit)
+        val items = list.filterNotNull().distinct().reversed()
 
         comboBox(
           items,
--- a/frontend/actions/src/main/kotlin/com/virtuslab/gitmachete/frontend/actions/dialogs/OverrideForkPointDialog.kt
+++ b/frontend/actions/src/main/kotlin/com/virtuslab/gitmachete/frontend/actions/dialogs/OverrideForkPointDialog.kt
@@ -49,30 +49,10 @@ class OverrideForkPointDialog(
         )
       )
     }
-    row {
-      label(
-        format(
-          getString(
-            "action.GitMachete.BaseOverrideForkPointAction.dialog.override-fork-point.radio-button.parent"
-          ),
-          parentBranch.name
-        )
-      )
-    }
-
-    row {
-      label(
-        format(
-          getString(
-            "action.GitMachete.BaseOverrideForkPointAction.dialog.override-fork-point.radio-button.inferred"
-          )
-        )
-      )
-    }
 
     row("The fork point commit:") {
       comboBox<ICommitOfManagedBranch?>(
-        (branch.commits.toMutableList() + parentBranch.pointedCommit + branch.forkPoint).filterNotNull(),
+        (listOf(parentBranch.pointedCommit, branch.forkPoint) + branch.commits.toMutableList()).filterNotNull(),
         object : DefaultListCellRenderer() {
           private val defaultBackground = UIManager.get("List.background") as Color
           override fun getListCellRendererComponent(
@@ -113,9 +93,16 @@ class OverrideForkPointDialog(
               var prefix =
 
                 if (parentBranch.pointedCommit.shortHash.equals(commit.shortHash)) {
-                  "parent pointed "
+                  format(
+                    getString(
+                      "action.GitMachete.BaseOverrideForkPointAction.dialog.override-fork-point.radio-button.parent"
+                    ),
+                    parentBranch.name
+                  )
                 } else if (branch.forkPoint?.shortHash.equals(commit.shortHash)) {
-                  "branch fork point "
+                  getString(
+                    "action.GitMachete.BaseOverrideForkPointAction.dialog.override-fork-point.radio-button.inferred"
+                  )
                 } else {
                   ""
                 }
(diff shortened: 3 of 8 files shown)


frontend/actions/src/main/kotlin/com/virtuslab/gitmachete/frontend/actions/dialogs/OverrideForkPointDialog.kt
```diff
@@ -92,5 +95,17 @@ class OverrideForkPointDialog(
         MutableProperty(::myOverrideOption) { myOverrideOption = it },
         OverrideOption::class.java
       )
+
+    row("The fork point commit:") {
+      comboBox(
```
PawelLipski, line 100: Hmmm for the sake of consistency... would it be possible to rather have a single `radioButton` for each commit? with parent comment and inferred commit highlighted as suggested options

vzmerr: Do we still need the existing descriptions in front of the radioButtons?

PawelLipski: 🤔  could you attach the screenshot of how the entire dialog looks for the current PR state?

vzmerr: I'm trying to get it compile, first. It seems that some null value in the combobox is getting passed to the  overrided method of the custom Renderer, which leads to a null pointer exception. I am trying to debug that first.

vzmerr: currently, at runtime, some null value is getting passed from the combobox to the overrided method of the cell renderer, leading to a null pointer exception.

vzmerr: image

vzmerr: image
latest

PawelLipski: Okay... @mkondratek as you're a local UXer, could you suggest where could go from there? I've got some ideas but don't want to suggest anything yet

vzmerr: Hey @mkondratek 👋

mkondratek: whoops, Idk why but I did not receive email notifications from this one. maybe I must be added as a reviewer? dunno. feel free to ping me directly sooner if I am not responding 😆

mkondratek: Hmmm, I think that the parent branch and the inferred commit are usually the right ones and we should not hide them among the other commits. I would rather keep them.

My idea is to have 3 radio buttons:
 [ ] parent branch 
 [ ] inferred commit
 [ ] manually pointed commit <combo box with all commits>


wdyt?

PawelLipski: ☝🏻 It was indeed the original Ina's idea... although it also looked confusing to me (as inferred commit is then displayed twice: first highlighted, and then listed among other non-highlighted commits).

I'd rather see **a single list**, but:
1. scrollable
2. with some rich(er) formatting (not just combobox with plain unformatted text)
3. with parent branch and inferred commit clearly highlighted

WDYT?

vzmerr: Sometimes also the selected commit in the combobox is invalid as a fork point.

vzmerr: In the latest commit of this branch, the inferred commit and the parent one are at the top of the list of the combobox. and labeled as parent and inferred. WDYT? what can we do about the richer format or can we put them separately?

mkondratek: @PawelLipski I am not convinced to this idea. I would rather list some commits twice than abandon the clear indication of what are the most possible commits to pick

and I cannot imagine that list containing all that information within. I mean - ofc we can bold smth, add `[tag-telling-what-I-am]` etc, but I doubt it will remain looking clearly and nice

PawelLipski: > Sometimes also the selected commit in the combobox is invalid as a fork point. We should be filtering out such commits, shouldn't we?

Hmmm what specifically do you mean by _invalid as fork point_? 🤔

PawelLipski: > I am not convinced to this idea. I would rather list some commits twice than abandon the clear indication of what are the most possible commits to pick

Okay, you've got me convinced/outvoted 😅  Let's go 3 radio buttons then, and some list (not necessarily combobox, better sth that allows for rich formatting) for the _manually pointed commit_ option

vzmerr: > not necessarily combobox, better sth that allows for rich formatting

@mkondratek any recommendations?

vzmerr: @PawelLipski sorry, I wanted to reply to your comment, and by mistake I ended up editing it somehow 😬

vzmerr: > Hmmm what specifically do you mean by invalid as fork point? 🤔

I would repro the error in the plugin and let you know

mkondratek: > > not necessarily combobox, better sth that allows for rich formatting
> 
> @mkondratek any recommendations?

@vzmerr idk, while looking for components/ui ideas I often depend on what I remember from the IDE, find it out in the intellij code and apply similar things in our code. you may want to take a look into https://plugins.jetbrains.com/docs/intellij/kotlin-ui-dsl-version-2.html too. for me combobox sounds most reasonable.

iiuc @PawelLipski would be happier with a list (seems to be anti-patternish to me, but anyway). this idea basicly expands to radiobutton set within radiobutton set:

(here checkboxes but single option required ofc)
- [ ] fork point
- [ ] parent
- [ ] manually picked commit
  - [ ] abc2137 Do a lot of things
  - [ ] 2a1b3c7 Do a lot of things - 1st round of fixes
  - [ ] 020897 Do a lot of things - final commit

PawelLipski: > list (seems to be anti-patternish to me, but anyway

Could you elaborate? 🤔

mkondratek: Where have you seen a list used like that? can you bring some examples of your idea applied? from intellij? of from some website (https://www.intercity.pl/pl/ does not count 😆 )?

PawelLipski: 😅 okay maybe actually a scrollable list with radio boxes isn't the best idea, let's stick to combobox indeed.

I'd suggest another thing however (can be done within a separate issue): to display, within the Override Fork Point dialog, **the list of commits that will be considered unique to the branch, assuming the select fork point**... WDYT?

vzmerr: image
The latest look. WDYT?

PawelLipski: Looks okay to me... probably the **subscript** under _Manually picked commit_ is redundant tho

frontend/actions/src/main/kotlin/com/virtuslab/gitmachete/frontend/actions/dialogs/OverrideForkPointDialog.kt
```diff
@@ -87,6 +100,99 @@ class OverrideForkPointDialog(
...
+            private val defaultBackground = UIManager.get("List.background") as Color
+            override fun getListCellRendererComponent(
+              list: JList<*>?,
+              value: Any?,
+              index: Int,
+              isSelected: Boolean,
+              cellHasFocus: Boolean
+            ): Component {
+              val commit: ICommitOfManagedBranch? = value as ICommitOfManagedBranch?
+              componentOrientation = list!!.componentOrientation
```
PawelLipski, line 130: Hmm what's the guarantee that `list` is non-null? 🤔

frontend/actions/src/main/kotlin/com/virtuslab/gitmachete/frontend/actions/dialogs/OverrideForkPointDialog.kt
```diff
@@ -87,6 +100,99 @@ class OverrideForkPointDialog(
...
+                background = list.background
+                foreground = list.foreground
+              }
+
+              icon = null
+
+              isEnabled = list.isEnabled
+
+              text = if (commit != null) {
+                var prefix =
```
PawelLipski, line 158: Hmmm currently the combo box list looks like that:

image

for list of commits like

image

PawelLipski: I'd rather avoid including _Inferred commit_ at the beginning of the list by default (?)

Also, let's order the commits in the same order as they appear in the branch status: from oldest ("closest" to the parent branch) to the newest ("closest" to the child branch).

frontend/actions/src/main/kotlin/com/virtuslab/gitmachete/frontend/actions/dialogs/OverrideForkPointDialog.kt
```diff
@@ -87,6 +100,99 @@ class OverrideForkPointDialog(
             thisRowComment
           )
       }
+
```
PawelLipski, line 103: The more I look at that, the more I think that a list of commits **unique** to the branch (i.e. what's gonna be displayed in the status for the branch) would be useful, right above or below the radio buttons... WDYT folks?

vzmerr: @PawelLipski 👌

frontend/actions/src/main/kotlin/com/virtuslab/gitmachete/frontend/actions/dialogs/OverrideForkPointDialog.kt
vzmerr: we're the one constructing it initially with the list of commits.

frontend/actions/src/main/kotlin/com/virtuslab/gitmachete/frontend/actions/dialogs/OverrideForkPointDialog.kt
mkondratek: yup, lets keep the order

mkondratek (changes requested):
image

1. I am against including `infered commit (branch)` and `parent branch (branch)` texts - it is redundant - these info is already given by the above radio options.  

2. If we are going to keep them then a space between the text and commit hash is missing - pls add it

3. There is inaccurate comment below the new radio option - pls fix it (probably remove); example below
image

mkondratek:
In spite of the above issues, this feature looks nice 🙂 I think we could have it soon

PawelLipski:
@zmerr if you could wrap up this PR before proceeding with traverse (as this PR requires like an order of magnitude less work to complete 😅)

mkondratek:
@vzmerr I am gonna take this PR from you :)

PawelLipski:
☝🏻 pls use @zmerr, @vzmerr is no longer active

If you Ina have some unpushed work her since y'day, pls push

mkondratek (approved):
image

mkondratek:
I think it looks pretty good

image

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseOverrideForkPointAction.java
```diff
@@ -122,7 +122,7 @@ private void setOverrideForkPointConfigValues(
     }
 
     try {
-      GitConfigUtil.setValue(project, root, whileDescendantOfKey, ancestorCommit.getHash());
+      GitConfigUtil.setValue(project, root, whileDescendantOfKey, forkPoint.getHash());
```
PawelLipski, line 125: This is not strictly related to the change at hand, it's a step towards getting rid of `whileDescendantOf` completely (https://github.com/VirtusLab/git-machete-intellij-plugin/issues/1030)

frontend/actions/src/main/kotlin/com/virtuslab/gitmachete/frontend/actions/dialogs/OverrideForkPointDialog.kt
```diff
@@ -3,15 +3,24 @@ package com.virtuslab.gitmachete.frontend.actions.dialogs
 import com.intellij.openapi.project.Project
```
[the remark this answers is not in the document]
PawelLipski: Let's assume that v3.1.1 will be released separately before v3.2.0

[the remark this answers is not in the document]
PawelLipski: Plus bump the version to 3.2.0

frontend/actions/src/main/kotlin/com/virtuslab/gitmachete/frontend/actions/dialogs/OverrideForkPointDialog.kt
```diff
@@ -87,6 +100,77 @@ class OverrideForkPointDialog(
...
+            getString(
+              "action.GitMachete.BaseOverrideForkPointAction.dialog.override-fork-point.radio-button.custom"
+            ),
+            radioButtonComment
+          ),
+          OverrideOption.CUSTOM
+        )
+
+        comboBox<ICommitOfManagedBranch?>(
+          (listOf(branch.forkPoint, parentBranch.pointedCommit) + branch.commits.toMutableList()).filterNotNull(),
```
PawelLipski, line 116: Hmmm there's a problem with taking `branch.commits` as a range of commits here, as `commits` already take existing fork point override into account (if one exists 😅)

As a quick workaround, I'd:
* add `commitsUntilParent` alongside `commits` to `NonRootManagedBranchSnapshot`, computed as `gitCoreRepository.deriveCommitRange(corePointedCommit, parentCoreLocalBranch.getPointedCommit())` regardless of `syncToParentStatus`
* rename `commits` to `uniqueCommits`
* use `branch.commitsUntilParent` instead of `branch.(unique)commits`

PawelLipski: (preferably on a PR under the current one)

mkondratek: done here: https://github.com/VirtusLab/git-machete-intellij-plugin/pull/1273

frontend/actions/src/main/kotlin/com/virtuslab/gitmachete/frontend/actions/dialogs/OverrideForkPointDialog.kt
```diff
@@ -87,6 +91,42 @@ class OverrideForkPointDialog(
...
+          object : DefaultListCellRenderer() {
+            override fun getListCellRendererComponent(
+              list: JList<*>?,
+              value: Any?,
+              index: Int,
+              isSelected: Boolean,
+              cellHasFocus: Boolean
+            ): Component {
+              val commit: ICommitOfManagedBranch = value as ICommitOfManagedBranch
+              text = "<html><tt>[${commit.shortHash}]</tt> ${commit.shortMessage}</html>"
```
mkondratek, line 122: image

monospace font for commit hashes - same len of them, better look imho

CHANGE-NOTES.md
```diff
@@ -3,6 +3,9 @@
 ## Unreleased
 - Removed automatic rediscovery in case of empty machete file
```
PawelLipski, line 4: Move this to v3.1.1 below

PawelLipski: Aaand lemme open a task to get rid of this section ✂️

PawelLipski: #1274

mkondratek: ok

PawelLipski: _słuchaj... okej_

PawelLipski:
Wow _it's so f**king ~big~ wide_:

image

Use commit subject, not full message

frontend/actions/src/main/kotlin/com/virtuslab/gitmachete/frontend/actions/dialogs/OverrideForkPointDialog.kt
PawelLipski: Lemme comment here so that it's easier to track:

Wow _it's so f**king ~big~ wide_:

image

Use commit subject, not full message

PawelLipski: Although huh, it's `shortMessage`... hm WTF

PawelLipski: Awww I see, check `git show 2a09`:

image

and now check the javadoc for `org.eclipse.jgit.revwalk.RevCommit#getShortMessage`:

image

It's a general convention that the **second line** of commit message, if exists, is always empty (as a separating between subject and extended description).

However, judging by the following line in the implementation of `getShortMessage`:

```java
                if (hasLF(raw, msgB, msgE)) {
			str = StringUtils.replaceLineBreaksWithSpace(str);
		}
```

they're indeed aware that the subject (part until first double LF) itself might be multiline.

frontend/resourcebundles/src/main/resources/GitMacheteBundle.properties
```diff
@@ -64,6 +64,7 @@ action.GitMachete.BaseOverrideForkPointAction.dialog.override-fork-point.label.H
 action.GitMachete.BaseOverrideForkPointAction.dialog.override-fork-point.title=Override Fork Point
 action.GitMachete.BaseOverrideForkPointAction.dialog.override-fork-point.radio-button.parent=Parent branch ({0})
 action.GitMachete.BaseOverrideForkPointAction.dialog.override-fork-point.radio-button.inferred=Inferred commit ({0})
+action.GitMachete.BaseOverrideForkPointAction.dialog.override-fork-point.radio-button.custom=Manually picked commit:
```
PawelLipski, line 67: Also, I'd automatically switch radio to `custom` if a choice is made within the combobox... as for now, one can end up making a choice in combobox, but the radio button remains unchanged

PawelLipski: I can think of a formerly existing place in IntelliJ which such the same UX glitch... coincidentally, also related to _overriding_ things 😅 

image

Fixed in 2022.1 or 2022.2, but in the past you could select association in combobox, and nothing happened ☹️

frontend/actions/src/main/kotlin/com/virtuslab/gitmachete/frontend/actions/dialogs/OverrideForkPointDialog.kt
PawelLipski: #1277

frontend/actions/src/main/kotlin/com/virtuslab/gitmachete/frontend/actions/dialogs/OverrideForkPointDialog.kt
```diff
@@ -50,28 +59,23 @@ class OverrideForkPointDialog(
         )
       )
     }
+
```
PawelLipski, line 62: Re-posting so that it's not lost:

> I'd suggest another thing however (can be done within a separate issue): to display, within the Override Fork Point dialog, the list of commits that will be considered unique to the branch, assuming the selected fork point... WDYT?

frontend/actions/src/main/kotlin/com/virtuslab/gitmachete/frontend/actions/dialogs/OverrideForkPointDialog.kt
```diff
@@ -87,6 +91,42 @@ class OverrideForkPointDialog(
...
+            ),
+            radioButtonComment
+          ),
+          OverrideOption.CUSTOM
+        )
+
+        val mutableList = branch.commitsUntilParent.toMutableList()
+        mutableList.remove(branch.forkPoint)
+        mutableList.remove(parent.pointedCommit)
+        val items = (listOf(branch.forkPoint, parent.pointedCommit) + mutableList.reversed()).filterNotNull()
```
PawelLipski, line 109: Whoops I'd rather keep the order of commits as-is, without putting inferred fork point & parent as first items... that's very confusing (I just tried to match the order from combobox to git history and could get it ;p)

frontend/actions/src/main/kotlin/com/virtuslab/gitmachete/frontend/actions/dialogs/OverrideForkPointDialog.kt
```diff
@@ -87,6 +91,42 @@ class OverrideForkPointDialog(
...
+          items,
+          object : DefaultListCellRenderer() {
+            override fun getListCellRendererComponent(
+              list: JList<*>?,
+              value: Any?,
+              index: Int,
+              isSelected: Boolean,
+              cellHasFocus: Boolean
+            ): Component {
+              val commit: ICommitOfManagedBranch = value as ICommitOfManagedBranch
```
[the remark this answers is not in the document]
mkondratek: image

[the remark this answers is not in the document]
mkondratek: whoops xd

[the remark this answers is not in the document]
mkondratek: ok, will be fixed and look like that:

image

frontend/actions/src/main/kotlin/com/virtuslab/gitmachete/frontend/actions/dialogs/OverrideForkPointDialog.kt
mkondratek: IIUC that would be a new look - not something we want to simply add to the present dialog, right?

anyway, instead of re-posting new ideas (or more critical changes) in resolvable and collapsable comments within a PR it is better to discuss them within the related issue or to open a new one, so that it's not lost 🙂

mkondratek: https://github.com/VirtusLab/git-machete-intellij-plugin/issues/1279

frontend/resourcebundles/src/main/resources/GitMacheteBundle.properties
mkondratek: done! 

image

frontend/actions/src/main/kotlin/com/virtuslab/gitmachete/frontend/actions/dialogs/OverrideForkPointDialog.kt
mkondratek: ok so to my understanding the most reasonable order is determined by these rules:
1. unique commits until parent in the same order as in git machete tree
2. fork point 
  - if fork point **is** present among the unique commits until parent then it remains in its place
  - if fork point **is not** present among the unique commits until parent then it is added BEFORE them
3. parent - the same rules as fork point (but applied later) 

@PawelLipski, is it correct?

PawelLipski: Okay, let's go with that 👍🏻

frontend/actions/src/main/kotlin/com/virtuslab/gitmachete/frontend/actions/dialogs/OverrideForkPointDialog.kt
```diff
@@ -79,14 +81,50 @@ class OverrideForkPointDialog(
...
               "action.GitMachete.BaseOverrideForkPointAction.dialog.override-fork-point.radio-button.inferred"
             ),
-            radioButtonComment
+            branch.forkPoint?.shortHash ?: "cannot resolve commit hash"
           ),
           OverrideOption.INFERRED
         )
           .comment(
-            thisRowComment
+            escapeHtml4(branch.forkPoint?.shortMessage) ?: "cannot resolve commit message"
```
PawelLipski, line 89: For `null` input param, what does `escapeHtml4` return? also `null` (no NPE)?

mkondratek: null - I ve check that before. I should have mention it here in the comment that I am aware of what I am doing

PawelLipski: Okay if so then pls mark both param and return type of `escapeHtml4` as `@PolyNull` in stubs, see https://checkerframework.org/manual/#handling-warnings-step2

frontend/actions/src/main/kotlin/com/virtuslab/gitmachete/frontend/actions/dialogs/OverrideForkPointDialog.kt
```diff
@@ -79,13 +81,45 @@ class OverrideForkPointDialog(
...
+          format(
+            getString(
+              "action.GitMachete.BaseOverrideForkPointAction.dialog.override-fork-point.radio-button.custom"
+            ),
+            branch.forkPoint?.shortHash ?: "cannot resolve commit hash"
+          ),
+          OverrideOption.CUSTOM
+        )
+
+        val list = branch.commitsUntilParent.reverse() + listOf(branch.forkPoint, parent.pointedCommit)
```
PawelLipski, line 102: As discussed... shouldn't the parent be at top (==beginning of the list)? to match the top-to-down==oldest-to-latest ordering from status 🤔

frontend/actions/src/main/kotlin/com/virtuslab/gitmachete/frontend/actions/dialogs/OverrideForkPointDialog.kt
```diff
@@ -2,25 +2,34 @@ package com.virtuslab.gitmachete.frontend.actions.dialogs
...
-  private val parentBranch: IManagedBranchSnapshot,
   private val branch: INonRootManagedBranchSnapshot
 ) : DialogWrapper(project, /* canBeParent */ true) {
 
   private var myOverrideOption = OverrideOption.PARENT
+  private var customCommit: ICommitOfManagedBranch? = branch.forkPoint
+
+  private val parent = branch.parent
+
+  private lateinit var rb: Cell<JBRadioButton>
```
PawelLipski, line 32: Can it be just a local val? 🤔

frontend/actions/src/main/kotlin/com/virtuslab/gitmachete/frontend/actions/dialogs/OverrideForkPointDialog.kt
mkondratek: good point. check it now pls

frontend/actions/src/main/kotlin/com/virtuslab/gitmachete/frontend/actions/dialogs/OverrideForkPointDialog.kt
mkondratek: done

frontend/ui/impl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/GitMacheteStorage.java
```diff
@@ -0,0 +1,6 @@
+package com.virtuslab.gitmachete.frontend.ui.impl.table;
+
+import com.intellij.openapi.components.Storage;
+
+@Storage("git-machete.xml")
```
PawelLipski, line 5: Wow that's it? 😯

PawelLipski: Not used on this PR so far, right?

PawelLipski (approved):
🍴

PawelLipski (approved):

mkondratek:
@zmerr, thank you for starting and doing most of the work with this one!

zmerr:
> @zmerr, thank you for starting and doing most of the work with this one!

Thanks Mikolaj, but I guess you were the one who ended up doing most of the hard work 😊🎊congrats🎉

PR #1174, opened 2022-09-26
Provide more specific guidelines on change notes

mkondratek:
Pls let me know if anything more comes to your mind in this scope. 

I thought I would be good to have such guidelines and be consistent about them.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -275,6 +275,9 @@ Change notes should only be added as new bullet points listed under `Unreleased`
 Change notes should contain only the **user-facing** plugin alterations like new features, public-requested bug fixes, etc. <br/>
 In spite of the above, change notes can sometimes contain a general description of the work done in the new version
 if the alternative is leaving them empty, which should be avoided.
+The change notes should be in present tense (e.g. use `Add...` instead of `Added...`);
+they should be described in full sentences and end with a period.
+Special phrases like action name references should be highlighted with `<i>...</i>`.
 
 ### IDE supported versions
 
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -273,6 +273,9 @@ Change notes should only be added as new bullet points listed under `Unreleased`
 Change notes should contain only the **user-facing** plugin alterations like new features, public-requested bug fixes, etc. <br/>
 In spite of the above, change notes can sometimes contain a general description of the work done in the new version
 if the alternative is leaving them empty, which should be avoided.
+The change notes should be in present tense (e.g. use `Add...` instead of `Added...`);
+they should be described in full sentences and end with a period.
+Special phrases like action name references should be highlighted with `<i>...</i>`.
```
mkondratek, line 278: It could be `` `...` `` but I think `<i>` tag looks better

CONTRIBUTING.md
```diff
@@ -273,6 +273,9 @@ Change notes should only be added as new bullet points listed under `Unreleased`
 Change notes should contain only the **user-facing** plugin alterations like new features, public-requested bug fixes, etc. <br/>
 In spite of the above, change notes can sometimes contain a general description of the work done in the new version
 if the alternative is leaving them empty, which should be avoided.
+The change notes should be in present tense (e.g. use `Add...` instead of `Added...`);
```
PawelLipski, line 276: Okay so we're changing the convention from now on right?

PawelLipski: To more git-commit-like

PawelLipski (approved):

CONTRIBUTING.md
mkondratek: Looking at the former change notes I cannot say there was any convention. 
We can go for past tense - probably it is more reasonable as in contrast to commit messages, the change notes are supposed to be read by humans (users in particular, and I do not consider developers to be humans).

mkondratek: @PawelLipski wdyt?

PawelLipski: > and I do not consider developers to be humans
:uuu:

> We can go for past tense 
Yeah, let's keep as it is

mkondratek: > Yeah, let's keep as it is

this order is not clear to me

PawelLipski: Sorry ;p keep past tense

PR #1176, opened 2022-09-27
Improve osx-related parts in CONTRIBUTING.md

LukaszKontowski:
Changes make guides for macOs / osx even more obvious and complete.


LukaszKontowski:
> It might be necessary to additionally restart the terminal after running mentioned commands.

That's what I had to do on my machine to get rid of the error.

PawelLipski (approved):
<3

PR #1178, opened 2022-09-27
Use the actual path to plugin zip in CI

### LINKED ISSUE #1154: Use the actual path to plugin zip in scripts/verify-artifact-contents
As for now, `plugin_zip` is inferred as

```shell
plugin_zip=$(ls -1t build/distributions/git-machete-intellij-plugin-${plugin_version}*.zip | head -1)
```

Let's instead add a Gradle task that would print out the zip path (see how `IDEPROBE_INTELLIJ_PLUGIN_URI` is set up in UiTests.kt and use that path here, instead of relying on the latest file from build/distributions/.

LukaszKontowski:
Closes #1154


scripts/verify-artifact-contents
```diff
@@ -6,7 +6,8 @@ self_dir=$(cd "$(dirname "$0")" &>/dev/null; pwd -P)
 source "$self_dir"/utils.sh
 
 plugin_version=$(extract_version_from_current_wd)
-plugin_zip=$(ls -1t build/distributions/git-machete-intellij-plugin-${plugin_version}*.zip | head -1)
```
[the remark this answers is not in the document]
LukaszKontowski: done as suggested

scripts/verify-artifact-contents
```diff
@@ -6,7 +6,8 @@ self_dir=$(cd "$(dirname "$0")" &>/dev/null; pwd -P)
...
+source "$HOME/.plugin_uri"
+plugin_zip=$IDEPROBE_INTELLIJ_PLUGIN_URI
```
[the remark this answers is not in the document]
LukaszKontowski: Ok, to make it clear:
```
java -jar marketplace-zip-signer-cli.jar sign\
  -in "git-machete-intellij-plugin-version-SNAPSHOT+git.number.zip"\
  -out "build/distributions/signed-machete-plugin.zip"\
  -cert-file "/path/to/chain.crt"\
  -key-file "/path/to/private.pem"\
  -key-pass "$PLUGIN_SIGN_PRIVATE_KEY_PASS"
```

Do I get it right that following line:
`  -in "git-machete-intellij-plugin-version-SNAPSHOT+git.number.zip"\`
should be replaced by something like:
`  -in $FULL_PLUGIN_ZIP_PATH\`
and `FULL_PLUGIN_ZIP_PATH` should be defined before starting `java` process like this:
`FULL_PLUGIN_ZIP_PATH=$(./gradlew -q printPluginZipPath | tail -n 1)`

And all of this should be changed in this part of CONRTIBUTING.md ?

buildSrc/src/main/kotlin/com/virtuslab/gitmachete/buildsrc/PluginZipPath.kt
```diff
@@ -0,0 +1,8 @@
+package com.virtuslab.gitmachete.buildsrc
+
+import org.gradle.api.Project
+
+fun Project.printPluginZipPath() {
+  val buildPlugin = tasks.findByPath(":buildPlugin")!!
+  println(buildPlugin.outputs.files.first().path)
```
PawelLipski, line 7: I don't think this function merits a a separate file... in fact these 2 lines of body could just go directly into `doLast { ... }` block

scripts/verify-artifact-contents
```diff
@@ -6,7 +6,7 @@ self_dir=$(cd "$(dirname "$0")" &>/dev/null; pwd -P)
 source "$self_dir"/utils.sh
 
 plugin_version=$(extract_version_from_current_wd)
-plugin_zip=$(ls -1t build/distributions/git-machete-intellij-plugin-${plugin_version}*.zip | head -1)
+plugin_zip=$(./gradlew -q printPluginZipPath | tail -n 1)
```
PawelLipski, line 9: Hmm why is `tail` needed? doesn't `printPluginZipPath` print just a single line? 🤔

scripts/verify-artifact-contents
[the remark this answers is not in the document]
PawelLipski: +1, can be just `-in $(./gradlew -q printPluginZipPath)` (TBD if `tail` is needed)

scripts/verify-artifact-contents
LukaszKontowski: Well, it does print one single line by itself. But I found this https://www.linkedin.com/posts/klaus-dambeck-69214a_how-to-make-akka-serialization-bulletproof-activity-6973186005717712897-R53h?utm_source=share&utm_medium=member_desktop - which suggests that sometimes there might appear some unexpected warning from Gradle. And that would be printed before our `println`). Seems smart to me.

scripts/verify-artifact-contents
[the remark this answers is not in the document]
LukaszKontowski: Ok, so `$()` should work for this java command as well? I was not sure about this, that's why I wanted to set the variable earlier

buildSrc/src/main/kotlin/com/virtuslab/gitmachete/buildsrc/PluginZipPath.kt
LukaszKontowski: sure, done

scripts/verify-artifact-contents
PawelLipski: Hmmm this discussion seems to refer to the error that happened after an update to Gradle **4.1**... I doubt the issue is still there in Gradle 7.x that we use. I think it's safe to skip `tail`, I'd never seen a case where that would be needed

LukaszKontowski: Ok, I'll skip it then

scripts/verify-artifact-contents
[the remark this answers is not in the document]
PawelLipski: Yes, `$(...)` will capture stdout of any command, even more complex scripts can be put inside `$(...)` (although this complicates the script massively and is generally discouraged)

scripts/verify-artifact-contents
```diff
@@ -6,7 +6,7 @@ self_dir=$(cd "$(dirname "$0")" &>/dev/null; pwd -P)
 source "$self_dir"/utils.sh
 
 plugin_version=$(extract_version_from_current_wd)
```
PawelLipski, line 8: `plugin_version` isn't needed anymore

PawelLipski:
To be merged after the release of v3.0.2 (version.gradle.kts will also need to be updated)

PawelLipski (changes requested):
One more place to use the new path: `hub release create --attach` (as now the `*` glob resolved to two files, one signed, the other unsigned; this led to v3.0.2 release pointing to a weirdly named tag, instead of `v3.0.2` tag.

LukaszKontowski:
> this led to v3.0.2 release pointing to a weirdly named tag, instead of `v3.0.2` tag

🤔 Can you explain this situation clearly? I checked the v3.0.2 release and it points to a tag, that has a standard, valid name: `v3.0.2`

LukaszKontowski:
> One more place to use the new path: `hub release create --attach` (as now the `*` glob resolved to two files, one signed, the other unsigned; this led to v3.0.2 release pointing to a weirdly named tag, instead of `v3.0.2` tag.

@PawelLipski - another thing I do not understand here. How would that be possible that the `--attach` part of this command should have an impact on the tag that release will point to? I just read the manual https://hub.github.com/hub-release.1.html for the command:
`hub release create [-dpoc] [-a FILE] [-m MESSAGE|-F FILE] [-t TARGET] TAG`
and it should choose (or create if not exists) the git tag that has a name equal to the `TAG`. The `--attach` (o `-a` - the same) part of the command should not matter at all for the git tag.

LukaszKontowski:
> One more place to use the new path: `hub release create --attach`

@PawelLipski - and the last question from my side: which of the two `.zip` files do we want to attach to the releases? the "signed" one or the basic one (unsigned)? As I can see, for previous releases we attached the unsigned files. And for the 3.0.2 release - the signed .zip file has been attached. Which file should be attached for next releases?

image

PawelLipski:
> can you explain this situation clearly? I checked the v3.0.2 release and it points to a tag, that has a standard, valid name: v3.0.2

Whoops sorry, 'twas a hectic release yesterday 😅 

The [second `master` build for v3.0.2](https://app.circleci.com/pipelines/github/VirtusLab/git-machete-intellij-plugin/6725/workflows/61a7faa7-48c8-44e3-911f-7faca898a3e7/jobs/7117/) produced two files under `build/distributions/`, as can seen under its Artifacts:

image

The latter came from `buildPlugin`; the former is the signed plugin, which has only been published to artifacts since my hotfix-ish PR https://github.com/VirtusLab/git-machete-intellij-plugin/pull/1185.

I didn't, however, predict one thing: the command in `.circleci/config.yml` that opens the GitHub release is as follows:

```shell
                  hub release create \
                    --attach build/distributions/git-machete-intellij-plugin-*.zip \
                    --message "$tag"$'\n\n'"$change_notes" \
                    $tag
```

`build/distributions/git-machete-intellij-plugin-*.zip` glob expanded into **two** files, not just one; after all the expansions, the executed command looked like:

```shell
                  hub release create \
                    --attach build/distributions/git-machete-intellij-plugin-3.0.2-signed.zip build/distributions/git-machete-intellij-plugin-3.0.2.zip \
                    --message "v3.0.2<stuff...> \
                    v3.0.2
```

`build/distributions/git-machete-intellij-plugin-3.0.2-signed.zip` has been interpreted as argument to `--attach`, but `build/distributions/git-machete-intellij-plugin-3.0.2.zip` has been interpreted as **argument to `hub release create`** itself, i.e. the tag to associate with the release 😅 

Later I've removed the `build/distributions/git-machete-intellij-plugin-3.0.2.zip` tag manually:
```shell
git push origin :build/distributions/git-machete-intellij-plugin-3.0.2.zip
```

This caused the release to fall back to draft state. Then I assigned tag v3.0.2 to that release, fixing the entire mess.

PawelLipski:
> another thing I do not understand here. How would that be possible that the --attach part of this command should have an impact on the tag that release will point to?

Explained above; this was a side effect of shell globbing ☹️

PawelLipski:
> - and the last question from my side: which of the two .zip files do we want to attach to the releases? the "signed" one or the basic one (unsigned)? As I can see, for previous releases we attached the unsigned files. And for the 3.0.2 release - the signed .zip file has been attached. Which file should be attached for next releases? (the new path would point to the unsigned file)

Let's stick to the output of `buildPlugin` for all purposes (UI tests, GH release etc.) for this PR. I've opened a follow-up task to switch to the signed plugins: #1187

LukaszKontowski:
> One more place to use the new path: hub release create --attach (as now the * glob resolved to two files, one signed, the other unsigned; this led to v3.0.2 release pointing to a weirdly named tag, instead of v3.0.2 tag.

Done! :)

.circleci/config.yml
```diff
@@ -237,7 +237,7 @@ jobs:
                   tag=$corresponding_milestone_title
                   change_notes=$(./gradlew getChangelog --no-header -q)
                   hub release create \
-                    --attach build/distributions/git-machete-intellij-plugin-*.zip \
+                    --attach $(./gradlew -q printPluginZipPath) \
```
PawelLipski, line 240: To be extra safe for future & avoid shell parsing surprises:
```suggestion
                    --attach "$(./gradlew -q printPluginZipPath)" \
```

PawelLipski (approved):

PR #1179, opened 2022-09-27
gh instead of hub

### LINKED ISSUE #1145: Migrate the scripts to use `gh` instead of `hub` for consistency
Since https://github.com/VirtusLab/git-machete-intellij-plugin/pull/1124 introduced a usage of `gh` which can't be simply replaced with `hub` (only with `curl`, but that's very suboptimal). Probably the existing usages of `hub` can be quite easily migrated to `gh`.

Also, remove `hub` from Docker image used for CI.

vzmerr:
It resolves #1145

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -241,6 +241,7 @@ jobs:
                     "$(./gradlew -q printPluginZipPath)" \
                     --notes "$tag"$'\n\n'"$change_notes" \
                   
+
                   scripts/close-milestone $corresponding_milestone_title
 
   checkForUpdates:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -235,13 +235,10 @@ jobs:
                   corresponding_milestone_title=v$(extract_version_from_current_wd)
                   tag=$corresponding_milestone_title
                   change_notes=$(./gradlew getChangelog --no-header -q)
-                  
                   gh release create \
                     $tag \
-                    "$(./gradlew -q printPluginZipPath)" \
-                    --notes "$tag"$'\n\n'"$change_notes" \
-                  
-
+                    "$(./gradlew -q printSignedPluginZipPath)" \
+                    --notes "$tag"$'\n\n'"$change_notes"
                   scripts/close-milestone $corresponding_milestone_title
 
   checkForUpdates:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -47,10 +47,9 @@ commands:
             git checkout -b "$branch_name"
             git commit -a -m "<<parameters.commit_message_prefix>> $(<<parameters.commit_message_command>>)"
             git push origin "$branch_name"
-            hub pull-request \
-              --no-edit \
-              --base="$CIRCLE_BRANCH" \
-              --labels="<<parameters.labels>>"
+            gh pr create\
+              --base "$CIRCLE_BRANCH" \
+              --label "<<parameters.labels>>"
             git checkout "$CIRCLE_BRANCH"
 
 jobs:
@@ -236,10 +235,12 @@ jobs:
                   corresponding_milestone_title=v$(extract_version_from_current_wd)
                   tag=$corresponding_milestone_title
                   change_notes=$(./gradlew getChangelog --no-header -q)
-                  hub release create \
-                    --attach "$(./gradlew -q printSignedPluginZipPath)" \
-                    --message "$tag"$'\n\n'"$change_notes" \
-                    $tag
+                  
+                  gh release create \
+                    $tag \
+                    "$(./gradlew -q printPluginZipPath)" \
+                    --notes "$tag"$'\n\n'"$change_notes" \
+                  
                   scripts/close-milestone $corresponding_milestone_title
 
   checkForUpdates:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ version: 2.1
 executors:
   docker_executor:
     docker:
-      - image: 'gitmachete/intellij-plugin-ci:8.1.0'
+      - image: 'gitmachete/intellij-plugin-ci:9.0.0'
     working_directory: ~/git-machete-intellij-plugin
     resource_class: large
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -47,7 +47,7 @@ commands:
             git checkout -b "$branch_name"
             git commit -a -m "<<parameters.commit_message_prefix>> $(<<parameters.commit_message_command>>)"
             git push origin "$branch_name"
-            gh pr create\
+            gh pr create -f \
               --base "$CIRCLE_BRANCH" \
               --label "<<parameters.labels>>"
             git checkout "$CIRCLE_BRANCH"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -228,7 +228,7 @@ jobs:
                   git push origin "$tag"
             - run:
                 name: Create GitHub release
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -233,10 +232,10 @@ jobs:
                   corresponding_milestone_title=v$(extract_version_from_current_wd)
                   tag=$corresponding_milestone_title
                   change_notes=$(./gradlew getChangelog --no-header -q)
-                  hub release create \
-                    --attach build/distributions/git-machete-intellij-plugin-*.zip \
-                    --message "$tag"$'\n\n'"$change_notes" \
-                    $tag
+                  gh release create \
```
PawelLipski, line 235: Looks that as the first param, the tag is expected, only then the files:

```
USAGE
  gh release create [<tag>] [<files>...]
```

.circleci/config.yml
```diff
@@ -233,10 +232,10 @@ jobs:
...
+                    build/distributions/git-machete-intellij-plugin-*.zip \
+                    --notes "$tag"$'\n\n'"$change_notes" \
+                    --notes-start-tag $tag
```
PawelLipski, line 238: What's the effect of using both `--notes` and `--notes-start-tag`? Looks that the latter makes sense only in conjunction with `--generate-notes`, to generate the notes out of commit messages between the `start-tag` and the `<tag>` passed as param to `gh release create`. Pls check on a separate repo

.circleci/config.yml
```diff
@@ -47,10 +47,9 @@ commands:
             git checkout -b "$branch_name"
```
PawelLipski, line 47: Also `scripts/enforce-change-notes-updated-on-pr-to-master` + Dockerfile to be patched

PawelLipski:
Lemme move that one to v3.0.3...

scripts/enforce-change-notes-updated-on-pr-to-master
```diff
@@ -10,8 +10,8 @@ change_notes_path=CHANGE-NOTES.md
 if [[ ${CI_PULL_REQUEST-} ]]; then
   # In case of a PR build, CI_PULL_REQUEST should be a link of the form https://github.com/VirtusLab/git-machete-intellij-plugin/pull/123
   pr_num=${CI_PULL_REQUEST##*/}
-  # Implicit usage of ${GITHUB_TOKEN} by hub
-  base_branch=$(hub pr show --format=%B "$pr_num")
+  # Implicit usage of ${GITHUB_TOKEN} by gh
+  base_branch=$(gh pr view -w "$pr_num") #the gh command does not have an equivalent option for --format
```
PawelLipski, line 14: `--json` and/or `--jq`

docker/Dockerfile
```diff
@@ -7,7 +7,7 @@ SHELL ["/bin/bash", "-euxo", "pipefail", "-c"]
 
 RUN \
   apt-get update; \
-  apt-get install --no-install-recommends -y git hub openssh-client
```
PawelLipski, line 10: Pls re-generate the image (CONTRIBUTING.md FTW) + bump the version in `docker_executor` in CI config

PawelLipski: This is technically a breaking change, so bump MAJOR part of the version, semver FTW

PawelLipski: > Pls re-generate the image (CONTRIBUTING.md FTW) + bump the version in `docker_executor` in CI config

Aww I forgot you did that already IIRC ;D

vzmerr: > This is technically a breaking change, so bump MAJOR part of the version, semver FTW

`4.0.0`? I thought it was `3.0.3`

PawelLipski: No no, I mean the Docker image version (https://hub.docker.com/r/gitmachete/intellij-plugin-ci/tags)

PawelLipski: So 9.0.0 now

vzmerr: `9.0.0` because we are removing support for `hub` right?

scripts/enforce-change-notes-updated-on-pr-to-master
vzmerr: but we don't need the `-w`  tag because we just want to extract the base branch name. we don't intend to open the PR in the browser right?

PawelLipski: Whoops yes, `-w` is definitely not expected in CI

PawelLipski: I mean, although Selenium tests in certain apps do run a browser in Docker/CI ;D but that's not our case

scripts/enforce-change-notes-updated-on-pr-to-master
```diff
@@ -10,8 +10,8 @@ change_notes_path=CHANGE-NOTES.md
 if [[ ${CI_PULL_REQUEST-} ]]; then
   # In case of a PR build, CI_PULL_REQUEST should be a link of the form https://github.com/VirtusLab/git-machete-intellij-plugin/pull/123
   pr_num=${CI_PULL_REQUEST##*/}
-  # Implicit usage of ${GITHUB_TOKEN} by hub
-  base_branch=$(hub pr show --format=%B "$pr_num")
+  # Implicit usage of ${GITHUB_TOKEN} by gh
+  base_branch=$(gh pr view "$pr_num" --json "baseRefName") #here, the goal is not to open in a browser
```
PawelLipski, line 14: Comment is rather redundant ;)

vzmerr, line 14:
```suggestion
  base_branch=$(gh pr view "$pr_num" --json "baseRefName")
```

.circleci/config.yml
vzmerr: it seems that `--notes-start-tag` is unnecessary. It does not seem to make a difference. What matters is putting the "tag" itself immediately after `create`

vzmerr: I tried it here https://github.com/vzmerr/gh-test/releases

.circleci/config.yml
```diff
@@ -47,10 +47,9 @@ commands:
...
             git commit -a -m "<<parameters.commit_message_prefix>> $(<<parameters.commit_message_command>>)"
             git push origin "$branch_name"
-            hub pull-request \
-              --no-edit \
-              --base="$CIRCLE_BRANCH" \
-              --labels="<<parameters.labels>>"
+            gh pr create\
+              --base "$CIRCLE_BRANCH" \
+              --label "<<parameters.labels>>"
```
vzmerr, line 52: In here, we should pass the title, body, and type of the PR (submit, submit as draft. ...) otherwise, it would interactively ask in the cli. what are good values for these params?

.circleci/config.yml
```diff
@@ -47,10 +47,9 @@ commands:
             git checkout -b "$branch_name"
             git commit -a -m "<<parameters.commit_message_prefix>> $(<<parameters.commit_message_command>>)"
             git push origin "$branch_name"
-            hub pull-request \
-              --no-edit \
-              --base="$CIRCLE_BRANCH" \
-              --labels="<<parameters.labels>>"
+            gh pr create\
```
vzmerr, line 50:
```suggestion
            gh pr create -f \
```

.circleci/config.yml
PawelLipski: (I suppose this is no longer a concern)

PawelLipski (approved):
All traces of `hub` removed, gr8 🎉

PR #1182, opened 2022-09-28
Fix missing open git machete tab action in Git menu (closes #1181)

### LINKED ISSUE #1181: Open Git Machete Tab action is missing in Git menu
ok, I see know. `GitRepositoryActions` used to be the action group available from menu `Git`. 
**Now, OpenMacheteTabAction is NOT available in that menu.**
git4idea must have change it in the meantime. We even was expecting that but forgot it; we have the comment 
```
```

I ve tested it, and the following:
```
 <add-to-group group-id="VcsGroups" anchor="last"/>
```
works fine now - as expected.

I am opening a separate issue for this one as the lack of our action in Git menu is a bug.

_Originally posted by @mkondratek in https://github.com/VirtusLab/git-machete-intellij-plugin/pull/1167#discussion_r982286348_

mkondratek:


src/main/resources/META-INF/plugin.xml
```diff
@@ -206,9 +206,7 @@
         <action id="GitMachete.OpenMacheteTabAction"
                 class="com.virtuslab.gitmachete.frontend.actions.vcsmenu.OpenMacheteTabAction"
                 icon="com.virtuslab.gitmachete.frontend.icons.MacheteIcons.MACHETE_FILE">
-            <!-- In future IntelliJ releases this group-id will probably be `Git.MainMenu`
-                 We can consider `VcsGroups` that seems to be more stable and place this action in VCS menu -->
-            <add-to-group group-id="GitRepositoryActions" anchor="last"/>
+            <add-to-group group-id="VcsGroups" anchor="last"/>
```
PawelLipski, line 209: Pls add a comment about where this `<add-group>` puts our action... I assume it's _Git menu_ as in, here:
image
?

mkondratek: image

PawelLipski (approved):

PR #1184, opened 2022-09-28
get next and previous branch in the BranchLayout

vzmerr:


branchLayout/api/src/main/java/com/virtuslab/branchlayout/api/BranchLayout.java
```diff
@@ -29,6 +30,28 @@ private static List<IBranchLayoutEntry> collectEntriesRecursively(IBranchLayoutE
...
+    }
+    return null;
+  }
+
+  @Override
+  public @Nullable IBranchLayoutEntry findPreviousEntry(String branchName) {
+    val entriesOrderedList = rootEntries.flatMap(BranchLayout::collectEntriesRecursively).map(IBranchLayoutEntry::getName);
+    val currentIndex = entriesOrderedList.indexOf(branchName);
+    val previousIndex = currentIndex - 1;
+    if (previousIndex > -1 && previousIndex < /* added for preventing a type error */entriesOrderedList.length()) {
```
PawelLipski, line 49: Maybe add `@LTLengthOf("entriesOrderedList")` to `previousIndex`, as in the above method?

vzmerr: I tried that, it is not allowed syntactically, for some strange reason.

vzmerr: sorry, the checker we have on the server, says that `currentIndex - 1` cannot be cast to that. Locally it compiles.

branchLayout/impl/src/test/java/com/virtuslab/branchlayout/unit/BranchLayoutTestSuite.java
```diff
@@ -49,6 +49,45 @@ public void withBranchSlideOut_givenNonRootExistingBranch_slidesOut() {
...
+
+    val entry1 = new BranchLayoutEntry(parentName1, /* customAnnotation */ null, childBranches1);
+    val rootEntry = new BranchLayoutEntry(rootName, /* customAnnotation */ null, List.of(entry0, entry1));
+    val branchLayout = new BranchLayout(List.of(rootEntry));
+
+    assertEquals(branchLayout.findNextEntry(childName1).getName(), parentName1);
+    assertEquals(branchLayout.findPreviousEntry(childName1).getName(), childName0);
+    assertEquals(branchLayout.findNextEntry(parentName0).getName(), childName0);
+    assertEquals(branchLayout.findPreviousEntry(childName0).getName(), parentName0);
+    assertEquals(branchLayout.findPreviousEntry(parentName1).getName(), childName1);
```
PawelLipski, line 88: I'd also add cases where the result of `findPreviousEntry` and `findNextEntry` are null.

branchLayout/impl/src/test/java/com/virtuslab/branchlayout/unit/BranchLayoutTestSuite.java
```diff
@@ -49,6 +49,45 @@ public void withBranchSlideOut_givenNonRootExistingBranch_slidesOut() {
...
+  }
+
```
PawelLipski, line 90: Also, a branch layout of size 1 (just a single root branch) would be useful, as well as a branch layout that's consisted only of roots (no child branches).

branchLayout/api/src/main/java/com/virtuslab/branchlayout/api/BranchLayout.java
PawelLipski: Note that locally `IndexChecker` isn't executed by default... does the code compile locally even with `-PrunAllCheckers`?

branchLayout/impl/src/test/java/com/virtuslab/branchlayout/unit/BranchLayoutTestSuite.java
```diff
@@ -49,6 +50,78 @@ public void withBranchSlideOut_givenNonRootExistingBranch_slidesOut() {
     assertEquals(children.get(1).getName(), childName1);
   }
 
+  @Test
+  public void shouldBeAbleToFindNextAndPreviousBranches() {
```
mkondratek, line 54: unit tests. awesome 👍

branchLayout/api/src/main/java/com/virtuslab/branchlayout/api/BranchLayout.java
```diff
@@ -29,6 +30,28 @@ private static List<IBranchLayoutEntry> collectEntriesRecursively(IBranchLayoutE
...
+  @Override
+  public @Nullable IBranchLayoutEntry findPreviousEntry(String branchName) {
+    val entriesOrderedList = rootEntries.flatMap(BranchLayout::collectEntriesRecursively).map(IBranchLayoutEntry::getName);
+    val currentIndex = entriesOrderedList.indexOf(branchName);
+    if (currentIndex > 0 && currentIndex < /* added for preventing a type error */entriesOrderedList.length() - 1) {
+      @LTLengthOf("entriesOrderedList") int previousIndex = currentIndex - 1;
+      return findEntryByName(entriesOrderedList.get(previousIndex));
+    }
+    return null;
+  }
```
mkondratek, line 53:
```suggestion
  @Override
  public @Nullable IBranchLayoutEntry findNextEntry(String branchName) {
    return findEntry(branchName, -1, entriesOrderedList -> entriesOrderedList.length() - 1, /* shift */ 1);
  }

  @Override
  public @Nullable IBranchLayoutEntry findPreviousEntry(String branchName) {
    return findEntry(branchName, 0, List::length, /* shift */ -1);
  }

  private @Nullable IBranchLayoutEntry findEntry(String branchName, int lowerBound, Function1<List<String>, Integer> upperBound, int shift) {
    val entriesOrderedList = rootEntries.flatMap(BranchLayout::collectEntriesRecursively).map(IBranchLayoutEntry::getName);
    val currentIndex = entriesOrderedList.indexOf(branchName);
    if (currentIndex > lowerBound && currentIndex < upperBound.apply(entriesOrderedList)) {
      @LTLengthOf("entriesOrderedList") int index = currentIndex + shift;
      return findEntryByName(entriesOrderedList.get(index));
    }
    return null;
  }
```

now, probably I am overthinking 😆 but it's a few lines shorter!

branchLayout/api/src/main/java/com/virtuslab/branchlayout/api/BranchLayout.java
```diff
@@ -29,6 +30,28 @@ private static List<IBranchLayoutEntry> collectEntriesRecursively(IBranchLayoutE
...
+      return findEntryByName(entriesOrderedList.get(nextIndex));
+    }
+    return null;
+  }
+
+  @Override
+  public @Nullable IBranchLayoutEntry findPreviousEntry(String branchName) {
+    val entriesOrderedList = rootEntries.flatMap(BranchLayout::collectEntriesRecursively).map(IBranchLayoutEntry::getName);
+    val currentIndex = entriesOrderedList.indexOf(branchName);
+    if (currentIndex > 0 && currentIndex < /* added for preventing a type error */entriesOrderedList.length() - 1) {
```
mkondratek, line 48: hmmm 🤔 what do you mean by `/* added for preventing a type error */` (there is missing space after the comment)? 

why the upper bound here is decreased by 1? It should be just `currentIndex < entriesOrderedList.length()`, am I right?

mkondratek (changes requested):

branchLayout/api/src/main/java/com/virtuslab/branchlayout/api/BranchLayout.java
vzmerr: it is for making our checker understand that the value passed to `.get()` is within the allowed bounds :(

branchLayout/api/src/main/java/com/virtuslab/branchlayout/api/BranchLayout.java
vzmerr: hmmm, that's a nice refactoring indeed 👍 I was just wondering about the use cases where we need to jump more than one branch up or down🤔 any ideas?

branchLayout/impl/src/test/java/com/virtuslab/branchlayout/unit/BranchLayoutTestSuite.java
```diff
@@ -49,6 +50,78 @@ public void withBranchSlideOut_givenNonRootExistingBranch_slidesOut() {
...
+
+    val rootEntry = new BranchLayoutEntry(rootName, /* customAnnotation */ null, List.empty());
+    val branchLayout = new BranchLayout(List.of(rootEntry));
+    //then
+    assertNull(branchLayout.findPreviousEntry(rootName));
+    assertNull(branchLayout.findNextEntry(rootName));
+  }
+
+  @Test
+  public void givenOnlyRootsEntryLayout_nextAndPreviousShouldBeNull() {
```
PawelLipski, line 107: Test name not really adequate (as for `rootName1` next and previous aren't null) ☹️

branchLayout/api/src/main/java/com/virtuslab/branchlayout/api/BranchLayout.java
mkondratek: ok, but now for branch layout

```
master
develop
```

```
findPreviousBranch("develop") == null
```

branchLayout/impl/src/test/java/com/virtuslab/branchlayout/unit/BranchLayoutTestSuite.java
```diff
@@ -49,6 +50,78 @@ public void withBranchSlideOut_givenNonRootExistingBranch_slidesOut() {
...
+    val rootEntry = new BranchLayoutEntry(rootName, /* customAnnotation */ null, List.empty());
+    val rootEntry1 = new BranchLayoutEntry(rootName1, /* customAnnotation */ null, List.empty());
+    val rootEntry2 = new BranchLayoutEntry(rootName2, /* customAnnotation */ null, List.empty());
+    val branchLayout = new BranchLayout(List.of(rootEntry, rootEntry1, rootEntry2));
+
+    //then
+    assertNull(branchLayout.findPreviousEntry(rootName));
+    assertNull(branchLayout.findNextEntry(rootName2));
+    assertEquals(branchLayout.findNextEntry(rootName1).getName(), rootName2);
+    assertEquals(branchLayout.findPreviousEntry(rootName1).getName(), rootName);
```
vzmerr, line 122: @mkondratek this is the test

branchLayout/api/src/main/java/com/virtuslab/branchlayout/api/BranchLayout.java
PawelLipski: Hmmm okay let's analyze that case... won't `currentIndex` for `develop` be `1`? with `entriesOrderedList.length()` being 2, the `if`'s condition isn't true 🤔 so I'd indeed see `null` being returned here... something's fishy here 🐟

branchLayout/impl/src/test/java/com/virtuslab/branchlayout/unit/BranchLayoutTestSuite.java
PawelLipski: That's a slightly different case though... what does `branchLayout.findPreviousEntry(rootName2)` return?

PawelLipski: Similarly, I'd also add a check for `branchLayout.findNextEntry(rootName)`

branchLayout/api/src/main/java/com/virtuslab/branchlayout/api/BranchLayout.java
vzmerr, line 48:
```suggestion
    if (currentIndex > 0 && currentIndex < /* added for preventing a type error */entriesOrderedList.length() + 1) {
```

vzmerr: because we want `currentIndex - 1 < length`

branchLayout/impl/src/test/java/com/virtuslab/branchlayout/unit/BranchLayoutTestSuite.java
mkondratek: hmmm 🤔 that's not exactly the case I am worried about.

Try this one:
```
    assertEquals(branchLayout.findPreviousEntry(rootName2).getName(), rootName1);
```
I think it should pass but it fails

branchLayout/api/src/main/java/com/virtuslab/branchlayout/api/BranchLayout.java
PawelLipski: I.e. `currentIndex <= entriesOrderedList.length()`? 🤔 is it even possible to get an index like that from `indexOf`?

branchLayout/impl/src/test/java/com/virtuslab/branchlayout/unit/BranchLayoutTestSuite.java
mkondratek: but without `- 1` it works correctly

branchLayout/api/src/main/java/com/virtuslab/branchlayout/api/BranchLayout.java
vzmerr: Locally, the build was successful with `-PrunAllCheckers`

branchLayout/impl/src/test/java/com/virtuslab/branchlayout/unit/BranchLayoutTestSuite.java
vzmerr: Sorry, that was a mistake in the array bound which is now corrected . Now I also added the test case and everything seems to work.

PawelLipski: Good spot @mkondratek

PawelLipski (approved):

branchLayout/api/src/main/java/com/virtuslab/branchlayout/api/BranchLayout.java
vzmerr: >  is it even possible to get an index equal to entriesOrderedList.length() from indexOf
oh, our checker cannot deduct that. But I removed the `-1` which was causing the null return.

mkondratek (approved):
lgtm now. great work 🎉

PR #1189, opened 2022-09-29
use signed plugin zip as an attachment to release

### LINKED ISSUE #1187: Use the output of `signPlugin` in the contexts where the output of `buildPlugin` is used
Basically, a follow-up to https://github.com/VirtusLab/git-machete-intellij-plugin/pull/1178/.
I'd also pass the signed zip to UI tests (instead of the unsigned one).

There's one catch: the name of the output zip from `signPlugin` seems to differ depending on whether the zip file exists.
If a zip file already exists, a new file with `-signed` suffix is created; if not, then the signed zip is saved under the base name.
To be researched if there are any traps here that could affect any of the contexts. Opening an issue against gradle-intellij-plugin might be needed.

LukaszKontowski:
Closes #1187

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -237,7 +237,7 @@ jobs:
                   tag=$corresponding_milestone_title
                   change_notes=$(./gradlew getChangelog --no-header -q)
                   hub release create \
-                    --attach "$(./gradlew printSignedPluginZipPath)" \
+                    --attach "$(./gradlew -q printSignedPluginZipPath)" \
                     --message "$tag"$'\n\n'"$change_notes" \
                     $tag
                   scripts/close-milestone $corresponding_milestone_title
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -137,11 +137,6 @@ jobs:
       - run:
           name: Sign plugin artifact
           command: ./gradlew signPlugin
-      - run:
-          name: test_sign_zip_path
-          command: |
-            test_var=$(./gradlew printSignedPluginZipPath)
-            echo "TESTING: $test_var"
       - store_artifacts:
           path: build/distributions/
           destination: .
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -236,9 +236,8 @@ jobs:
                   corresponding_milestone_title=v$(extract_version_from_current_wd)
                   tag=$corresponding_milestone_title
                   change_notes=$(./gradlew getChangelog --no-header -q)
-                  signed_plugin_zip_path=$(./gradlew -q printPluginZipPath | sed 's/.zip/-signed.zip/')
                   hub release create \
-                    --attach "$signed_plugin_zip_path" \
+                    --attach "$(./gradlew printSignedPluginZipPath)" \
                     --message "$tag"$'\n\n'"$change_notes" \
                     $tag
                   scripts/close-milestone $corresponding_milestone_title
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -236,8 +236,9 @@ jobs:
                   corresponding_milestone_title=v$(extract_version_from_current_wd)
                   tag=$corresponding_milestone_title
                   change_notes=$(./gradlew getChangelog --no-header -q)
+                  signed_plugin_zip_path=$(./gradlew -q printPluginZipPath | sed 's/.zip/-signed.zip/')
```
PawelLipski, line 239: This name format is brittle, might change with subsequent release.

I'd rather add a separate `printSignedPluginZipPath`.

Also, how about
> There's one catch: the name of the output zip from signPlugin seems to differ depending on whether the zip file exists.
If a zip file already exists, a new file with -signed suffix is created; if not, then the signed zip is saved under the base name.
To be researched if there are any traps here that could affect any of the contexts.

❓

LukaszKontowski: Both unsigned and signed plugin zips should be there already, as we have the `./gradlew publishPlugin` invoked earlier. At least this is what we have in CONTRIBUTING.md so it should be true:
> Then run `./gradlew publishPlugin` to produce the unsigned and signed `.zip` files in the `build/distributions/` directory.

So, this problem should not be the case here.

image

LukaszKontowski:
> This name format is brittle, might change with subsequent release.
>
> I'd rather add a separate printSignedPluginZipPath task.

Sure, I will do that

build.gradle.kts
```diff
@@ -57,6 +57,13 @@ tasks.register("printPluginZipPath") {
     println(buildPlugin.outputs.files.first().path)
   }
 }
+tasks.register("printSignedPluginZipPath") {
+  doLast {
+    val signPlugin = tasks.findByPath(":signPlugin")!!
+    val signedPluginZipPath = signPlugin.outputs.files.first { file -> file.name.contains("signed") }.path
```
PawelLipski, line 63: Huh there are multiple files in the outputs? 🤔 what is the other file(s) that is/are filtered out?

LukaszKontowski: I'm not sure, what exactly are the outputs of this `signPlugin` task (it's description was not that clear to me). Let me test that with `first()` without this filtering.

LukaszKontowski: Ok, looks like we can go without filtering: 
```
Task :printSignedPluginZipPath
/root/git-machete-intellij-plugin/build/distributions/git-machete-intellij-plugin-3.0.3-6-SNAPSHOT+git.b63155a-signed.zip
```

PawelLipski: You can also use the very awesome `./gradlew :<task> :taskTree`; with some options it displays the inputs and/or outputs

.circleci/config.yml
```diff
@@ -237,7 +237,7 @@ jobs:
                   tag=$corresponding_milestone_title
                   change_notes=$(./gradlew getChangelog --no-header -q)
                   hub release create \
-                    --attach "$(./gradlew -q printPluginZipPath)" \
+                    --attach "$(./gradlew printSignedPluginZipPath)" \
```
PawelLipski, line 240:
```suggestion
                    --attach "$(./gradlew -q printSignedPluginZipPath)" \
```
To make sure no spam slips through

PawelLipski (approved):

PR #1191, opened 2022-09-29
Enable ctrl click as right click for mac os

### LINKED ISSUE #1190: Mac OS - actions on branch not displayed after "right click" on touchbar
Actions on branch are not displayed after "right click" on touchbar (Ctrl + click). See how it should work: https://github.com/VirtusLab/git-machete-intellij-plugin#scenario-1-review
After `Ctrl + click` on the branch name - nothing is displayed. 

There is **no** similar problem when using a proper mouse on the Mac laptop (where you have a real right-click). This should work the same way as with a right-click using a mouse.

Ctrl + click works fine in other places in IntelliJ IDEA

Appears for:
MacBook Apple M1 Pro 16-inch 2021

LukaszKontowski:
Closes #1190

Tested locally with `./gradlew runIde` - worked well

### DIFF
--- a/frontend/ui/impl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/EnhancedGraphTable.java
+++ b/frontend/ui/impl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/EnhancedGraphTable.java
@@ -420,7 +420,7 @@ public final class EnhancedGraphTable extends BaseEnhancedGraphTable
     @UIEffect
     private void performActionAfterChecks(MouseEvent e, Point point) {
       ActionManager actionManager = ActionManager.getInstance();
-      if (SwingUtilities.isRightMouseButton(e) || isMacCtrlClick(e)) {
+      if (SwingUtilities.isRightMouseButton(e) || isCtrlClick(e)) {
         ActionGroup contextMenuActionGroup = (ActionGroup) actionManager.getAction(ActionGroupIds.ACTION_GROUP_CONTEXT_MENU);
         val actionPopupMenu = actionManager.createActionPopupMenu(ACTION_PLACE_CONTEXT_MENU, contextMenuActionGroup);
         JPopupMenu popupMenu = actionPopupMenu.getComponent();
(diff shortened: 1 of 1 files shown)


frontend/ui/impl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/EnhancedGraphTable.java
```diff
@@ -445,6 +445,12 @@ private void performActionAfterChecks(MouseEvent e, Point point) {
         actionManager.getAction(ACTION_CHECK_OUT).actionPerformed(actionEvent);
       }
     }
+
+    @UIEffect
+    private boolean isMacCtrlClick(MouseEvent e) {
+      return System.getProperty("os.name").contains("Mac OS X") &&
```
PawelLipski, line 451: Note that https://mkyong.com/java/how-to-detect-os-in-java-systemgetpropertyosname/ gives an even broader condition. In cases like that it's better to be more **lax** than strict... dunno how the format might change b/w different JVMs/OS versions 🤔

PawelLipski: In fact... is the check for OS X even needed in the first place? I think we can also show the menu if someone left-clicks with Ctrl down on other systems (which is extremely unlikely to happen by accident anyway) 😅

mkondratek: hmmm i do not think jetbrains are doing such an "if-ology" about right click - at least not at that level. see

```
    PopupHandler.installPopupMenu(myGraphTable, VcsLogActionIds.POPUP_ACTION_GROUP, ActionPlaces.VCS_LOG_TABLE_PLACE);
```

from `com/intellij/vcs/log/ui/frame/MainFrame.java`

PawelLipski: Hmmm how is that piece related to right click or Mac tho? 🤔 

image

mkondratek: `installPopupMenu` -> idk how it works exactly but it works for git log. it makes context menu for git log available (with both "right clicks")

LukaszKontowski: > In fact... is the check for OS X even needed in the first place? I think we can also show the menu if someone left-clicks with Ctrl down on other systems (which is extremely unlikely to happen by accident anyway)

Sounds like a good idea to me :)

LukaszKontowski:
> I think we can also show the menu if someone left-clicks with Ctrl down on other systems

Done :)

PR #1208, opened 2022-10-03
Provide `intellijVersions` by `rootProject.extra` instead of an extension

PawelLipski:
image

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -13,11 +13,6 @@ plugins {
   scala
 }
 
-project.extensions.add(
-  "intellijVersions",
-  IntellijVersions(IntellijVersionHelper.getProperties(), project.properties["overrideBuildTarget"] as String?)
-)
-
 apply<GradleVersionsFilterPlugin>()
 apply<VersionCatalogUpdatePlugin>()
 apply<TaskTreePlugin>()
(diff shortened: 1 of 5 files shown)


build.gradle.kts
```diff
@@ -13,11 +13,6 @@ plugins {
   scala
 }
 
-project.extensions.add(
```
PawelLipski, line 16: WDYT @MaciejG604? I don't think using an extension is justified/necessary here, `intellijVersions` don't provide any methods, just read-only properties

buildSrc/src/main/kotlin/com/virtuslab/gitmachete/buildsrc/UpdateIntellijVersions.kt
```diff
@@ -6,8 +6,8 @@ import com.virtuslab.gitmachete.buildsrc.IntellijVersionHelper.toBuildNumber
 import com.virtuslab.gitmachete.buildsrc.IntellijVersionHelper.versionIsNewerThan
 import org.gradle.api.DefaultTask
 import org.gradle.api.tasks.TaskAction
-import org.gradle.kotlin.dsl.get
-import org.gradle.kotlin.dsl.getByType
+import org.gradle.kotlin.dsl.extra
+import org.gradle.kotlin.dsl.provideDelegate
```
PawelLipski, line 10: `extra` don't provide `getValue`/`setValue` methods [required for a delegate](https://kotlinlang.org/docs/delegated-properties.html#property-delegate-requirements) out of themselves, hence this extra magic is needed

PawelLipski:
Spurious UI test failure, yet unknown one:

image

Opened #1210 for that...

build.gradle.kts
MaciejG604: Yeah, that is true.

MaciejG604 (approved):

PR #1216, opened 2022-10-04
Machete file/generate error for repeated branch

### LINKED ISSUE #1195: Repeated branch isn't underlined with red squiggles in the .git/machete editor
Repeated branch **should be** underlined with red squiggles - like an error - in the .git/machete editor. Currently we do not have it - as on the example below:
image

LukaszKontowski:
Closes #1195


frontend/resourcebundles/src/main/resources/GitMacheteBundle.properties
```diff
@@ -314,6 +314,7 @@ string.GitMachete.RediscoverSuggester.backgroundable-check-task.title=Checking f
 # Machete file editor
 
 string.GitMachete.MacheteAnnotator.cannot-find-local-branch-in-repo=Can''t find local branch ''{0}'' in git repository
+string.GitMachete.MacheteAnnotator.branch-name-entry-is-duplicate=Duplicate entry for branch ''{0}''
```
mkondratek, line 317:
```suggestion
string.GitMachete.MacheteAnnotator.branch-entry-already-defined=Entry for branch ''{0}'' is  already defined
```

mkondratek (changes requested):
note that we have this notification too:

image


I ve also tested it against non existing branch (no local branch - it works) however I managed to add the second entry of develop without the red underline 

image

LukaszKontowski:
> I ve also tested it against non existing branch (no local branch - it works) however I managed to add the second entry of develop without the red underline

@mkondratek 
This should be fixed now

mkondratek:
@LukaszKontowski, Ive tested it locally and for me it looks like **the law of conservation of bugs** is maintained, great work! 🤣 

image

so now only the last occurrence is highlighted. the question is if it should be 
- all 
- second and later 

i have no strong feeling about any of these so use your own judgment

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/highlighting/MacheteAnnotator.java
```diff
@@ -80,6 +86,28 @@ private void processMacheteGeneratedEntry(MacheteGeneratedEntry macheteEntry, An
                   .format(processedBranchName))
           .range(branch).create();
     }
+
+    // saveDocumentBeforeCheck needed in order to update the state of the .git/machete VirtualFile before actual check
```
mkondratek, line 90: nit: I do not like method/class references from the usual comments. it is another thing to remember while refactoring. I would rather see this comment as javadoc in the place where the method is declared

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/highlighting/MacheteAnnotator.java
```diff
@@ -4,16 +4,22 @@
 import static com.virtuslab.gitmachete.frontend.resourcebundles.GitMacheteBundle.getNonHtmlString;
```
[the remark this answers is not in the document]
LukaszKontowski: Ok, I don't understand. Is there something wrong with this alert? Is it related to this PR?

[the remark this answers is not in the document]
PawelLipski: Yeah, the text is cropped... but that's absolutely outside the scope of this PR 😅

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/highlighting/MacheteAnnotator.java
```diff
@@ -80,6 +86,28 @@ private void processMacheteGeneratedEntry(MacheteGeneratedEntry macheteEntry, An
...
+    try {
+      val branchNamesFromFile = VfsUtil.loadText(file.getVirtualFile());
+      if (branchNamesFromFile.indexOf(processedBranchName) != branchNamesFromFile.lastIndexOf(processedBranchName)) {
+        holder.newAnnotation(HighlightSeverity.ERROR,
+            getNonHtmlString("string.GitMachete.MacheteAnnotator.branch-entry-already-defined")
+                .format(processedBranchName))
+            .range(TextRange.from(branchNamesFromFile.lastIndexOf(processedBranchName), processedBranchName.length()))
+            .create();
+      }
+    } catch (PluginException | IllegalStateException ignored) { // ignore dubious IDE checks against annotation range
```
PawelLipski, line 101: 😯 what are the examples of such exceptions? if uncaught, do they end up in an error notification (like, a one that has a link to open an issue on GitHub etc.)?

PawelLipski:
> so now only the last occurrence is highlighted. the question is if it should be

I'd vote for **all**, as the user might want to remove the first occurrence as well, so better to highlight as well

PawelLipski:
>  Ive tested it locally and for me it looks like the law of conservation of bugs is maintained, great work

Wait what's the **incoming** bug to fill up the place of #1195? 🤔

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/highlighting/MacheteAnnotator.java
```diff
@@ -80,6 +86,28 @@ private void processMacheteGeneratedEntry(MacheteGeneratedEntry macheteEntry, An
...
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
+  }
```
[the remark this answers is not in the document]
PawelLipski: If this is executed on EDT, then running `@UIThreadUnsafe` (aka heavyweight/potentially blocking) actions needs to be reconsidered (on separate PR ofc, this is a pre-existing issue).

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/highlighting/MacheteAnnotator.java
```diff
@@ -80,6 +86,28 @@ private void processMacheteGeneratedEntry(MacheteGeneratedEntry macheteEntry, An
...
+
+  @UIEffect
+  private void saveDocumentBeforeCheck(PsiFile file) {
+    val fileDocManager = FileDocumentManager.getInstance();
+    fileDocManager.saveDocument(Objects.requireNonNull(fileDocManager.getDocument(file.getVirtualFile())));
```
PawelLipski, line 110: Hmm if `fileDocManager.getDocument(...)` is null, then a runtime exception is thrown right? 🤔 do we have a guarantee that it'll indeed always be non-null? I'd add some `if (... != null)` if not

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/highlighting/MacheteAnnotator.java
LukaszKontowski: I encountered two types of such exceptions (which are not valid in my opinion):

1. `com.intellij.diagnostic.PluginException` coming from the `com.intellij.lang.annotation.AnnotationBuilder#range(@NotNull TextRange range)` - appears when adding a duplicate entry. It says something like:
"Range must be inside element being annotated: " + currentElementRange + "; but got: " + range;
Where `currentElementRange` is expected as valid and `range` is the range defined in our code. The cause of this exception is the fact, that there is some race condition in mentioned `#range` method. In particular - it's implementation uses a variable `private final PsiElement myCurrentElement;` which often points to an element that is **not** currently checked entry. It might point to any other entry in the machete file.

2. `IllegalStateException` - that one I did not study much. It sometimes occurs when deleting a duplicate line from the file. Anyway, it did not say anything helpful but caused an error in IDE.

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/highlighting/MacheteAnnotator.java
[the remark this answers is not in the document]
mkondratek: but. what is related to this PR - I am afraid that we may produce too much notification spam about unmanaged branches. having the proper highlight of duplicates we should consider the drop of the notifications. 

just thinking ... intellij do not send notifications about variable already defined in the scope. however there will be compilation error in that case. 

we could go for not showing the notifications while the file is opened but it seems to look inconsistent - I dont like it.

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/highlighting/MacheteAnnotator.java
LukaszKontowski: > if uncaught, do they end up in an error notification

Yes, there is a red notification saying that "IDE error occured". You can click on it and see details.

image

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/highlighting/MacheteAnnotator.java
LukaszKontowski: In my opinion, we do should not add additional check. Reasoning:
1. This method is called from within `private void processMacheteGeneratedEntry` method
2. The `private void processMacheteGeneratedEntry` method creates a PsiFile object earlier as follows:
`PsiFile file = macheteEntry.getContainingFile();`
3. The `PsiFile file` parameter used in this method is the same `file` object
4. Therefore, on this stage of the `processMacheteGeneratedEntry` logic, `file` object should not be null. If it would be null, then an exception would be thrown earlier, for example here:
`val branchNames = file.getBranchNamesForPsiMacheteFile();`
5. `file` should not be null, so `fileDocManager.getDocument(...)` should not be null too.
6. If for some reason `file` would be in fact a `null` at the moment this helper method is called - then a `NullPointerException` will occur. But this should be good in my opinion. We should get an exception, if something has been broken unexpectedly with the file.

What do you think?

LukaszKontowski:
> so now only the last occurrence is highlighted. the question is if it should be
> - all
> - second and later

All occurrences should be highlighted now. Just fixed it.

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/highlighting/MacheteAnnotator.java
LukaszKontowski: Changed the content of this comment. In fact, this is not a comment which tells about the method itself, so it should not be a javadoc. This is just an explanation, **why** we need to update the state of the .git/machete VirtualFile before performing the actual check within the `processMacheteGeneratedEntry` logic.

mkondratek:
> > Ive tested it locally and for me it looks like the law of conservation of bugs is maintained, great work
> 
> Wait what's the **incoming** bug to fill up the place of #1195? 🤔

I cannot wait to find it out! 🤞

mkondratek:
the highlight seems to work for all repeated branches indeed now.
However I see that **the law of conservation of bugs** has not been maintained - I see **two** new problems:

image

1. first develop is not highlight - but possible its okay since the formating is wrong  
2. dev is not already defined

mkondratek (changes requested):
image

LukaszKontowski:
> dev is not already defined

fixed now @mkondratek - compares branch names properly.
Indeed, the previous logic was very dummy

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/highlighting/MacheteAnnotator.java
```diff
@@ -80,6 +87,38 @@ private void processMacheteGeneratedEntry(MacheteGeneratedEntry macheteEntry, An
...
+  }
+
+  private boolean isBranchNameRepeated(String fileContent, String branchName) {
```
mkondratek, line 114: oh, yes. comparing to the previous this one is much more reasonable

mkondratek (changes requested):
but there is som thing wong still  :c 

image

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/highlighting/MacheteAnnotator.java
PawelLipski: Okay! let's leave as is then

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/highlighting/MacheteAnnotator.java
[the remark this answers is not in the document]
PawelLipski: Checked, behaves as expected 🎉

PawelLipski:
> but there is som thing wong still :c

image

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/highlighting/MacheteAnnotator.java
```diff
@@ -80,6 +87,38 @@ private void processMacheteGeneratedEntry(MacheteGeneratedEntry macheteEntry, An
                   .format(processedBranchName))
           .range(branch).create();
     }
+
+    // update the state of the .git/machete VirtualFile so that new entry is available in the VirtualFile
```
PawelLipski, line 91: Hmmm but note that the lambda passed as param to `invokeLaterIfNeeded` executes on **UI thread**, while the entire `processMacheteGeneratedEntry` method executes on a different, **non-UI thread**.

This means that there's **no** guarantee that at the point when you get to `VfsUtil.loadText(file.getVirtualFile());`, `saveDocumentBeforeCheck(file)` is already complete. UI thread might be busy with other operations, and it might take while for the execution of `saveDocumentBeforeCheck(file)` to start 🤔

PawelLipski: I don't have a clear idea yet how to properly do this stuff... how about moving the entire logic in `try { ... } catch { ... }` to `saveDocumentBeforeCheck()`, so that it executes only **after** `fileDocManager.saveDocument(...)` is complete, on the same (UI) thread? 🤔

PawelLipski: Ofc then `saveDocumentBeforeCheck()` would need to be renamed adequately

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/highlighting/MacheteAnnotator.java
```diff
@@ -80,6 +87,38 @@ private void processMacheteGeneratedEntry(MacheteGeneratedEntry macheteEntry, An
...
+    ModalityUiUtil.invokeLaterIfNeeded(NON_MODAL, () -> saveDocumentBeforeCheck(file));
+    // check for duplicate entries in the machete file
+    try {
+      val branchNamesFromFile = VfsUtil.loadText(file.getVirtualFile());
```
PawelLipski, line 95: For clarity, the purpose of `fileDocManager.saveDocument(...)` is so that `VfsUtil.loadText(...)` returns an up-to-date value here? 🤔

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/highlighting/MacheteAnnotator.java
```diff
@@ -80,6 +87,38 @@ private void processMacheteGeneratedEntry(MacheteGeneratedEntry macheteEntry, An
...
+      }
+    } catch (PluginException | IllegalStateException ignored) { // ignore dubious IDE checks against annotation range
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  @UIEffect
+  private void saveDocumentBeforeCheck(PsiFile file) {
+    val fileDocManager = FileDocumentManager.getInstance();
```
PawelLipski, line 110: Actually... does `fileDocManager.saveDocument(...)` actually save the file **to disk**?

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/highlighting/MacheteAnnotator.java
```diff
@@ -80,6 +87,38 @@ private void processMacheteGeneratedEntry(MacheteGeneratedEntry macheteEntry, An
...
+    fileDocManager.saveDocument(Objects.requireNonNull(fileDocManager.getDocument(file.getVirtualFile())));
+  }
+
+  private boolean isBranchNameRepeated(String fileContent, String branchName) {
+    boolean result = false;
+    java.util.List<String> lines = Arrays.stream(fileContent.split(System.lineSeparator())).map(String::trim)
```
PawelLipski, line 116: Won't the current impl also catch repeated **empty lines**?

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/highlighting/MacheteAnnotator.java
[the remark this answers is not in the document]
PawelLipski: > having the proper highlight of duplicates we should consider the drop of the notifications.

Hmmm isn't the duplicate branch notification currently only displayed when **git machete status is refreshed**? 🤔 which I think is still acceptable

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/highlighting/MacheteAnnotator.java
LukaszKontowski: Yes. Without it:
- `VfsUtil.loadText` "_doesn't know_" that there is some new entry in the file. 
- `val branchNamesFromFile` would not contain the new added entry

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/highlighting/MacheteAnnotator.java
LukaszKontowski: yes - `Saves the specified document to disk` is the first thing from Javadoc

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/highlighting/MacheteAnnotator.java
LukaszKontowski: Hm, later on we have a filter: `filter(line -> line.equals(branchName))` which should filter out empty lines IMO. However, for a while I cannot test it myself due to problems with `java.lang.NoClassDefFoundError: Could not initialize class com.intellij.ui.mac.foundation.Foundation` (mentioned on slack) 😢 
Hope I will be able to check it soon

LukaszKontowski: Ok, I checked it. My logic works fine. Blank lines are not treated as branches. We do not catch them.

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/highlighting/MacheteAnnotator.java
LukaszKontowski: Ok, I tried it. Unfortunately, it does not work. If we add both saving file and performing the check in the 
`ModalityUiUtil.invokeLaterIfNeeded(NON_MODAL, () -> ...`
block - then duplicate entries are not caught for some reason. Let's leave it as it is IMO. It works well.

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/highlighting/MacheteAnnotator.java
LukaszKontowski: > but there is som thing wong still :c
> 
> image

@mkondratek - I did some more tests on my machine... And all works fine. Can you repeat your test and let me know, if the lack of highlighting occurred again?

LukaszKontowski:
> but there is som thing wong still :c
> 
> image

@mkondratek Can you please test it again and let me know if lack of highlighting happened again? I tried testing multiple times and worked well every time. Also with the same scenario as on your screenshot.

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/highlighting/MacheteAnnotator.java
```diff
@@ -80,6 +87,34 @@ private void processMacheteGeneratedEntry(MacheteGeneratedEntry macheteEntry, An
...
+    } catch (PluginException | IllegalStateException ignored) { // ignore dubious IDE checks against annotation range
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  @UIEffect
+  private void saveDocumentBeforeCheck(PsiFile file) {
+    val fileDocManager = FileDocumentManager.getInstance();
+    fileDocManager.saveDocument(Objects.requireNonNull(fileDocManager.getDocument(file.getVirtualFile())));
```
PawelLipski, line 111: Ok one more thing, from javadoc of `com.intellij.openapi.fileEditor.FileDocumentManager#getDocument`:

```
Should be invoked in a read action.
```

Either fix it now (https://plugins.jetbrains.com/docs/intellij/general-threading-rules.html#read-write-lock) or at least add a `TODO` for https://github.com/VirtusLab/git-machete-intellij-plugin/issues/1227

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/highlighting/MacheteAnnotator.java
PawelLipski: Okay, let's leave as is... maybe just add a comment about the potential caveat (as explained above in https://github.com/VirtusLab/git-machete-intellij-plugin/pull/1216/files#r989323712) in case we observe any race conditions in the future

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/highlighting/MacheteAnnotator.java
LukaszKontowski: Hm, for me it looks like current logic is valid in that aspect. Reasoning:
1. From the https://plugins.jetbrains.com/docs/intellij/general-threading-rules.html#read-write-lock guide:
> Reading data is allowed from any thread. Reading data **from the UI thread does not require any special effort.**
2. `saveDocumentBeforeCheck` method is annotated with `@UIEffect` and (in effect) it runs on the Event Dispatch Thread (checked it with SwingUtilities#isEventDispatchThread)
3. So, it looks like  `com.intellij.openapi.fileEditor.FileDocumentManager#getDocument` called inside `saveDocumentBeforeCheck` reads data from the UI thread, and thus it does not have to be wrapped in a special read action.

Does it make sense?

PawelLipski: Awww okay good spot, I forgot that on UI thread one doesn't need ReadActions. So good to go as is

PawelLipski:
Huh same problem with lack of red squiggles for me as well... just a single IntelliJ opened, plugin installed from zip (from CI artfifacts):

image

Same if the branch is repeated just twice :/

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/highlighting/MacheteAnnotator.java
```diff
@@ -80,6 +87,34 @@ private void processMacheteGeneratedEntry(MacheteGeneratedEntry macheteEntry, An
...
+  }
+
+  private boolean isBranchNameRepeated(String fileContent, String branchName) {
+    java.util.List<String> lines = Arrays.stream(fileContent.split(System.lineSeparator())).map(String::trim)
+        .collect(Collectors.toList());
```
PawelLipski, line 116: Awww I see where the problem lies 🤦🏻 (at least partially)

This algorithm for detecting duplicates does NOT take into account custom annotations:
```
foo
  bar PR #123
  bar PR #456
```
Then `isBranchNameRepeated(..., "bar")` will return 0️⃣  (as there are no lines which contain exactly `bar` after trimming).

I'd suggest the following:
1. if possible, rely on **parsed data** delivered by IntelliJ APIs, rather than do rudimentary parsing directly in this method

and/or

2. given that a non-trivial bug has been detected on this PR, pls add a regression test for the logic of detecting the duplicates (unless that requires heavily mocking IntelliJ APIs)

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/highlighting/MacheteAnnotator.java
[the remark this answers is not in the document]
PawelLipski: Huh now I see that it's displayed all the time during the edit... Not Good ™️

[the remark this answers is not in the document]
PawelLipski: https://github.com/VirtusLab/git-machete-intellij-plugin/issues/1238

mkondratek:
@LukaszKontowski, I see :pawlacz: just reviewed it and found a new problem 😆 I will re-review it once the mentioned is fixed

LukaszKontowski:
@PawelLipski 
Problems with comments - fixed in https://github.com/VirtusLab/git-machete-intellij-plugin/pull/1216/commits/154819313b3e0ed0b273de5a0b68f538aa409a1c

In my opinion this is enough. 
Reasoning: (correct me if I am wrong) :
- each line in the machete file might contain only one branch name (if it is not a blank line)
- our parser gets the first _word_ from the line and treats it as the branch name
- if there is a whitespace character after branch name on the same line, **and** after this whitespace (space/tab) there are some other words - they are all treated as a comment and not as a part of branch name

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/highlighting/MacheteAnnotator.java
```diff
@@ -80,6 +87,36 @@ private void processMacheteGeneratedEntry(MacheteGeneratedEntry macheteEntry, An
...
+  @UIEffect
+  private void saveDocumentBeforeCheck(PsiFile file) {
+    val fileDocManager = FileDocumentManager.getInstance();
+    fileDocManager.saveDocument(Objects.requireNonNull(fileDocManager.getDocument(file.getVirtualFile())));
+  }
+
+  private boolean isBranchNameRepeated(String fileContent, String branchName) {
+    java.util.List<String> lines = Arrays.stream(fileContent.split(System.lineSeparator()))
+        .map(String::trim)
+        .map(line -> line.split("\\s")[0]) // ignore possible comments after branch name
```
PawelLipski, line 117:
```suggestion
        .map(line -> line.split("\\s")[0]) // ignore possible custom annotation after branch name
```

PawelLipski:
> @PawelLipski Problems with comments - fixed in [1548193](https://github.com/VirtusLab/git-machete-intellij-plugin/commit/154819313b3e0ed0b273de5a0b68f538aa409a1c)
> 
> In my opinion this is enough. Reasoning: (correct me if I am wrong) :
> 
> * each line in the machete file might contain only one branch name (if it is not a blank line)
> * our parser gets the first _word_ from the line and treats it as the branch name
> * if there is a whitespace character after branch name on the same line, **and** after this whitespace (space/tab) there are some other words - they are all treated as a comment and not as a part of branch name

Okay, should be enough indeed... .git/machete file format is extremely unlikely to ever change in non-backward-compatible manner anyway ;)

LukaszKontowski:
Ok, @mkondratek - issue reported by Paweł is fixed now. Please, re-review 😸

PawelLipski (approved):
LGTM! let @mkondratek approve as well

mkondratek (approved):
lgtm 🚀

PR #1224, opened 2022-10-05
Bugfix/uitest race condition fix

MaciejG604:
Turns out the failing tests were the result of reading the _machete_ file right when it was being overwritten by the `UiTestSuite` which made the file empty for a short period of time. Code reacts to empty machete file by doing a rediscover thus the rest of the test fails.

### DIFF
--- a/frontend/resourcebundles/src/main/resources/GitMacheteBundle.properties
+++ b/frontend/resourcebundles/src/main/resources/GitMacheteBundle.properties
@@ -300,10 +300,10 @@ string.GitMachete.EnhancedGraphTable.automatic-discover.notification.title.canno
 string.GitMachete.EnhancedGraphTable.automatic-discover.task-title=Automatically discovering branch layout
 action.GitMachete.EnhancedGraphTable.automatic-discover.slide-out-skipped=Slide out skipped branches
 string.GitMachete.EnhancedGraphTable.automatic-discover.success-message=Branch layout has been automatically discovered. Edit the layout manually if needed.
-string.GitMachete.EnhancedGraphTable.empty-table-text.try-running-discover=Provided machete file ({0}) is empty, try running ''Discover Branch Layout''
 string.GitMachete.EnhancedGraphTable.empty-table-text.cannot-discover-layout=Provided machete file ({0}) is empty and branch layout can''t be automatically detected
 string.GitMachete.EnhancedGraphTable.empty-table-text.loading=Loading\u2026
 string.GitMachete.EnhancedGraphTable.empty-table-text.only-skipped-in-machete-file=None of the branches provided by machete file ({0}) exists within the local repository
+string.GitMachete.EnhancedGraphTable.empty-table-text.try-running-discover=Provided machete file ({0}) is empty, try running ''Discover Branch Layout''
 string.GitMachete.EnhancedGraphTable.duplicated-branches-text=The following branches appear more than once in machete file:
 string.GitMachete.EnhancedGraphTable.skipped-branches-text=Some branches listed in machete file do not exist: {0}
 action.GitMachete.EnhancedGraphTable.branch-layout-write-failure=Writing new branch layout failed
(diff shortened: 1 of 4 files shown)


src/uiTest/scala/com/virtuslab/gitmachete/uitest/UITestSuite.scala
```diff
@@ -72,6 +71,7 @@ class UITestSuite extends BaseGitRepositoryBackedIntegrationTestSuite(SETUP_WITH
         |master
         |  hotfix/add-trigger""".stripMargin
     )
+    project.openGitMacheteTab()
```
PawelLipski, line 74: Actually... now after a consideration, I think this still isn't a proper solution :/

I'd rather refine the condition for autodiscover to either:
* **not** run the autodiscover when .git/machete is **empty** (only run when it's **missing**)
* or in case the file is empty, check its modification date and refrain from autodiscover if it's modified too recently

PawelLipski: Still, lemme smuggle the current workaround of yours onto #1167 since I'm tired of re-running the pipelines :trollface:

mkondratek: IIRC we had some issue related to sliding out all branches - I do not remember what was the exact problem and how it behaves now. I worry that even though the first option seems to be more reasonable there will be some complicating edge case. anyway I think the effort is not significant and we can try it

PawelLipski: > we had some issue related to sliding out all branches

Hmmm I think changing the logic to _run autodiscover only when .git/machete is missing, not when empty_ will actually fix the problem with autodiscover happening automatically when someone slides out all branches (which is extremaly unlikely IRL) ;p coz then autodiscover simply won't run, .git/machete will remain empty

frontend/ui/impl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/EnhancedGraphTable.java
```diff
@@ -190,9 +190,10 @@ private void refreshModel(
       repositoryGraph = repositoryGraphCache.getRepositoryGraph(gitMacheteRepositorySnapshot, isListingCommits);
       if (gitMacheteRepositorySnapshot.getRootBranches().isEmpty()) {
         if (gitMacheteRepositorySnapshot.getSkippedBranchNames().isEmpty()) {
-          LOG.info("Machete file (${macheteFilePath}) is empty, so auto discover is running");
-          queueDiscover(macheteFilePath, doOnUIThreadWhenReady);
-          return;
+          LOG.info("Machete file (${macheteFilePath}) is empty");
+          setTextForEmptyTable(
+              getString("string.GitMachete.EnhancedGraphTable.empty-table-text.try-running-discover")
+                  .format(macheteFilePath.toString()));
```
PawelLipski, line 196: Shouldn't `return` be retained? 🤔

frontend/resourcebundles/src/main/resources/GitMacheteBundle.properties
```diff
@@ -300,6 +300,7 @@ string.GitMachete.EnhancedGraphTable.automatic-discover.notification.title.canno
 string.GitMachete.EnhancedGraphTable.automatic-discover.task-title=Automatically discovering branch layout
 action.GitMachete.EnhancedGraphTable.automatic-discover.slide-out-skipped=Slide out skipped branches
 string.GitMachete.EnhancedGraphTable.automatic-discover.success-message=Branch layout has been automatically discovered. Edit the layout manually if needed.
+string.GitMachete.EnhancedGraphTable.empty-table-text.try-running-discover=Provided machete file ({0}) is empty, try running ''Discover Branch Layout''
```
PawelLipski, line 303: Nit: keep abc order

PawelLipski:
Also:
- [ ] pls check if there's anything in `docs/features.md` to update
- [ ] an entry in CHANGE_NOTES.md might make sense for this change

PawelLipski (approved):

PR #1213, opened 2022-10-04
Traverse feature

vzmerr:
It resolves #246

PR #1230, opened 2022-10-06
Rebase on Remote

vzmerr:
This is an implementation for the preferred action during [traverse](https://github.com/VirtusLab/git-machete-intellij-plugin/pull/1213), when the remote tracking branch is diverged from the local one.


vzmerr:
> is it not a prerequisite for the traverse feature? b/c now the traverse is the base branch 🤔

@mkondratek 🤔 provided that we do want to include it in the traverse.
One alternative is to merge the traverse by refining it then compliment it with this feature afterwards. WDYT?
If we want to make this one the base of traverse, we have to provide a menu item for Remote on Rebase, in order to be able to call it independently.
Still, this feature is far more complicated than the traverse skeleton.

mkondratek:
hmmm. probably I still do not understand it correctly but

the facts that I assume are the facts XD:
- this one is required for traverse -> it should be done before 
- we can add action without placing it in the toolbar or context menu 
- the "complication" of the task does not change the order of what must be done (no offense!  I am just thinking 🦆 )

am I right? 🙂

vzmerr:
> hmmm. probably I still do not understand it correctly but
> 
> the facts that I assume are the facts XD:
> 
> * this one is required for traverse -> it should be done before
> * we can add action without placing it in the toolbar or context menu
> * the "complication" of the task does not change the order of what must be done (no offense!  I am just thinking 🦆 )
> 
> am I right? 🙂

ok, no problem, so I would base traverse over this instead.

PawelLipski:
> oops, this PR got automatically closed by GitHub when I rebased the traverse PR on it

Hmmm still a problem? 🤔 

I can see now there are 2 related PRs: #1235, and on top of it, #1213... so this PR #1230 can probably remain closed right?

PR #1235, opened 2022-10-06
rebase on remote

vzmerr:
this PR is created as a prerequisite for the traverse feature.


PawelLipski:
Hmmm okay so a high-level question... **why** is that a prerequisite for traverse? 🤔 will traverse, instead of using the existing rebase logic (i.e. onto parent, taking commits from after fork point) use the logic of rebasing onto remote tracking branch instead?

PawelLipski:
Okay... I see that within #1235, `Rebase on Remote` is suggested for branches that are diverged from their remote, right? 🤔 in general, git-machete suggests either:
* `git push --force`, if local branch points to **newer** commit than remote or
* `git reset --keep <remote-branch>`, if local branch points to **older** commit than remote

Generally, offering rebase onto remote as a third option would be viable... but I'd rather keep git-machete as opinionated as possible wrt. what's suggested to users.

In fact, it'd best to first analyze the sequence of events that might lead to a situation where a user wants to rebase their local branch onto its remote counterpart... esp. who's the original author of that branch? who's the last person to push commits to that branch?

vzmerr:
> sequence of events that might lead to a situation where a user wants to rebase their local branch onto its remote counterpart

committing some reviews on the remote branch, then also committing some fixes or a half feature I was working on locally.

backend/api/src/main/java/com/virtuslab/gitmachete/backend/api/IRootManagedBranchSnapshot.java
```diff
@@ -17,4 +17,9 @@ default IRootManagedBranchSnapshot asRoot() {
   default INonRootManagedBranchSnapshot asNonRoot() {
     throw new NotImplementedError();
   }
+
+  @Override
+  default boolean isTracked() {
+    throw new NotImplementedError();
```
PawelLipski, line 23: 🤔

PawelLipski: Shouldn't rebase onto remote tracking branch be allowed for a root branch as well?

backend/impl/src/main/java/com/virtuslab/gitmachete/backend/impl/GitMacheteRepository.java
```diff
@@ -481,15 +483,89 @@ private CreatedAndDuplicatedAndSkippedBranches<NonRootManagedBranchSnapshot> cre
...
       val result = new NonRootManagedBranchSnapshot(branchName, branchFullName, childBranches.getCreatedBranches(),
-          pointedCommit, remoteTrackingBranch, relationToRemote, customAnnotation, statusHookOutput, forkPoint,
+          pointedCommit, remoteTrackingBranch, relationToRemote, customAnnotation, statusHookOutput, forkPoint, remoteForkPoint,
           commits.map(CommitOfManagedBranch::new), syncToParentStatus);
       return CreatedAndDuplicatedAndSkippedBranches.of(List.of(result),
           childBranches.getDuplicatedBranchNames(), childBranches.getSkippedBranchNames());
     }
 
+    @UIThreadUnsafe
+    private @Nullable ForkPointCommitOfManagedBranch deriveRemoteAwareForkPoint(IGitCoreLocalBranchSnapshot coreLocalBranch)
```
PawelLipski, line 499: Okay... so in a nutshell, how does this algorithm compare to the existing fork point algorithm? 🤔

vzmerr: it only checks the reflog of the remote

vzmerr: the current implementation is a tad bit inefficient though.
There is a bug in the refactoring of the existing refactor functionality as well. I would look into that tomorrow.

mkondratek:
1. As we have discussed it somewhere - rebase on remote is not required for traverse it; it's also not a priority now then
2. As we have discussed it somewhere (@zmerr finding) - rebase on remote is implemented by git4idea - possibly we could re-use it 
3.  these changes are outdated now 

closing! 📕

PR #1236, opened 2022-10-06
Add javadocs to annotations under `qual/`

PawelLipski:


config/checker/com.intellij.astub
```diff
@@ -90,10 +90,10 @@ class PushController {
 class DataManager {
 
   @SafeEffect
-  public static DataManager getInstance();
```
PawelLipski, line 93: Unrel; `public` in Checker stubs doesn't matter anyway

config/checker/javax.swing.astub
```diff
@@ -41,6 +41,12 @@ class JTable {
   void initializeLocalVars(@UnknownInitialization JTable this);
 }
 
+class SwingUtilities {
+  @SafeEffect
+  static boolean isEventDispatchThread();
```
PawelLipski, line 46: Just realized today that shouldn't block calling this method from outside UI thread

qual/src/main/java/com/virtuslab/qual/guieffect/UIThreadUnsafe.java
```diff
@@ -5,9 +5,26 @@
...
 import java.lang.annotation.Target;
 
-// TODO (typetools/checker-framework#3252): replace with proper @Heavyweight
+import org.checkerframework.checker.guieffect.qual.UIEffect;
 
-// Must have runtime retention to be visible to ArchUnit tests.
+/**
+ * Used to mark methods that should NOT be executed on UI thread
+ * (in case of IntelliJ Platform, on Swing's Event Dispatch Thread aka EDT)
+ * to avoid a UI freeze.
```
LukaszKontowski, line 13:
```suggestion
 * Used to mark methods that should NOT be executed on UI thread to avoid a UI freeze.
 * In case of IntelliJ Platform, the UI thread is Swing's Event Dispatch Thread aka EDT.
```

qual/src/main/java/com/virtuslab/qual/guieffect/UIThreadUnsafe.java
```diff
@@ -5,9 +5,26 @@
...
+ * These include methods that are heavyweight or blocking: typically, this means accessing disk and/or network.
```
LukaszKontowski, line 14:
```suggestion
 * These include methods that are heavyweight or blocking. Typically, this means accessing disk and/or network.
```

qual/src/main/java/com/virtuslab/qual/guieffect/UIThreadUnsafe.java
```diff
@@ -5,9 +5,26 @@
...
+ * <p>
+ * Whether the annotation is used correctly is currently NOT enforced during compilation or runtime.
+ * TODO (typetools/checker-framework#3252): replace with proper {@code @Heavyweight} annotation.
+ * As for now, we use a ArchUnit test {@code com.virtuslab.archunit.UIThreadUnsafeMethodInvocationsTestSuite}
```
LukaszKontowski, line 18:
```suggestion
 * As for now, we use an ArchUnit test {@code com.virtuslab.archunit.UIThreadUnsafeMethodInvocationsTestSuite}
```
should be "an" instead of "a" ?

qual/src/main/java/com/virtuslab/qual/guieffect/UIThreadUnsafe.java
```diff
@@ -5,9 +5,26 @@
...
+import org.checkerframework.checker.guieffect.qual.UIEffect;
 
-// Must have runtime retention to be visible to ArchUnit tests.
+/**
+ * Used to mark methods that should NOT be executed on UI thread
+ * (in case of IntelliJ Platform, on Swing's Event Dispatch Thread aka EDT)
+ * to avoid a UI freeze.
+ * These include methods that are heavyweight or blocking: typically, this means accessing disk and/or network.
+ * <p>
+ * Whether the annotation is used correctly is currently NOT enforced during compilation or runtime.
```
LukaszKontowski, line 16: This sentence sounds a bit strange to me. I think that: 
- it should start with something like "It is NOT enforced during compilation that ... " (drop `currently`, it does not add a value IMO)
- Maybe I'm wrong, but it seems to me that in case of annotations - if something should be enforced/checked/prohibited etc. - then it should happen in compilation - and **not** in runtime... So skip the "or runtime" part?

LukaszKontowski, line 16: This sentence sounds a bit strange to me. I think that: 
- it should start with something like "It is NOT enforced during compilation that ... " and then finish with the "whether the..." part. (+ drop `currently`, it does not add a value IMO)
- Maybe I'm wrong, but it seems to me that in case of annotations - if something should be enforced/checked/prohibited etc. - then it should happen in compilation - and **not** in runtime... So skip the "or runtime" part?

qual/src/main/java/com/virtuslab/qual/guieffect/IgnoreUIThreadUnsafeCalls.java
```diff
@@ -5,7 +5,18 @@
...
-// Must have runtime retention to be visible to ArchUnit tests.
+/**
+ * Used to mark methods that get a free pass on calling {@link UIThreadUnsafe} methods,
+ * despite NOT being marked as {@link UIThreadUnsafe} themselves.
+ * <p>
+ * This is only taken into account by ArchUnit test {@code com.virtuslab.archunit.UIThreadUnsafeMethodInvocationsTestSuite}
+ * that enforces the correct usage of {@link UIThreadUnsafe} annotation.
+ * <p>
+ * Needs to be used sparingly, as this basically allows for a method to call potentially heavyweight operations on UI thread,
+ * which might lead to UI freeze.
```
LukaszKontowski, line 16:
```suggestion
 * Needs to be used sparingly, as this basically allows for a method to call potentially heavyweight operations on UI thread.
 * And that might lead to UI freeze.
```

LukaszKontowski:
@PawelLipski - great job with these javadocs for 4 annotations. 👍 But it looks like there are still at least 8 more annotations that are defined in our codebase that would need Javadoc comments as well?
image

qual/src/main/java/com/virtuslab/qual/guieffect/UIThreadUnsafe.java
PawelLipski: All 3 done

PawelLipski:
> But it looks like there are still at least 8 more annotations defined in our codebase that would need Javadoc comments as well?

All done ✍🏻

LukaszKontowski (approved):

PR #1217, opened 2022-10-04
Prevent NPE in `OpenMacheteFileAction`

PawelLipski:

Issue #430, opened 2020-07-08
Expose `getParent` from the `BranchLayoutEntry`


mkondratek:
Hmm consider exposing `getParent` from the `BranchLayoutEntry`... just as done for `GitMacheteNonRootBranch`, or on #415 for `MyBranchLayoutEntry`

_Originally posted by @PawelLipski in https://github.com/VirtusLab/git-machete-intellij-plugin/pull/407_

PawelLipski:
Taking this one over myself...

PR #1243, opened 2022-10-11
Feature/alt enter to create non existing branch

### LINKED ISSUE #429: Machete file - provide Alt + Enter option to create non-existing branch
Scenario:
1. Open machete file with IDE
2. Add new branch layout entry (root?/non-root) - a branch name that does not exist within the repo
// Newly added entry gets red underlining thanks to our syntax highlight.
// The red bulb icon appears & the `Alt + Enter` option to resolve the "issue" is available.
3. Press `Alt + Enter` and select `Create new branch`
4. Red underlining disappears and the new branch is  created 

TBD whether:
 - `new branch dialog` is required
 - the new branch shall be checked out

LukaszKontowski:
Closes #429


frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/grammar/MacheteParserDefinition.java
```diff
@@ -28,8 +28,8 @@ public TokenSet getWhitespaceTokens() {
     return WHITE_SPACES;
   }
 
-  // Even if we doesn't support comments this element is part of ParserDefinition interface,
-  // so we need to provide something here
+  // Even if we don't support comments, this element is part of the ParserDefinition interface.
+  // So - we need to provide something here.
```
LukaszKontowski, line 32: unrelated - small fix done by the way

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/MacheteFileUtils.java
```diff
@@ -30,13 +32,7 @@ public static String getSampleMacheteFileContents() {
 
   @UIThreadUnsafe
   public static List<String> getBranchNamesForPsiMacheteFile(PsiFile psiFile) {
-    val project = psiFile.getProject();
-
-    val gitRepository = List.ofAll(GitRepositoryManager.getInstance(project).getRepositories())
-        .find(repository -> {
-          val macheteFile = repository.getMacheteFile();
-          return macheteFile != null && macheteFile.equals(psiFile.getVirtualFile());
-        });
```
LukaszKontowski, line 39: extracting to `findGitRepositoryForPsiMacheteFile` method as it would be useful in the `CreateBranchQuickFix` class in another file

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/highlighting/MacheteAnnotator.java
```diff
@@ -80,23 +84,16 @@ private void processMacheteGeneratedEntry(MacheteGeneratedEntry macheteEntry, An
...
     val processedBranchName = branch.getText();
 
-    if (!branchNames.contains(processedBranchName)) {
-      holder
-          .newAnnotation(HighlightSeverity.ERROR,
-              getNonHtmlString("string.GitMachete.MacheteAnnotator.cannot-find-local-branch-in-repo")
-                  .format(processedBranchName))
-          .range(branch).create();
-    }
-
```
LukaszKontowski, line 90: Moving this check after the check for duplicate entries. So that `saveDocumentBeforeCheck(file)` should be completed before performing it (file content is used in updated logic).

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/highlighting/MacheteAnnotator.java
```diff
@@ -123,6 +134,27 @@ private boolean isBranchNameRepeated(String fileContent, String branchName) {
     return lines.stream().filter(line -> line.equals(branchName)).count() > 1;
   }
 
+  @SuppressWarnings("allcheckers")
+  private String getParentBranchName(String fileContent, String branchName) {
```
PawelLipski, line 138: Huh I'd reeeealy consider using `BranchLayoutReader` instead of custom logic 😅

LukaszKontowski: Ok, done :)

branchLayout/api/src/main/java/com/virtuslab/branchlayout/api/BranchLayout.java
```diff
@@ -30,6 +33,20 @@ private static List<IBranchLayoutEntry> collectEntriesRecursively(IBranchLayoutE
     return entryByName.get(branchName).getOrNull();
   }
 
+  @Override
+  public @Nullable IBranchLayoutEntry findDirectParentBranch(@NonNull IBranchLayoutEntry actualEntry,
```
PawelLipski, line 37: Hmm what's the meaning of `actualEntry` and `currentEntry`? pls add javadoc as it isn't clear

branchLayout/api/src/main/java/com/virtuslab/branchlayout/api/BranchLayout.java
```diff
@@ -30,6 +33,20 @@ private static List<IBranchLayoutEntry> collectEntriesRecursively(IBranchLayoutE
...
+      @NonNull IBranchLayoutEntry currentEntry) {
```
PawelLipski, line 38: `@NonNull` should never be needed

branchLayout/api/src/main/java/com/virtuslab/branchlayout/api/BranchLayout.java
PawelLipski: This method won't be needed soon... I'm almost ready with https://github.com/VirtusLab/git-machete-intellij-plugin/issues/430 ;)

LukaszKontowski: Hm, so maybe I'll wait for your solution to be merged to develop. Then I will update my logic to use your solution.

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/highlighting/MacheteAnnotator.java
```diff
@@ -123,6 +148,19 @@ private boolean isBranchNameRepeated(String fileContent, String branchName) {
     return lines.stream().filter(line -> line.equals(branchName)).count() > 1;
   }
 
+  private String getParentBranchName(Path macheteFilePath, String branchName) throws BranchLayoutException {
```
PawelLipski, line 151: While you're at this, pls also get rid of custom parser from the recent code for detecting duplicate branches (can be on a separate PR)

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/highlighting/MacheteAnnotator.java
```diff
@@ -123,6 +148,19 @@ private boolean isBranchNameRepeated(String fileContent, String branchName) {
...
+    val branchLayoutReader = RuntimeBinding.instantiateSoleImplementingClass(IBranchLayoutReader.class);
+    val branchLayout = branchLayoutReader.read(macheteFilePath);
+    val branchLayoutEntry = branchLayout.findEntryByName(branchName);
+    assert branchLayoutEntry != null;
+    val parentBranchEntry = branchLayout.findDirectParentBranch(branchLayoutEntry, branchLayoutEntry);
+    if (parentBranchEntry == null) {
+      return "";
```
PawelLipski, line 158: Hmmm can `null` be returned from this method instead? 🤔 or does empty string make more sense here?

LukaszKontowski: Empty String makes more sense to me here. This method should return a String - and empty String is a sign, that evaluation went well, but no parent branch was found (branchName is for a root entry). 
If I can avoid returning `null` then I go for it. Less `null`s, less problems.

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/highlighting/MacheteAnnotator.java
```diff
@@ -123,6 +139,21 @@ private boolean isBranchNameRepeated(String fileContent, String branchName) {
...
   }
 
+  private String getParentBranchName(PsiFile file, String branchName) throws BranchLayoutException {
+    val branchLayoutReader = RuntimeBinding.instantiateSoleImplementingClass(IBranchLayoutReader.class);
+    val branchLayout = branchLayoutReader.read(Path.of(file.getVirtualFile().getPath()));
+    BranchLayoutEntry parentEntry;
+    try {
+      parentEntry = Objects.requireNonNull(branchLayout.getEntryByName(branchName)).getParent();
+    } catch (NullPointerException e) { // might appear if saveDocumentBeforeCheck(file) has not completed yet
+      return "";
```
LukaszKontowski, line 149: This is a workaround for scenarios where `saveDocumentBeforeCheck(file)` has not completed before `getParentBranchName` is invoked. In such case we just return `""` which means that _checked branch has no parent_. In reality - new branch entry might have a parent, so returning `""` does not seem right. However - in practice, when `saveDocumentBeforeCheck(file)` is completed, then whole `MacheteAnnotator#annotate` gets invoked again... which results in valid result with proper parent branch name.

LukaszKontowski: So, it saves us from `NullPointerException` and **does not** break the logic in practice.

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/highlighting/MacheteAnnotator.java
LukaszKontowski: https://github.com/VirtusLab/git-machete-intellij-plugin/issues/1272

LukaszKontowski:
@PawelLipski Can we get rid of this error from checks?
`[ant:checkstyle] [ERROR] /root/git-machete-intellij-plugin/frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/quickfix/CreateBranchQuickFix.java:18:1: Illegal import - org.jetbrains.annotations.NotNull. [IllegalImport]`
In fact:
`import org.jetbrains.annotations.NotNull;`
is required as _CreateBranchQuickFix implements IntentionAction_ - and `org.jetbrains.annotations.NotNull` is used in a bunch of methods defined in `IntentionAction` interface.

PawelLipski:
> 

Huh okay... so compilation of `CreateBranchQuickFix` fails without that `@NotNull`? if so sure, remove from checkstyle config

LukaszKontowski:
> Huh okay... so compilation of CreateBranchQuickFix fails without that @NotNull? if so sure, remove from checkstyle config
Ok, it showed up that I can just remove usage of these annotations from our code. 👍

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/quickfix/CreateBranchQuickFix.java
```diff
@@ -0,0 +1,72 @@
...
+  @UIThreadUnsafe
+  private void createNewBranchFromParent(Project project) {
+    Option<GitRepository> gitRepositoryOption = MacheteFileUtils.findGitRepositoryForPsiMacheteFile(macheteFile);
+    if (gitRepositoryOption.isDefined()) {
+      GitBrancher.getInstance(project).createBranch(branch, Collections.singletonMap(gitRepositoryOption.get(), parentBranch));
+    } else {
+      throw new RuntimeException("Unable to create new branch due to: git repository not found for .git/machete file.");
+    }
+  }
+}
```
PawelLipski, line 72: Pls bump version to 3.2.0 (as it's a new feature) + update the changelog

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/quickfix/CreateBranchQuickFix.java
```diff
@@ -0,0 +1,72 @@
...
+
+  public CreateBranchQuickFix(String processedBranchName, String parentBranchName, PsiFile processedFile) {
+    branch = processedBranchName;
+    parentBranch = parentBranchName;
+    macheteFile = processedFile;
+  }
+
+  @Override
+  public @IntentionName String getText() {
+    return getNonHtmlString("action.GitMachete.BaseSlideInBelowAction.dialog.create-new-branch.title")
```
PawelLipski, line 38: I'd add a separate property solely for the purpose of this quick fix:
1. quick fix intentions should be written in _Sentence case_, not _Title Case_
2. there should be no need to concat the strings this way, just pass `branch` as an argument to format

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/quickfix/CreateBranchQuickFix.java
```diff
@@ -0,0 +1,72 @@
...
+import com.virtuslab.qual.guieffect.UIThreadUnsafe;
+
+@ExtensionMethod({GitMacheteBundle.class})
+public class CreateBranchQuickFix implements IntentionAction {
+
+  private final String branch;
+  private final String parentBranch;
+  private final PsiFile macheteFile;
+
+  public CreateBranchQuickFix(String processedBranchName, String parentBranchName, PsiFile processedFile) {
```
PawelLipski, line 30: Just `@RequiredArgsConstructor`

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/quickfix/CreateBranchQuickFix.java
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  @Override
+  public boolean isAvailable(Project project, Editor editor, PsiFile file) {
+    return true;
+  }
+
+  @Override
+  @UIThreadUnsafe
+  public void invoke(Project project, Editor editor, PsiFile file) throws IncorrectOperationException {
```
PawelLipski, line 54: `throws` probably redundant; an overridden method can always throw a **narrower** set of exceptions than its original base method (but never wider)

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/quickfix/CreateBranchQuickFix.java
```diff
@@ -0,0 +1,72 @@
...
+    return "Git Machete";
+  }
+
+  @Override
+  public boolean isAvailable(Project project, Editor editor, PsiFile file) {
+    return true;
+  }
+
+  @Override
+  @UIThreadUnsafe
```
PawelLipski, line 53: This is invoked outside of UI thread right?

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/quickfix/CreateBranchQuickFix.java
```diff
@@ -0,0 +1,72 @@
...
+    return false;
+  }
+
+  @UIThreadUnsafe
+  private void createNewBranchFromParent(Project project) {
+    Option<GitRepository> gitRepositoryOption = MacheteFileUtils.findGitRepositoryForPsiMacheteFile(macheteFile);
+    if (gitRepositoryOption.isDefined()) {
+      GitBrancher.getInstance(project).createBranch(branch, Collections.singletonMap(gitRepositoryOption.get(), parentBranch));
+    } else {
+      throw new RuntimeException("Unable to create new branch due to: git repository not found for .git/machete file.");
```
PawelLipski, line 69: Nit: use the actual path of `macheteFile`, this won't always be just `.git/machete` (https://git-machete.readthedocs.io/en/stable/#file)

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/MacheteFileUtils.java
```diff
@@ -45,4 +41,14 @@ public static List<String> getBranchNamesForPsiMacheteFile(PsiFile psiFile) {
     return List.ofAll(gitRepository.get().getInfo().getLocalBranchesWithHashes().keySet())
         .map(localBranch -> localBranch.getName());
   }
+
+  @UIThreadUnsafe
+  public static Option<GitRepository> findGitRepositoryForPsiMacheteFile(PsiFile psiFile) {
```
PawelLipski, line 46: Nit: we generally just return `@Nullable`s instead of `Option`s, they're typically more wieldy and &mdash; paradoxically &mdash; more null safe

PawelLipski: For instance, see https://github.com/VirtusLab/git-machete-intellij-plugin/pull/1217/ + associated issues, using `Option`s can **increase** the risk of NPEs

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/quickfix/CreateBranchQuickFix.java
```diff
@@ -0,0 +1,72 @@
...
+  @Override
+  public boolean startInWriteAction() {
+    return false;
+  }
+
+  @UIThreadUnsafe
+  private void createNewBranchFromParent(Project project) {
+    Option<GitRepository> gitRepositoryOption = MacheteFileUtils.findGitRepositoryForPsiMacheteFile(macheteFile);
+    if (gitRepositoryOption.isDefined()) {
+      GitBrancher.getInstance(project).createBranch(branch, Collections.singletonMap(gitRepositoryOption.get(), parentBranch));
```
PawelLipski, line 67: Hmmm would it be possible to trigger a reparsing (?) of the machete file once the branch is created? There seems to be a delay after creating the branch before the red squiggle disappears. Ofc skip if complicated, not that important

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/highlighting/MacheteAnnotator.java
```diff
@@ -107,6 +105,24 @@ private void processMacheteGeneratedEntry(MacheteGeneratedEntry macheteEntry, An
...
+                getNonHtmlString("string.GitMachete.MacheteAnnotator.cannot-find-local-branch-in-repo")
+                    .format(processedBranchName))
+            .range(branch);
+        if (parentBranchName.isEmpty()) { // do not suggest creating a new root branch
+          basicAnnotationBuilder.create();
+        } else { // suggest creating a new branch from the parent branch
+          basicAnnotationBuilder.withFix(new CreateBranchQuickFix(processedBranchName, parentBranchName, file)).create();
+        }
+      } catch (BranchLayoutException e) {
+        throw new RuntimeException(e);
```
PawelLipski, line 123: Huh probably throwing an exception here is not a good choice:

```
In file: file:///home/plipski/git-machete-intellij-plugin/.git/machete

java.lang.RuntimeException: com.virtuslab.branchlayout.api.BranchLayoutException: Line no 9 in branch layout file (/home/plipski/git-machete-intellij-plugin/.git/machete) has unexpected indentation character inconsistent with previous one. Problematic line number: 9
	at com.virtuslab.gitmachete.frontend.file.highlighting.MacheteAnnotator.processMacheteGeneratedEntry(MacheteAnnotator.java:123)
	at com.virtuslab.gitmachete.frontend.file.highlighting.MacheteAnnotator.annotate(MacheteAnnotator.java:55)
	at com.intellij.codeInsight.daemon.impl.DefaultHighlightVisitor.runAnnotators(DefaultHighlightVisitor.java:131)
	at com.intellij.codeInsight.daemon.impl.DefaultHighlightVisitor.visit(DefaultHighlightVisitor.java:108)
	at com.intellij.codeInsight.daemon.impl.GeneralHighlightingPass.runVisitors(GeneralHighlightingPass.java:347)
	at com.intellij.codeInsight.daemon.impl.GeneralHighlightingPass.lambda$collectHighlights$7(GeneralHighlightingPass.java:279)
	at com.intellij.codeInsight.daemon.impl.GeneralHighlightingPass.analyzeByVisitors(GeneralHighlightingPass.java:306)
	at com.intellij.codeInsight.daemon.impl.GeneralHighlightingPass.lambda$analyzeByVisitors$8(GeneralHighlightingPass.java:309)
	at com.intellij.codeInsight.daemon.impl.DefaultHighlightVisitor.analyze(DefaultHighlightVisitor.java:93)
	at com.intellij.codeInsight.daemon.impl.GeneralHighlightingPass.analyzeByVisitors(GeneralHighlightingPass.java:309)
	at com.intellij.codeInsight.daemon.impl.GeneralHighlightingPass.collectHighlights(GeneralHighlightingPass.java:276)
	at com.intellij.codeInsight.daemon.impl.GeneralHighlightingPass.collectInformationWithProgress(GeneralHighlightingPass.java:221)
	at com.intellij.codeInsight.daemon.impl.ProgressableTextEditorHighlightingPass.doCollectInformation(ProgressableTextEditorHighlightingPass.java:97)
	at com.intellij.codeHighlighting.TextEditorHighlightingPass.collectInformation(TextEditorHighlightingPass.java:57)
	at com.intellij.codeInsight.daemon.impl.PassExecutorService$ScheduledPass.lambda$doRun$1(PassExecutorService.java:400)
	at com.intellij.openapi.application.impl.ApplicationImpl.tryRunReadAction(ApplicationImpl.java:1154)
	at com.intellij.codeInsight.daemon.impl.PassExecutorService$ScheduledPass.lambda$doRun$2(PassExecutorService.java:393)
	at com.intellij.openapi.progress.impl.CoreProgressManager.lambda$executeProcessUnderProgress$12(CoreProgressManager.java:608)
	at com.intellij.openapi.progress.impl.CoreProgressManager.registerIndicatorAndRun(CoreProgressManager.java:683)
	at com.intellij.openapi.progress.impl.CoreProgressManager.computeUnderProgress(CoreProgressManager.java:639)
	at com.intellij.openapi.progress.impl.CoreProgressManager.executeProcessUnderProgress(CoreProgressManager.java:607)
	at com.intellij.openapi.progress.impl.ProgressManagerImpl.executeProcessUnderProgress(ProgressManagerImpl.java:60)
	at com.intellij.codeInsight.daemon.impl.PassExecutorService$ScheduledPass.doRun(PassExecutorService.java:392)
	at com.intellij.codeInsight.daemon.impl.PassExecutorService$ScheduledPass.lambda$run$0(PassExecutorService.java:368)
	at com.intellij.openapi.application.impl.ReadMostlyRWLock.executeByImpatientReader(ReadMostlyRWLock.java:174)
	at com.intellij.openapi.application.impl.ApplicationImpl.executeByImpatientReader(ApplicationImpl.java:215)
	at com.intellij.codeInsight.daemon.impl.PassExecutorService$ScheduledPass.run(PassExecutorService.java:366)
	at com.intellij.concurrency.JobLauncherImpl$VoidForkJoinTask$1.exec(JobLauncherImpl.java:184)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:373)
	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1182)
	at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1655)
	at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1622)
	at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:165)
Caused by: com.virtuslab.branchlayout.api.BranchLayoutException: Line no 9 in branch layout file (/home/plipski/git-machete-intellij-plugin/.git/machete) has unexpected indentation character inconsistent with previous one. Problematic line number: 9
	at com.virtuslab.branchlayout.impl.readwrite.BranchLayoutFileReader.parseToArrayRepresentation(BranchLayoutFileReader.java:7)
	at com.virtuslab.branchlayout.impl.readwrite.BranchLayoutFileReader.read(BranchLayoutFileReader.java:44)
	at com.virtuslab.gitmachete.frontend.file.highlighting.MacheteAnnotator.getParentBranchName(MacheteAnnotator.java:145)
	at com.virtuslab.gitmachete.frontend.file.highlighting.MacheteAnnotator.processMacheteGeneratedEntry(MacheteAnnotator.java:111)
	... 32 more
```

IDE error just because I've got some mess in machete file for a moment... we're gonna get flooded with IDE errors ☹️

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/quickfix/CreateBranchQuickFix.java
LukaszKontowski: Yes. From `ivoke` javadoc:
> Called when user invokes intention. This method is called inside command. **If startInWriteAction() returns true**, this method is also called inside write action

And in `CreateBranchQuickFix` code we have:
```
public boolean startInWriteAction() {
    return false;
  }
```

LukaszKontowski: So, it is **not** called inside a write action. And write actions are run on Swing's EDT if I get it rigth.

PawelLipski: > If startInWriteAction() returns true, this method is also called inside write action

The implication is only in one direction: `startInWriteAction()` => call inside write action => we're on EDT

But it doesn't mean that the reverse is true ;) `startInWriteAction()` might be false and yet the code might execute on EDT (even if not within a write action) 🤔

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/quickfix/CreateBranchQuickFix.java
LukaszKontowski: I tried to add a solution to trigger a reparsing but failed for now. Maybe it would be possible, but:
- either the solution did not work at all
- or solution did work, but caused multiple recursive calls of `MacheteAnnotator#annotate` even if nothing changed
I think I will add an issue to fix that. It is probably posssible, but somehow tricky and needs some work.

PawelLipski: Okay, pls open an issue then

LukaszKontowski: issue created: https://github.com/VirtusLab/git-machete-intellij-plugin/issues/1281

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/highlighting/MacheteAnnotator.java
LukaszKontowski: Same workaround for `BranchLayoutException` in the same method...

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/quickfix/CreateBranchQuickFix.java
LukaszKontowski: Ok, you are right - I checked that and in fact `SwingUtilities.isEventDispatchThread()` returns `true` for this method. Let me remove this `@UIThreadUnsafe` annotation.

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/MacheteFileUtils.java
```diff
@@ -30,19 +36,30 @@ public static String getSampleMacheteFileContents() {
...
     }
 
-    return List.ofAll(gitRepository.get().getInfo().getLocalBranchesWithHashes().keySet())
+    return List.ofAll(gitRepository.getInfo().getLocalBranchesWithHashes().keySet())
         .map(localBranch -> localBranch.getName());
   }
+
+  @Nullable
+  @UIThreadUnsafe
+  public static GitRepository findGitRepositoryForPsiMacheteFile(PsiFile psiFile) {
```
PawelLipski, line 51: Nit:
```suggestion
  @UIThreadUnsafe
  public static @Nullable GitRepository findGitRepositoryForPsiMacheteFile(PsiFile psiFile) {
```
As `@Nullable` is about the type, not the method

frontend/resourcebundles/src/main/resources/GitMacheteBundle.properties
```diff
@@ -317,6 +317,7 @@ string.GitMachete.RediscoverSuggester.backgroundable-check-task.title=Checking f
 
 # Machete file editor
 
+action.GitMachete.MacheteAnnotator.IntentionAction.create-nonexistent-branch=Create branch {0} from {1}
```
PawelLipski, line 320:
```suggestion
action.GitMachete.MacheteAnnotator.IntentionAction.create-nonexistent-branch=Create branch ''{0}'' from ''{1}''
```

AFAICS intentions always quote branch names etc.

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/quickfix/CreateBranchQuickFix.java
LukaszKontowski: I removed these annotations, which results in following error from our checks:
```
java.lang.AssertionError: Architecture Violation [Priority: MEDIUM] - Rule 'methods that are not annotated with @UIThreadUnsafe and are not annotated with @IgnoreUIThreadUnsafeCalls should never call any @UIThreadUnsafe methods' was violated (1 times):
a non-@UIThreadUnsafe method com.virtuslab.gitmachete.frontend.file.quickfix.CreateBranchQuickFix.createNewBranchFromParent(com.intellij.openapi.project.Project) calls a @UIThreadUnsafe method com.virtuslab.gitmachete.frontend.file.MacheteFileUtils.findGitRepositoryForPsiMacheteFile(com.intellij.psi.PsiFile)
```
@PawelLipski Can we do something about it? Maybe disable this _violation_ for this one file?

LukaszKontowski: As `createNewBranchFromParent` calls `@UIThreadUnsafe` `MacheteFileUtils.findGitRepositoryForPsiMacheteFile`

LukaszKontowski: Ok, I just realised that I can fix this without such dirty tricks. Set `GitRepository gitRepository` in MacheteAnnotator and use it as another field in the `CreateBranchQuickFix` class

PawelLipski: Okay :) in general there's `@IgnoreUIThreadUnsafeCalls` for allowing UI-thread-unsafe calls from UI thread... but it's to be used sparingly (we've got it in like 3 places only so far).

LukaszKontowski: Yeah, I used this one

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/quickfix/CreateBranchQuickFix.java
```diff
@@ -0,0 +1,66 @@
...
+
+  private final String branch;
+  private final String parentBranch;
+  private final PsiFile macheteFile;
+  private final @Nullable GitRepository gitRepository;
+
+  @Override
+  public @IntentionName String getText() {
+    return getNonHtmlString("action.GitMachete.MacheteAnnotator.IntentionAction.create-nonexistent-branch")
+        .format(branch, parentBranch);
```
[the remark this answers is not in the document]
LukaszKontowski: Works properly - both branches from snippet created properly. I was able to checkout the `<html>test</html>` branch locally.
image

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/quickfix/CreateBranchQuickFix.java
```diff
@@ -0,0 +1,66 @@
...
+    return "Git Machete";
+  }
+
+  @Override
+  public boolean isAvailable(Project project, Editor editor, PsiFile file) {
+    return true;
+  }
+
+  @Override
+  public void invoke(Project project, Editor editor, PsiFile file) {
```
PawelLipski, line 48: Okay... this one can be marked as `@UIEffect` right? as you've checked experimentally that it's executed on UI thread right?

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/quickfix/CreateBranchQuickFix.java
```diff
@@ -0,0 +1,66 @@
...
+
+  @Override
+  public boolean startInWriteAction() {
+    return false;
+  }
+
+  @IgnoreUIThreadUnsafeCalls
+  private void createNewBranchFromParent(Project project) {
+    if (gitRepository != null) {
+      GitBrancher.getInstance(project).createBranch(branch, Collections.singletonMap(gitRepository, parentBranch));
```
PawelLipski, line 60: Actually, AFAICS `createBranch` here ends up calling `git4idea.branch.GitBrancherImpl#createBranch(java.lang.String, java.util.Map<git4idea.repo.GitRepository,java.lang.String>, boolean)`, which itself just schedules a background task.

So it can be considered safe for running on UI thread. Pls remove `@IgnoreUIThreadUnsafeCalls` and instead patch `com.virtuslab.archunit.UIThreadUnsafeMethodInvocationsTestSuite#whitelistedMethodFullNames_git4idea` ✍🏻

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/quickfix/CreateBranchQuickFix.java
LukaszKontowski: It is executed on the Swing's event dispatch thread. Let me add this and. we'll see if that is OK.

PawelLipski (approved):
Other than that, approving 🎉 but #1238 must be shipped as well 😅

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/quickfix/CreateBranchQuickFix.java
LukaszKontowski: Ok, I can't add this annotation, since:
> /root/git-machete-intellij-plugin/frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/quickfix/CreateBranchQuickFix.java:49: error: [[guieffect, allcheckers]:override.effect] A method override may only be @UI if it overrides an @UI method.

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/quickfix/CreateBranchQuickFix.java
LukaszKontowski: Huh, I tried to do that, but looks like my solution does not work.

src/test/java/com/virtuslab/archunit/UIThreadUnsafeMethodInvocationsTestSuite.java
```diff
@@ -97,6 +97,7 @@ public void check(JavaMethod method, ConditionEvents events) {
       "git4idea.GitUtil.getRepositories(com.intellij.openapi.project.Project)",
       "git4idea.GitUtil.getRepositoryManager(com.intellij.openapi.project.Project)",
       "git4idea.branch.GitBranchUtil.sortBranchNames(java.util.Collection)",
+      "git4idea.branch.GitBrancherImpl.createBranch(java.lang.String, java.util.Map<git4idea.repo.GitRepository,java.lang.String>, boolean)",
```
PawelLipski: Nope! you don't call this method anywhere directly, you call 

```suggestion
      "git4idea.branch.GitBrancher#createBranch(java.lang.String, java.util.Map<git4idea.repo.GitRepository,java.lang.String>)",
```

This is a whitelist of methods from git4idea that could be called even on UI thread

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/quickfix/CreateBranchQuickFix.java
PawelLipski: See below 👇🏻

src/test/java/com/virtuslab/archunit/UIThreadUnsafeMethodInvocationsTestSuite.java
LukaszKontowski: Ok, I managed to add these method names properly

LukaszKontowski: In fact, it looks like we should use `.` instead of `#` in this place of code

PR #1264, opened 2022-10-12
Expose `getParent` from `BranchLayoutEntry`

PawelLipski:


branchLayout/api/src/main/java/com/virtuslab/branchlayout/api/BranchLayoutEntry.java
```diff
@@ -16,27 +16,43 @@
...
 @ToString
-public class BranchLayoutEntry {
+public final class BranchLayoutEntry {
   @Getter
   private final String name;
 
   @Getter
   private final @Nullable String customAnnotation;
 
+  @Getter
```
PawelLipski, line 29: To be used on PR #1243

branchLayout/api/src/main/java/com/virtuslab/branchlayout/api/BranchLayoutEntry.java
```diff
@@ -16,27 +16,43 @@
...
+    this.name = name;
+    this.customAnnotation = customAnnotation;
+    this.children = children;
+
+    // Note: since the class is final, `this` is already @Initialized at this point.
+    // We're thus safe to store `this` in other objects' fields.
+
+    for (val child : children) {
+      child.parent = this;
+    }
```
LukaszKontowski, line 46: Are you sure this logic works as expected? I would suggest adding some test / tests to make sure that `parent` is set properly.

branchLayout/api/src/main/java/com/virtuslab/branchlayout/api/BranchLayoutEntry.java
```diff
@@ -16,27 +16,43 @@
...
-public class BranchLayoutEntry {
+public final class BranchLayoutEntry {
   @Getter
   private final String name;
 
   @Getter
   private final @Nullable String customAnnotation;
 
+  @Getter
+  private @Nullable BranchLayoutEntry parent = null;
```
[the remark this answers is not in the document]
PawelLipski: _Parent_ is rather universally understood to mean exactly **one level up** in trees (and even in DAGs, like with git commit structure: https://slides.com/plipski/git-internals#/7, even though a commit might have multiple parents). So `branch_C.getParent()` will be `branch_B`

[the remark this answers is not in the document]
PawelLipski: `BranchLayout`s are **trees**, so there is no ambiguity as would be in DAGs.

Rel: https://stackoverflow.com/a/43046393/2750812

LukaszKontowski:
We can additionally provide a convenient method in the `com.virtuslab.branchlayout.api.BranchLayout` - that would just call `BranchLayoutEntry#getParent`. Something like:
`public @Nullable BranchLayoutEntry findParent(String branchName)`
It would make sense, as similar user-facing methods are already declared in this class (easy to find for developers).

branchLayout/api/src/main/java/com/virtuslab/branchlayout/api/BranchLayoutEntry.java
```diff
@@ -16,27 +16,43 @@
...
+  public BranchLayoutEntry(String name, @Nullable String customAnnotation, List<BranchLayoutEntry> children) {
+    this.name = name;
+    this.customAnnotation = customAnnotation;
+    this.children = children;
+
+    // Note: since the class is final, `this` is already @Initialized at this point.
+    // We're thus safe to store `this` in other objects' fields.
+
+    for (val child : children) {
+      child.parent = this;
```
mkondratek, line 45: and it is possible to set the parent even though the child class is final? 

image

branchLayout/api/src/main/java/com/virtuslab/branchlayout/api/BranchLayoutEntry.java
PawelLipski: Whoops of course it didn't, as usually with untested code & mutable structures, there was a subtle bug around slide-outs :trollface: 

I added tests + changed the structure to completely immutable. Children are copied in c'tor to new objects with a correct parent.

branchLayout/api/src/main/java/com/virtuslab/branchlayout/api/BranchLayoutEntry.java
PawelLipski: Finality of the class shouldn't matter here ;) but generally I've changed this entire structure to immutable now

PawelLipski:
> We can additionally provide a convenient method in the `com.virtuslab.branchlayout.api.BranchLayout` - that would just call `BranchLayoutEntry#getParent`. Something like: `public @Nullable BranchLayoutEntry findParent(String branchName)` It would make sense, as similar user-facing methods are already declared in this class (easy to find for developers).

That would be just `layout.getEntryByName(name).getParent()` right? 🤔 hmm one problem I can see with this method is that `null` result is ambiguous: it could mean either that entry with `name` has not been found, or that `parent` of entry is null 🤔 I'd rather leave handling of that up to callers... otherwise I've got a feeling this might introduce subtle bugs 🤔

branchLayout/impl/src/test/java/com/virtuslab/branchlayout/unit/BranchLayoutTestSuite.java
```diff
@@ -121,6 +86,50 @@ public void givenOnlyRootsEntryLayout_nextAndPreviousShouldBeCalculated() {
     assertEquals(branchLayout.findPreviousEntry(rootName2).getName(), rootName1);
   }
 
+  @Test
+  public void withBranchSlideOut_givenNonRootExistingBranch_slidesOut() {
```
PawelLipski, line 90: Moved closer to other tests for `slideOut`

LukaszKontowski:
Pease, add a test for `BranchLayoutFileReader#read` - generated `BranchLayout`. To test created layout if parents are set properly.

LukaszKontowski (approved):
Looks good

PR #1296, opened 2022-10-22
Fix `TestProcessUtils` to make `gradle build` pass

PawelLipski:


.circleci/config.yml
```diff
@@ -89,7 +89,7 @@ jobs:
           command: ./scripts/verify-no-string-interpolation-leftovers
       - run:
           name: Run static code analyzer
-          command: ./gradlew --warn checkstyleMain
+          command: ./gradlew --warn checkstyleMain checkstyleTestFixtures
```
mkondratek, line 92: what is this option about?

mkondratek (approved):

.circleci/config.yml
PawelLipski: To run checkstyle on `testCommon/src/testFixtures/`! https://docs.gradle.org/current/userguide/java_testing.html#sec:java_test_fixtures

Issue #1227, opened 2022-10-05
What operations should be wrapped into `ReadAction` and `WriteAction`?


PawelLipski:
There's currently just a single use of `com.intellij.openapi.application.WriteAction` across all codebase, in `com.virtuslab.gitmachete.frontend.actions.toolbar.OpenMacheteFileAction`:

```java
    val macheteFile = WriteAction.compute(() -> Try
        .of(() -> GitVfsUtils.getMacheteFile(repo))
        .onFailure(e -> VcsNotifier.getInstance(project).notifyWeakError(/* displayId */ null,
            /* title */ "",
            /* message */ getString("action.GitMachete.OpenMacheteFileAction.notification.title.cannot-open"))))
        .getOrNull();
```

Let's determine where exactly `ReadAction` and `WriteAction` should be used, in particular if the existing invocation is correct.

It might be hard to deduce just from JetBrains docs... probably it might be needed to check the code of intellij-communty/official plugins.

See https://plugins.jetbrains.com/docs/intellij/general-threading-rules.html -> _Writing data is only allowed from the UI thread_ wuuuut 😯 I suppose it refers to PSI/VFS, and not to writing files directly to disk (as we do with .git/machete) 😅 It means that `WriteAction.compute` is effectively a `@UIEffect` method in Checker terms, sth to reflect in Checker stubs.

Whatever the conclusions, let's _immortalize_ them via ArchUnit tests or sth wherever applicable.

PawelLipski:
Load balancing, as @mkondratek is busy with features

PawelLipski:
Okay @MaciejG604 ... anything left to be done here? 🤔

PR #1301, opened 2022-10-25
Wrap machete file IO operations in ReadAction/WriteAction

MaciejG604:
uiTests with headless flag don't work on Mac, @PawelLipski could you test them on Linux?


frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/backgroundables/BaseSlideInBackgroundable.java
```diff
@@ -86,11 +88,14 @@ public void run(ProgressIndicator indicator) {
       return;
     }
 
-    Try.run(() -> branchLayoutWriter.write(macheteFilePath, newBranchLayout, /* backupOldLayout */ true))
-        .onFailure(t -> VcsNotifier.getInstance(project).notifyError(/* displayId */ null,
-            /* title */ getString(
-                "action.GitMachete.BaseSlideInBelowAction.notification.title.branch-layout-write-fail"),
-            getMessageOrEmpty(t)));
+    ApplicationManager.getApplication()
+        .invokeLater(() -> WriteAction
```
mkondratek, line 92: is it needed to run `compute` within `invokeLater`?

MaciejG604: According to this:
image
And this [[source]](https://plugins.jetbrains.com/docs/intellij/general-threading-rules.html#modality-and-invokelater):
image

I think it should be inside `invokeLater` when we're using a background thread

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/toolbar/DiscoverAction.java
```diff
@@ -127,7 +130,17 @@ private void saveDiscoveredLayout(IGitMacheteRepositorySnapshot repositorySnapsh
...
-        branchLayoutWriter.write(macheteFilePath, branchLayout, /* backupOldLayout */ true);
+        ApplicationManager.getApplication()
+            .invokeLater(() -> WriteAction
+                .compute(() -> {
+                  try {
+                    branchLayoutWriter.write(macheteFilePath, branchLayout, /* backupOldLayout */ true);
+                  } catch (BranchLayoutException e) {
+                    throw new RuntimeException(e);
+                  }
+                  return true;
```
mkondratek, line 141: return true just to return smth, right? if so, we probably could use 
`com.intellij.openapi.application.WriteAction#run(com.intellij.util.ThrowableRunnable<E>)` here

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/toolbar/DiscoverAction.java
```diff
@@ -142,6 +155,7 @@ public void onSuccess() {
       @Override
       @UIEffect
       public void onThrowable(Throwable e) {
+        val branchLayoutException = e.getCause();
```
mkondratek, line 158: why is it needed now?

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseSlideOutAction.java
```diff
@@ -171,20 +173,23 @@ private void slideOutBranch(AnActionEvent anActionEvent, String branchName) {
...
-      val errorMessage = "Error occurred while sliding out '${branchName}' branch" +
-          (exceptionMessage == null ? "" : ": " + exceptionMessage);
-      LOG.error(errorMessage);
-      VcsNotifier.getInstance(project).notifyError(/* displayId */ null,
-          getString("action.GitMachete.BaseSlideOutAction.notification.title.slide-out-fail.HTML").format(branchName),
-          exceptionMessage == null ? "" : exceptionMessage);
-    }
+    ApplicationManager.getApplication()
+        .invokeLater(() -> WriteAction
+            .compute(() -> Try.run(() -> {
```
mkondratek, line 178: `com.intellij.openapi.application.WriteAction#run(com.intellij.util.ThrowableRunnable<E>)` could be used here too, right?

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/backgroundables/BaseSlideInBackgroundable.java
```diff
@@ -86,11 +88,14 @@ public void run(ProgressIndicator indicator) {
...
+            .compute(() -> Try.run(() -> branchLayoutWriter.write(macheteFilePath, newBranchLayout, /* backupOldLayout */ true))
```
mkondratek, line 93: and here - compute returns, run runs

frontend/ui/impl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/GitMacheteRepositoryDiscoverer.java
```diff
@@ -78,19 +79,22 @@ public void run(ProgressIndicator indicator) {
...
-          ModalityUiUtil.invokeLaterIfNeeded(NON_MODAL, () -> VcsNotifier.getInstance(project)
-              .notifyError(
-                  /* displayId */ null,
-                  getString(
-                      "string.GitMachete.EnhancedGraphTable.automatic-discover.notification.title.cannot-discover-layout-error"),
-                  exception.getMessage() != null ? exception.getMessage() : ""));
-        }
+        ApplicationManager.getApplication()
+            .invokeLater(() -> WriteAction
+                .compute(() -> Try.run(() -> {
```
mkondratek, line 84: com.intellij.openapi.application.WriteAction#run(com.intellij.util.ThrowableRunnable<E>)

mkondratek (changes requested):

PawelLipski:
> uiTests with headless flag don't work on Mac, @PawelLipski could you test them on Linux?

Passing CI is basically a passing test on Linux (or in Linux-like Docker environment, to be precise) ;)

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/backgroundables/BaseSlideInBackgroundable.java
PawelLipski: I'm not sure if we should just use plain `write` (which uses some plain `java.(n)io` under the hood)... this should probably happen via VFS. Same with reading the branch layout file. Lemme open an issue...

PawelLipski: This requirement to write files on UI thread is weird as hell, completely against how I'd intuitively see that 🤔

PawelLipski: https://github.com/VirtusLab/git-machete-intellij-plugin/issues/1227

In fact, while the write is realized by `java.nio.file.Files#write` (in `com.virtuslab.branchlayout.impl.readwrite.BranchLayoutFileWriter#write`), there's probably no need for `WriteAction` here; pls put a TODO for #1227 instead

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseSlideOutAction.java
```diff
@@ -171,20 +173,23 @@ private void slideOutBranch(AnActionEvent anActionEvent, String branchName) {
...
+              val macheteFilePath = gitRepository.getMacheteFilePath();
+              LOG.info("Writing new branch layout into ${macheteFilePath}");
+              branchLayoutWriter.write(macheteFilePath, newBranchLayout, /* backupOldLayout */ true);
```
PawelLipski, line 181: Same here, TBD if `WriteAction` is needed currently. Albeit it will if #1303 ends up in a switch to a VFS write... so maybe better retain the `WriteAction`? or even better! @MaciejG604 pls first research #1303 and only then return to this PR 🔙

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/backgroundables/BaseSlideInBackgroundable.java
PawelLipski: Huh, see my comments below; IMO #1303 needs to be done first before this PR is further progressed

MaciejG604:
> > uiTests with headless flag don't work on Mac, @PawelLipski could you test them on Linux?
> 
> Passing CI is basically a passing test on Linux (or in Linux-like Docker environment, to be precise) ;)

But CI has a virtual display instead of running the test in headless mode, right? And headless mode is what might not work as I've found [[here]](https://developerlife.com/2021/03/13/ij-idea-plugin-advanced/#side-effect---invokelater-vs-submittransaction-and-impacts-on-test-code).

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/toolbar/DiscoverAction.java
MaciejG604: The compiler complains that code inside WriteAction.run() throws an uncaught exception, so the solution for that is a try-catch with rethrowing it as RuntimeException.

PawelLipski:
> But CI has a virtual display instead of running the test in headless mode, right? And headless mode is what might not work as I've found [[here]](https://developerlife.com/2021/03/13/ij-idea-plugin-advanced/#side-effect---invokelater-vs-submittransaction-and-impacts-on-test-code).

Huh we've never even meant to run UI tests in the actual headless mode (with `-Djava.awt.headless=true`). I don't think these tests were ever even close to working in such setup, and there isn't much value in that (as users, well, aren't likely to run IDE headlessly)

PawelLipski:
Btw... move the PR to draft? until #1303 is addressed

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/MacheteFileWriter.java
```diff
@@ -0,0 +1,64 @@
...
+import com.virtuslab.branchlayout.api.readwrite.IBranchLayoutWriter;
+import com.virtuslab.branchlayout.impl.readwrite.BranchLayoutFileUtils;
+import com.virtuslab.qual.guieffect.UIThreadUnsafe;
+
+@CustomLog
+public class MacheteFileWriter {
+
+  @UIThreadUnsafe
+  @SuppressWarnings("regexp") // to allow for `synchronized`
+  public synchronized void writeBranchLayout(Path path, BranchLayout branchLayout, boolean backupOldFile,
```
mkondratek, line 27: it can be static, right?

mkondratek, line 27: even we use WriteAction we still need `synchronized`? 😮

branchLayout/impl/src/main/java/com/virtuslab/branchlayout/impl/readwrite/BranchLayoutReader.java
```diff
@@ -171,7 +172,7 @@ private Array<Tuple2<Integer, Integer>> parseToArrayRepresentation(Path path, In
 
     if (indent % indentSpec.getIndentWidth() != 0) {
       throw new BranchLayoutException(lineNumber + 1,
-          "Levels of indentation are not matching in branch layout file (${path.toAbsolutePath()}): " +
+          "Levels of indentation are not matching in branch layout file: " +
```
mkondratek, line 175: I am wondering if it is worth to provide the path here 🤔 a specific path could be useful, wdyt?

mkondratek: probably we can still determine the specific path based on the other logs, right?

mkondratek (changes requested):
Not a very careful review for now as it's still a draft. 

Answering your question about MacheteFileReader/Writer location - your idea seems to be reasonable; the classes reference the file (the machete file) so placing them in `frontend:file` is ok, imho

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/MacheteFileWriter.java
MaciejG604: I wanted to make it static, but then there's this argument in some methods connected to the VFS:
image
And here:
image

So we need some kind of object reference.

Looks like this requestor information could then be used when reacting to VirtualFile changes using some listener e.g. like in EnhancedGraphTable. As we don't use it I guess it could be null. What do you think?

backend/impl/build.gradle.kts
```diff
@@ -7,6 +7,7 @@ dependencies {
   api(project(":backend:api"))
   api(project(":branchLayout:api"))
 
+  testImplementation(project(":frontend:file"))
```
PawelLipski, line 10: Hmmm backend depending on frontend (even if for the test)... looks questionable... lemme see how that dependency is used, however...

PawelLipski: Huh is that dependency really needed tho? 🤔

frontend/file/build.gradle.kts
```diff
@@ -17,12 +17,14 @@ apply<GrammarKitPlugin>()
 dependencies {
   api(project(":qual"))
   implementation(project(":binding"))
+  implementation(project(":branchLayout:impl"))
```
PawelLipski, line 20: Don't depend on implementations in general, depend just on API.

See javadoc for `com.virtuslab.binding.RuntimeBinding#instantiateSoleImplementingClass` for the rationale

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/MacheteFileWriter.java
PawelLipski: Yeah, let's reconsider why this `synchronized` is needed here, it's definitely a smell (and hence forbidden via checkstyle) 😅

backend/impl/build.gradle.kts
PawelLipski: Yeah, project builds without that, assuming it's some artifact from draft-ness of the PR

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/MacheteFileWriter.java
```diff
@@ -0,0 +1,64 @@
...
+        val indentSpec = BranchLayoutFileUtils.deriveIndentSpec(
+            List.ofAll(VfsUtilCore.loadText(macheteVFile).lines()));
+
+        try (val outputStream = new BufferedOutputStream(macheteVFile.getOutputStream(/* requestor */ this))) {
+          branchLayoutWriter.write(outputStream, branchLayout, indentSpec);
+        }
+      }
+
+    } catch (IOException e) {
+      throw new BranchLayoutException("Error while writing to (${macheteVFile.toNioPath()})", e);
```
PawelLipski, line 61: Hmm I wouldn't throw `BranchLayoutException` at this level... in fact, just letting `IOException` be thrown up the stack doesn't look like a bad idea

PawelLipski: This would also save you a `Try.of` above

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/MacheteFileWriter.java
```diff
@@ -0,0 +1,64 @@
...
+import com.virtuslab.branchlayout.api.BranchLayout;
+import com.virtuslab.branchlayout.api.BranchLayoutException;
+import com.virtuslab.branchlayout.api.readwrite.IBranchLayoutWriter;
+import com.virtuslab.branchlayout.impl.readwrite.BranchLayoutFileUtils;
+import com.virtuslab.qual.guieffect.UIThreadUnsafe;
+
+@CustomLog
+public class MacheteFileWriter {
+
+  @UIThreadUnsafe
```
PawelLipski, line 25: Given that this function is only supposed to be called within `WriteAction`, which are in turn executed on UI thread... shouldn't it rather be `@UIEffect`? and then `synchronized` won't be needed, as with all code that is only ever executed on UI thread (no risk of race conditions... that's why UI thread is designated in most UI frameworks in the first place)

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/MacheteFileWriter.java
```diff
@@ -0,0 +1,64 @@
...
+    val macheteVFile = VirtualFileManager.getInstance().findFileByNioPath(path);
+
+    try {
+      if (macheteVFile != null) {
+        if (backupOldFile) {
+          val parentDir = macheteVFile.getParent();
+          assert parentDir != null : "Can't get parent directory of branch layout file";
+
+          val backupFileName = macheteVFile.getName() + "~";
+          val backupVFile = ReadAction.compute(() -> Option.of(parentDir.findChild(backupFileName)));
```
PawelLipski, line 40: Same, aren't we supposed to be within `WriteAction` here? hence `ReadAction.compute` is probably redundant (if not erroneous)

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/MacheteFileReader.java
```diff
@@ -0,0 +1,35 @@
...
+    val macheteVFile = VirtualFileManager.getInstance().findFileByNioPath(path);
+
+    BranchLayout resultBranchLayout = new BranchLayout(List.empty());
+
+    if (macheteVFile != null) {
+      try (val inputStream = macheteVFile.getInputStream()) {
+
+        resultBranchLayout = branchLayoutReader.read(inputStream);
+      } catch (IOException e) {
+        throw new BranchLayoutException("Error while reading (${path})", e);
```
PawelLipski, line 30: Same as in Writer, maybe redundant

MaciejG604:
Stumbled upon an interesting thing in this [file](https://github.com/JetBrains/intellij-community/blob/master/platform/core-api/src/com/intellij/openapi/application/Application.java). I think this makes the usage a little bit easier and **explicitly** states that there are locks imposed on Read and Write actions as well as that they can be acquired from any thread.

branchLayout/api/src/main/java/com/virtuslab/branchlayout/api/readwrite/IBranchLayoutWriter.java
```diff
@@ -1,12 +1,14 @@
...
+import java.io.OutputStream;
+
+import io.vavr.collection.List;
 
 import com.virtuslab.branchlayout.api.BranchLayout;
-import com.virtuslab.branchlayout.api.BranchLayoutException;
 import com.virtuslab.qual.guieffect.UIThreadUnsafe;
 
 public interface IBranchLayoutWriter {
   @UIThreadUnsafe
```
PawelLipski, line 12: Actually, let's ditch that `@UIThreadUnsafe`, now that we write to VFS on UI thread anyway :trollface:

branchLayout/impl/src/main/java/com/virtuslab/branchlayout/impl/readwrite/BranchLayoutWriter.java
```diff
@@ -1,52 +1,40 @@
...
-
-    LOG.debug(() -> "Writing branch layout to ${path} with indent character ASCII " +
+    LOG.debug(() -> "Writing branch layout with indent character ASCII " +
         "code = ${indentSpec.getIndentCharacter()} and indent width = ${indentSpec.getIndentWidth()}");
     lines.forEach(LOG::debug);
 
-    Try.of(() -> Files.write(path, lines))
-        .getOrElseThrow(e -> new BranchLayoutException("Unable to write new branch layout file to ${path}", e));
+    // loops because of this part vvvv
+    for (String line : lines) {
```
PawelLipski, line 34: Hmm can't it just be sth like `outputStream.write(lines.mkString(System.lineSeparator()).getBytes())`, without a loop? or whatnot if `mkString` isn't available

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseSlideOutAction.java
```diff
@@ -171,20 +174,23 @@ private void slideOutBranch(AnActionEvent anActionEvent, String branchName) {
...
-          exceptionMessage == null ? "" : exceptionMessage);
-    }
+    ApplicationManager.getApplication()
+        .invokeLater(() -> WriteAction
+            .run(() -> Try.run(() -> {
+              val macheteFilePath = gitRepository.getMacheteFilePath();
+              LOG.info("Writing new branch layout into ${macheteFilePath}");
+              new MacheteFileWriter().writeBranchLayout(macheteFilePath, newBranchLayout, /* backupOldLayout */ true,
+                  branchLayoutWriter);
+            }).onFailure(t -> {
```
PawelLipski, line 184: Hmm I'd say that `try...catch` is more readable than `Try.run(...).onFailure()`

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/toolbar/DiscoverAction.java
```diff
@@ -127,7 +131,18 @@ private void saveDiscoveredLayout(IGitMacheteRepositorySnapshot repositorySnapsh
...
-        branchLayoutWriter.write(macheteFilePath, branchLayout, /* backupOldLayout */ true);
+        ApplicationManager.getApplication()
+            .invokeLater(() -> {
+              try {
+                WriteAction
+                    .run(() -> new MacheteFileWriter().writeBranchLayout(macheteFilePath, branchLayout,
+                        /* backupOldLayout */ true,
+                        branchLayoutWriter));
+              } catch (BranchLayoutException e) {
+                throw new RuntimeException(e);
```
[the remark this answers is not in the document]
PawelLipski: Or just ignore `try...catch` completely, since you've got a `@SneakyThrows` already.

Rel: #1317

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/MacheteFileReader.java
```diff
@@ -0,0 +1,34 @@
...
+import lombok.val;
+
+import com.virtuslab.branchlayout.api.BranchLayout;
+import com.virtuslab.branchlayout.api.BranchLayoutException;
+import com.virtuslab.branchlayout.api.readwrite.IBranchLayoutReader;
+import com.virtuslab.qual.guieffect.UIThreadUnsafe;
+
+@CustomLog
+public class MacheteFileReader {
+  @UIThreadUnsafe
```
PawelLipski, line 18: TBH I'd say it's safe for UI thread. Since we are even expected to **write** to VFS on UI thread (via `WriteAction`s)... then reading is even safer. What are the UI-thread-unsafe calls reported by ArchUnit in this method?

frontend/ui/impl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/EnhancedGraphTable.java
```diff
@@ -182,7 +183,6 @@ public void after(java.util.List<? extends VFileEvent> events) {
               if (slideInNotification != null && !slideInNotification.isExpired()) {
                 slideInNotification.expire();
               }
-              queueRepositoryUpdateAndModelRefresh();
```
PawelLipski, line 185: Huh is that intentional? 🤔

PawelLipski: & related to this PR?

frontend/ui/impl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/EnhancedGraphTable.java
```diff
@@ -400,20 +400,22 @@ private void slideOutSkippedBranches(IGitMacheteRepositorySnapshot repositorySna
...
-      String errorMessage = "Error occurred while sliding out skipped branches" +
-          (exceptionMessage == null ? "" : ": " + exceptionMessage);
-      LOG.error(errorMessage);
-      VcsNotifier.getInstance(project).notifyError(/* displayId */ null,
-          getString("action.GitMachete.EnhancedGraphTable.branch-layout-write-failure"),
-          exceptionMessage == null ? "" : exceptionMessage);
-    }
+    val finalNewBranchLayout = newBranchLayout;
+    ApplicationManager.getApplication()
+        .invokeLater(() -> WriteAction.run(() -> Try.run(() -> {
```
PawelLipski, line 405: Yeah... `try...catch` is more readable

branchLayout/impl/src/main/java/com/virtuslab/branchlayout/impl/readwrite/BranchLayoutWriter.java
```diff
@@ -0,0 +1,55 @@
...
+import com.virtuslab.branchlayout.api.BranchLayout;
+import com.virtuslab.branchlayout.api.BranchLayoutEntry;
+import com.virtuslab.branchlayout.api.readwrite.IBranchLayoutWriter;
+import com.virtuslab.branchlayout.api.readwrite.IndentSpec;
+
+@CustomLog
+@RequiredArgsConstructor
+public class BranchLayoutWriter implements IBranchLayoutWriter {
+  @Override
+  public void write(OutputStream outputStream, BranchLayout branchLayout, List<String> oldFileLines)
```
PawelLipski, line 21: Hmmm `oldFileLines` are just needed to derive indent spec... maybe just pass an `IndentSpec` as an argument? 🤔

branchLayout/impl/src/main/java/com/virtuslab/branchlayout/impl/readwrite/BranchLayoutWriter.java
```diff
@@ -0,0 +1,55 @@
...
+    val lines = printEntriesOntoStringList(branchLayout.getRootEntries(), indentSpec, /* level */ 0);
+
+    LOG.debug(() -> "Writing branch layout with indent character ASCII " +
+        "code = ${indentSpec.getIndentCharacter()} and indent width = ${indentSpec.getIndentWidth()}");
+    lines.forEach(LOG::debug);
+
+    outputStream.write(lines.mkString(System.lineSeparator()).getBytes());
+  }
+
+  private List<String> printEntriesOntoStringList(
```
PawelLipski, line 34: I wouldn't use the verb `print` here 🤔 maybe `flatten`? 🤔

branchLayout/impl/src/main/java/com/virtuslab/branchlayout/impl/readwrite/BranchLayoutFileUtils.java
```diff
@@ -45,6 +47,16 @@ public static IndentSpec deriveIndentSpec(Path path) {
     return deriveIndentSpec(lines);
   }
 
+  @UIThreadUnsafe
+  public static IndentSpec deriveIndentSpec(InputStream inputStream) throws BranchLayoutException {
+    List<String> lines = Try.of(() -> List.ofAll(new String(inputStream.readAllBytes()).lines()))
```
PawelLipski, line 52: While you're here... `try ... catch`?

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/backgroundables/BaseSlideInBackgroundable.java
```diff
@@ -86,11 +89,18 @@ public void run(ProgressIndicator indicator) {
...
     }
 
-    Try.run(() -> branchLayoutWriter.write(macheteFilePath, newBranchLayout, /* backupOldLayout */ true))
-        .onFailure(t -> VcsNotifier.getInstance(project).notifyError(/* displayId */ null,
-            /* title */ getString(
-                "action.GitMachete.BaseSlideInBelowAction.notification.title.branch-layout-write-fail"),
-            getMessageOrEmpty(t)));
+    ApplicationManager.getApplication()
+        .invokeLater(() -> WriteAction
+            .run(() -> Try.run(() -> new MacheteFileWriter().writeBranchLayout(
```
PawelLipski, line 94: `try ... catch`?

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseSlideOutAction.java
```diff
@@ -171,20 +174,25 @@ private void slideOutBranch(AnActionEvent anActionEvent, String branchName) {
...
-      val errorMessage = "Error occurred while sliding out '${branchName}' branch" +
-          (exceptionMessage == null ? "" : ": " + exceptionMessage);
-      LOG.error(errorMessage);
-      VcsNotifier.getInstance(project).notifyError(/* displayId */ null,
-          getString("action.GitMachete.BaseSlideOutAction.notification.title.slide-out-fail.HTML").format(branchName),
-          exceptionMessage == null ? "" : exceptionMessage);
-    }
+    ApplicationManager.getApplication()
+        .invokeLater(() -> WriteAction
+            .run(() -> {
```
PawelLipski, line 179: Hmmm isn't there sth like `Application#runWriteAction` that would simplify that from 2 to 1 method call?

frontend/ui/impl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/GitMacheteRepositoryDiscoverer.java
```diff
@@ -96,20 +98,25 @@ public void run(ProgressIndicator indicator) {
...
-          ModalityUiUtil.invokeLaterIfNeeded(NON_MODAL, () -> VcsNotifier.getInstance(project)
-              .notifyError(
-                  /* displayId */ null,
-                  getString(
-                      "string.GitMachete.EnhancedGraphTable.automatic-discover.notification.title.cannot-discover-layout-error"),
-                  exception.getMessage() != null ? exception.getMessage() : ""));
-        }
+        ApplicationManager.getApplication()
+            .invokeLater(() -> WriteAction
+                .run(() -> Try.run(() -> {
```
PawelLipski, line 103: `try ... catch` while you're here

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/MacheteFileWriter.java
MaciejG604: Okay, both methods will be static, but they require one more argument for the sake of using this `requestor` mechanism correctly. WDYT?

PawelLipski: Hmmm
>  I guess it could be null. What do you think?

but
>  they will require one more argument for the sake of using this requestor mechanism correctly

So after all, the `requestor` canNOT be `null` in this case? 🤔

MaciejG604: Well it CAN be for sure, if we never plan on using any more sophisticated VFS changes listeners that's totally fine (if IntelliJ doesn't use this field in any other way but that's hard to tell).
I just thought that putting a new argument there won't be a problem but we use the `requestor` mechanism as intended.

PawelLipski: Okay, won't harm to pass `requestor` as argument probably

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/MacheteFileReader.java
MaciejG604: ArchUnit reports all the operations connected with streams (package `java.io`) as unsafe. I deleted the `UIThreadUnsafe` annotations and put `IgnoreUIThreadUnsafeCalls` where it was needed.

But maybe the ArchUnit test should be somehow refactored based on this new idea that those actions are not unsafe and even need to be executed on UIThread?

PawelLipski: > But maybe the ArchUnit test should be somehow refactored based on this new idea that those actions are not unsafe and even need to be executed on UIThread?

Hmmm theoretically it's just a matter of updating `com.virtuslab.archunit.UIThreadUnsafeMethodInvocationsTestSuite#whitelistedMethodFullNames_java_io`... could you list the specific method from `java.io` reported here? so that we can decide whether:
1. keep `@IgnoreUIThreadUnsafeCalls` here in `frontend:file` method
or rather
2. update `whitelistedMethodFullNames_java_io`

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseSlideOutAction.java
MaciejG604: "Must be called from the Swing dispatch thread" so won't suit us.

image

PawelLipski: Awww okay... maybe add a helper method of our own in e.g. `com.virtuslab.gitmachete.frontend.actions.common.ActionUtils`

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/toolbar/DiscoverAction.java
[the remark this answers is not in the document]
MaciejG604: The catch is required by the compiler since `invokeLater` does not throw any checked exception.

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/MacheteFileReader.java
MaciejG604: From `MacheteFileWriter`: `{"java.io.BufferedOutputStream.<init>(java.io.OutputStream)", "java.io.BufferedOutputStream.close()"}`

From `MacheteFileReader`: `"java.io.InputStream.close()"`

No stream constructors reported since we use `VirtualFile.getInputStream()` (not from java.io).

PawelLipski: Okay, I think all of `java.io.BufferedOutputStream.<init>(java.io.OutputStream)`, `java.io.BufferedOutputStream.close()` and `java.io.InputStream.close()` can be safely added to `whitelistedMethodFullNames_java_io`, thus removing the need for `@IgnoreUIThreadUnsafeCalls`

branchLayout/impl/src/main/java/com/virtuslab/branchlayout/impl/readwrite/BranchLayoutReader.java
PawelLipski: Let's leave as is, it shouldn't introduce much mess

config/checker/com.intellij.astub
```diff
@@ -360,8 +373,23 @@ package com.intellij.openapi.vfs;
 
 class VirtualFile {
   @Nullable VirtualFile getParent();
+
+  OutputStream getOutputStream(@Nullable Object requestor) throws IOException {}
+
+  void delete(@Nullable Object requestor) throws IOException {}
+
+  @NonNull VirtualFile createChildData(@Nullable Object requestor, String name) throws IOException {}
```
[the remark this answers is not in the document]
PawelLipski: Pls keep methods in abc order (with respect to their names)

frontend/file/src/main/java/com/virtuslab/gitmachete/frontend/file/MacheteFileWriter.java
```diff
@@ -0,0 +1,66 @@
...
+   * @param path a path to file where branch layout should be written
+   * @param branchLayout layout to be written
+   * @param backupOldFile flag stating if old layout file should be backed-up
+   * @param requestor object requesting write execution (can be used for debugging when listening for file changes)
+   */
+  @IgnoreUIThreadUnsafeCalls({"java.io.BufferedOutputStream.<init>(java.io.OutputStream)",
+      "java.io.BufferedOutputStream.close()"})
+  @UIEffect
+  public static void writeBranchLayout(Path path, BranchLayout branchLayout, boolean backupOldFile,
+      IBranchLayoutWriter branchLayoutWriter, @Nullable Object requestor)
```
PawelLipski, line 35: Hmmm I'd reorder the params a bit... having `branchLayoutWriter` after `branchLayout` feels odd

frontend/ui/impl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/EnhancedGraphTable.java
```diff
@@ -473,11 +417,87 @@ public void refreshModel() {
   }
 
   @UIEffect
-  private void initColumns() {
-    createDefaultColumnsFromModel();
+  private void refreshModel(
```
PawelLipski, line 420: Could you move the methods back to their original place? so that the actual changes are easier to see

frontend/ui/impl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/EnhancedGraphTable.java
```diff
@@ -166,6 +167,22 @@ public EnhancedGraphTable(Project project) {
...
+    val gitRepository = gitRepositorySelectionProvider.getSelectedGitRepository();
+    if (gitRepository == null) {
+      LOG.warn("Selected repository is null");
+      return;
+    }
+
+    Path macheteFilePath = gitRepository.getMacheteFilePath();
+    val macheteVFile = VirtualFileManager.getInstance().findFileByNioPath(macheteFilePath);
+    if (macheteVFile != null) {
+      macheteVFile.refresh(false, false);
```
PawelLipski, line 182: Name the params

frontend/ui/impl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/EnhancedGraphTable.java
```diff
@@ -220,7 +232,7 @@ private void inferParentForUnmanagedBranchNotificationAndNotify(String branchNam
 
     Path macheteFilePath = gitRepository.getMacheteFilePath();
     val macheteVFile = VirtualFileManager.getInstance().findFileByNioPath(macheteFilePath);
-    boolean isMacheteFilePresent = macheteVFile != null && macheteVFile.isValid();
+    boolean isMacheteFilePresent = macheteVFile != null && !macheteVFile.isDirectory();
```
PawelLipski, line 235: Maybe `isFile` instead of `!isDirectory`? if such method for checking for _regular file_ is available

frontend/ui/impl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/EnhancedGraphTable.java
```diff
@@ -167,6 +167,18 @@ public EnhancedGraphTable(Project project) {
     subscribeToMacheteFileChange();
   }
 
+  public void refreshMacheteFile(boolean asynchronous) {
```
PawelLipski, line 170: Is there a case where we'd want to call it async'ly?

branchLayout/impl/src/main/java/com/virtuslab/branchlayout/impl/readwrite/BranchLayoutWriter.java
MaciejG604: This would require putting MacheteFileUtils into branchlayout.api, is this valid?

PawelLipski: Hmmm why specifically... which method from `MacheteFileUtils` would be required in `branchlayout.api`?

frontend/ui/impl/build.gradle.kts
```diff
@@ -4,7 +4,9 @@ dependencies {
   implementation(project(":binding"))
   implementation(project(":branchLayout:api"))
   implementation(project(":backend:api"))
+  implementation(project(":frontend:actions"))
```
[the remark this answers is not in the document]
MaciejG604: It depends on the new function in `ActionUtils` -> `runWriteActionOnUIThread`.

frontend/ui/impl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/EnhancedGraphTable.java
MaciejG604: Unfortunately any method like `isFile` is not available.

frontend/ui/impl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/EnhancedGraphTable.java
MaciejG604: You're right, there's probably none.

frontend/ui/impl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/EnhancedGraphTable.java
```diff
@@ -421,83 +475,19 @@ public void refreshModel() {
...
-          /* doOnUIThreadWhenReady */ () -> {});
+              NullGitMacheteRepositorySnapshot.getInstance(),
+              /* doOnUIThreadWhenReady */ () -> {});
     } else {
       LOG.warn("Selected git repository is undefined; unable to refresh model");
     }
   }
 
   @UIEffect
-  private void refreshModel(
```
PawelLipski: Pls move method to the previous position to focus the diff on the actual changes

branchLayout/impl/src/main/java/com/virtuslab/branchlayout/impl/readwrite/BranchLayoutWriter.java
MaciejG604: Oh sorry, got the class name wrong. I meant `BranchLayoutFileUtils.deriveIndentSpec`. So `BranchLayoutFileUtils` will need to be moved into `branchlayout.api`

src/test/java/com/virtuslab/archunit/UIThreadUnsafeMethodInvocationsTestSuite.java
```diff
@@ -153,6 +153,9 @@ public void check(JavaMethod method, ConditionEvents events) {
       "java.io.File.getAbsolutePath()",
       "java.io.File.isFile()",
       "java.io.File.toString()",
+      "java.io.BufferedOutputStream.<init>(java.io.OutputStream)",
```
PawelLipski, line 156: Since these methods are already whitelisted here, there's no need to include them in `@IgnoreUIThreadUnsafeCalls` at specific methods (e.g. `com.virtuslab.gitmachete.frontend.file.MacheteFileWriter#writeBranchLayout`).

Although... TBH I'm not sure whether methods like `close()` are **really** safe to call on UI thread in general case. Depending on the I/O stream implementation, `close()` might be the place where e.g. the buffer is actually flushed to the disk. So I'd rather remove these global whitelisting, and leave `@IgnoreUIThreadUnsafeCalls` at specific methods.

branchLayout/impl/src/main/java/com/virtuslab/branchlayout/impl/readwrite/BranchLayoutWriter.java
PawelLipski: `deriveIndentSpec` an actually be moved into `BranchLayoutWriter` (as an instance, not static method), and exposed in `IBranchLayoutWriter` 🤔

branchLayout/impl/src/main/java/com/virtuslab/branchlayout/impl/readwrite/BranchLayoutWriter.java
```diff
@@ -0,0 +1,59 @@
...
+import org.checkerframework.checker.index.qual.NonNegative;
+
+import com.virtuslab.branchlayout.api.BranchLayout;
+import com.virtuslab.branchlayout.api.BranchLayoutEntry;
+import com.virtuslab.branchlayout.api.readwrite.IBranchLayoutWriter;
+import com.virtuslab.branchlayout.api.readwrite.IndentSpec;
+import com.virtuslab.qual.guieffect.IgnoreUIThreadUnsafeCalls;
+
+@CustomLog
+@RequiredArgsConstructor
```
PawelLipski, line 19: Nit: there are no required args, right? so the `@RequiredArgsConstructor` will just be the default zero-param c'tor (which is generated by default anyway)?

PawelLipski (approved):

mkondratek (changes requested):

frontend/ui/impl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/EnhancedGraphTable.java
```diff
@@ -166,6 +167,18 @@ public EnhancedGraphTable(Project project) {
     subscribeToMacheteFileChange();
   }
 
+  public void refreshMacheteFile() {
```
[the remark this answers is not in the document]
MaciejG604: It's a little hack for the UITest to work properly, basicly the UITest now modifies the machete file not using the VFS API. Because of that we cannot ensure when the change will be seen by VFS, so we tell the VFS to refresh machete file explicitly during the test. This does VFS update delay is not noticable when working with the plugin as tested with `runIde`, but the test needs to have to have consistnet results.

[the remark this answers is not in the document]
PawelLipski: Actually, pls put this comment of yours ☝🏻 directly into code (either here or in UI test code), so that it's not lost

branchLayout/impl/src/main/java/com/virtuslab/branchlayout/impl/readwrite/BranchLayoutFileUtils.java
```diff
@@ -1,21 +1,16 @@
 package com.virtuslab.branchlayout.impl.readwrite;
```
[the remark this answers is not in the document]
MaciejG604: Deleting the method, since it's neither used in tests right now nor is it necessary for mocking the default IndentSpec.

mkondratek (approved):
lgtm

PR #1306, opened 2022-10-27
Make checker stub files input to gradle java compile

### LINKED ISSUE #1300: Checker stub files (`config/checker/`) should be considered an input to Gradle JavaCompile tasks
So that when stubs change, the inputs to compilation are considered changed.

LukaszKontowski:
Closes #1300

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -63,6 +63,8 @@ tasks.register("printSignedPluginZipPath") {
   }
 }
 
+val configCheckerDirectory = rootProject.file("config/checker")
+
 configure<VersionCatalogUpdateExtension> {
   sortByKey.set(false)
 
@@ -139,6 +141,9 @@ allprojects {
     // `options.release = X` makes sure that regardless of Java version used to run the compiler,
     // only Java X-compatible APIs are available to the compiled code.
     options.release.set(Integer.parseInt(targetJavaVersion.majorVersion))
+
+    // Add files from config/checker directory as inputs to java compilation (so that changes trigger recompilation)
+    inputs.dir(configCheckerDirectory.path)
   }
 
   tasks.withType<Javadoc> {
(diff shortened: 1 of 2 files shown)


build.gradle.kts
```diff
@@ -139,6 +141,9 @@ allprojects {
     // `options.release = X` makes sure that regardless of Java version used to run the compiler,
     // only Java X-compatible APIs are available to the compiled code.
     options.release.set(Integer.parseInt(targetJavaVersion.majorVersion))
+
+    // Add files from config/checker directory as inputs to java compilation (so that changes trigger recompilation)
```
mkondratek, line 145: hmmm, but what about kotlin compile? is it enough to add it here only?

LukaszKontowski: It seems to me that the Checker Framework is for Java only? Also connected issue #1300 suggests that only JavaCompile should be affected. To be sure about this, @PawelLipski please refer.

PawelLipski: :yes:, Checker is for Java exclusively (and hence I'm reluctant to increase the presence of Kotlin in this project, given our reliance on Checker)

build.gradle.kts
```diff
@@ -63,6 +63,8 @@ tasks.register("printSignedPluginZipPath") {
   }
 }
 
+val configCheckerDirectory = rootProject.file("config/checker")
```
PawelLipski, line 66: Nit: pls wrap into an `extra(...)` as stuff above is, and refer from CheckerFramework.kt in `-Astubs=` flag

build.gradle.kts
PawelLipski: Pls mention that in the code comment actually for clarity ☝🏻

mkondratek (approved):
ok, my questions have been answered. lgtm but pls apply Pawel's comments

PawelLipski (approved):

PR #1316, opened 2022-11-02
Fix unmanaged branch notification show in case of adding a branch from an external terminal/console (closes #1308)

### LINKED ISSUE #1308: Unmanaged current branch notification is shown too... proactively
image

Two problems on this screenshot:

1. notification doesn't expire once the branch is added (it's been added via `git machete github checkout-prs`, which adds the branch to branch layout almost instantly after checking out... maybe some delay in how long the current branch is checked out would be useful)
2. notification is shown twice for the same branch

mkondratek:


CHANGE-NOTES.md
```diff
@@ -1,6 +1,7 @@
 # Changelog
 
 ## v3.2.1
+- Fixed the issue with the unmanaged branch notification showing up after adding a branch from an external terminal.
```
PawelLipski, line 4:
```suggestion
- Fixed the issue with the unmanaged branch notification showing up after adding a branch from an external (non-IntelliJ) terminal.
```

frontend/ui/impl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/EnhancedGraphTable.java
```diff
@@ -121,7 +122,7 @@ public final class EnhancedGraphTable extends BaseEnhancedGraphTable
   private String mostRecentlyCheckedOutBranch = "?!@#$%^&";
 
   @UIEffect
-  private @Nullable Notification slideInNotification;
+  private @Nullable UnmanagedBranchNotification unmanagedBranchNotification;
```
PawelLipski, line 125:
```suggestion
  private @MonotonicNonNull UnmanagedBranchNotification unmanagedBranchNotification;
```

Still compiles, but is more restrictive (only a non-null value can be assigned)

PawelLipski (approved):

PawelLipski:
Seems fixed on my Linux! 🎉

Issue #1317, opened 2022-11-03
Scale down the usage of checked exceptions as much as possible


PawelLipski:
## Steps
- [ ] Remove all (?) `throws` clauses from methods
- [ ] Review all `try...catch` stanzas (most should probably be removed)
- [x] Review all `Try...` calls (most can probably be removed)
- [ ] Where exceptions that we don't control are thrown (like `IOException`), use `@SneakyThrows`
- [ ] For exceptions that we control, either change their base to `RuntimeException` or use `@SneakyThrows`, **TBD**
- [ ] Document the guideline in CONTRIBUTING.md
- [ ] Optionally, enforce some rules with ArchUnit or Checkstyle

## Pros
- [ ] Kotlin/Scala work well without checked exceptions
- [ ] Unclutter the codebase from `Try`s and `try`s

## Cons
- [ ] Some error handling might be gone (like messages of `GitMacheteExceptions` that wrap a `GitCoreException`); probably we can sacrifice that
- [ ] Esp. see places like `com.virtuslab.gitmachete.frontend.file.highlighting.MacheteAnnotator#getParentBranchName`, here we rely on knowing that a checked exception is thrown so that we can deliberately ignore that error 🤔 

WDYT @mkondratek ❓

mkondratek:
wow, that's an interesting idea 😮 at first I thought you ve gone mad XD 

I have been taught (at the university!) to use and throw checked exceptions. And I think we did very nice exception handling and hierarchy (back in the times, with Michał P) - by the book, clear, transparent.

Kotlin/Scala-point seems to be valid, but I do not feel fully informed to consciously agree. I mean - there must be some way, probably some mechanism that is not present in Java that allows working with exceptions in a different way. Or, at least some reasoning that justifies why there are no such exceptions in these languages. Give me some time to learn more about it (I will start from [[here]](https://kotlinlang.org/docs/exceptions.html#checked-exceptions). 

I do agree that try/vavr.Try usage got a bit messy. BUT, I think we could get rid of vavr in that scope. But, bear in mind that it's a different issue and the combination of these changes with the checked exceptions thing would create a huge PR. Even though we decided to proceed with the drop of vavr.Try in the favor of a usual try, and abandon the checked exceptions, we should introduce the changes on separate PRs. 

---

tl;dr
1. Let me get more info (and reasoning) on checked exception pros/cons (in the scope of Java/Kotlin/Scala), or pls share it if you are aware of them.
[...]

PR #1325, opened 2022-11-16
Replace unnecessary Try.of with try ... catch

MaciejG604:


frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseSyncToParentByRebaseAction.java
```diff
@@ -216,9 +229,8 @@ public void run(ProgressIndicator indicator) {
         if (maybeExecutionResult != null && maybeExecutionResult.getExitCode() != 0) {
           val message = "machete-pre-rebase hooks refused to rebase (exit code ${maybeExecutionResult.getExitCode()})";
           LOG.error(message);
-          val executionResult = maybeExecutionResult;
```
PawelLipski, line 219: Actually... it looks it can be just called `executionResult` now (isn't wrapped in any `Option` or `Try`)

gitCore/jGit/src/main/java/com/virtuslab/gitcore/impl/jgit/GitCoreRepository.java
```diff
@@ -162,13 +170,20 @@ private boolean isBranchPresent(String branchFullName) {
...
       val testedPrefix = segments.take(numOfSegmentsToUse).mkString("/");
-      if (Try.of(() -> jgitRepoForMainGitDir.resolve(testedPrefix)).getOrNull() != null) {
-        return false;
-      }
+      try {
+        val objectID = jgitRepoForMainGitDir.resolve(testedPrefix);
+        if (objectID != null) {
+          return false;
+        }
+      } catch (IOException ignored) {}
```
PawelLipski, line 178: Hmmm the behavior here equivalent to what was previously? 🤔

MaciejG604: This was a tricky one, but from what I understood it goes like this:

When result of `jgitRepoForMainGitDir` is:
- **nonnull** return false.
- **null** or **exception** ignore it and move forward.

Then for the next one (line 171) result of `jgitRepoForMainGitDir` is:
- any result resturn result != null
- **exception** return false

Please correct me if I'm wrong

PawelLipski: Awww right. Indeed the logic is equivalent, no action needed

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/backgroundables/RebaseOnParentBackgroundable.java
```diff
@@ -93,72 +94,69 @@ private GitRebaseParams getIdeaRebaseParamsOf(GitRepository repository, IGitReba
...
+       * and then perform the actual rebase.
+       */
+      if (shouldExplicitlyCheckout) {
+        val uiHandler = new GitBranchUiHandlerImpl(project, indicator);
+        new GitBranchWorker(project, Git.getInstance(), uiHandler)
+            .checkout(/* reference */ gitRebaseParameters.getCurrentBranch().getName(), /* detach */ false,
+                Collections.singletonList(gitRepository));
+      }
+      GitRebaseUtils.rebase(project, Collections.singletonList(gitRepository), params, indicator);
+    } catch (GitMacheteMissingForkPointException e) {
```
PawelLipski, line 155: Hmmm should the `try { ... }` span over this entire ~50-line section? can it be narrowed down?

MaciejG604: Sure it could. So we prefer it short and setting the variable value inside of the `try { ... }`?

PawelLipski: Yeah, let's do it this way, even at the expense of one more `var` (rather than `val`)

backend/impl/src/main/java/com/virtuslab/gitmachete/backend/impl/GitMacheteRepositoryCache.java
```diff
@@ -36,7 +36,7 @@ public IGitMacheteRepository getInstance(Path rootDirectoryPath, Path mainGitDir
     val key = Tuple.of(rootDirectoryPath, worktreeGitDirectoryPath);
     val valueReferenceOption = gitMacheteRepositoryCache.get(key);
```
PawelLipski, line 37: Nit: while you're here, just convert to `@Nullable` (some `getOrNull()` or sth) and then check `!= null`

PawelLipski (approved):

PawelLipski (approved):

backend/impl/src/main/java/com/virtuslab/gitmachete/backend/impl/hooks/StatusBranchHookExecutor.java
```diff
@@ -134,11 +134,9 @@ public static String hashFile(String algorithm, File f) throws IOException, NoSu
     var hookContentMD5Hash = "";
     try {
       hookContentMD5Hash = hashFile("MD5", hookFile);
-    } catch (IOException e) {
+    } catch (IOException | NoSuchAlgorithmException ignored) {
       // we are using the constant empty String as a neutral value, so that the functionality would
```
mkondratek, line 138: nit: the capital letter "We ..."

frontend/base/src/main/java/com/virtuslab/gitmachete/frontend/vfsutils/GitVfsUtils.java
```diff
@@ -1,5 +1,6 @@
 package com.virtuslab.gitmachete.frontend.vfsutils;
```
mkondratek, line 1: do we still need to depend on vavr in all modules? it cannot be dropped anywhere yet? or can it be?

gitCore/api/src/main/java/com/virtuslab/gitcore/api/IGitCoreBranchSnapshot.java
```diff
@@ -40,11 +39,11 @@ static boolean defaultEquals(@FindDistinct IGitCoreBranchSnapshot self, @Nullabl
     } else {
       val o = (IGitCoreBranchSnapshot) other;
       return self.getFullName().equals(o.getFullName())
-          && Try.of(() -> self.getPointedCommit().equals(o.getPointedCommit())).getOrElse(false);
```
mkondratek, line 43: why was it try?

PawelLipski: I'd assume that `getPointedCommit()` originally could throw an exception... @MaciejG604 if you could check what was the original implementation (`git bisect` might be useful, given the size of the history)

frontend/base/src/main/java/com/virtuslab/gitmachete/frontend/vfsutils/GitVfsUtils.java
MaciejG604: Checked all the modules I touched here, every one of them still uses at least one class from vavr. So yes, we need the dependencies.

mkondratek: ok, thanks!

gitCore/api/src/main/java/com/virtuslab/gitcore/api/IGitCoreBranchSnapshot.java
MaciejG604: Commit 93bb80728a17c285e11a36708ee44583c1d6ad2a

mkondratek (approved):

gitCore/api/src/main/java/com/virtuslab/gitcore/api/IGitCoreBranchSnapshot.java
PawelLipski: Gr8... so indeed the `throws` clause has been dropped from the method, but `Try.of` unnecessarily retained... makes sense

PawelLipski (approved):

PawelLipski:
Lemme unlink from the issue #1317... and move that issue to be done at some point in the future. This PR can ofc be shipped regardless

PawelLipski (approved):

PawelLipski (approved):

Issue #1079, opened 2022-08-22
`squashBranch` UI test failing: `expected:<15> but was:<16>`


PawelLipski:
This is rare, so far only seen in:

https://app.circleci.com/pipelines/github/VirtusLab/git-machete-intellij-plugin/6116/workflows/50e67100-2a7f-445d-bb56-406b24cbf9e3/jobs/6459/steps:

```
com.virtuslab.gitmachete.uitest.UITestSuite > squashBranch STANDARD_OUT
    [intellij-stdout] closed com.intellij.ui.Splash
    IntelliJ build number is 2021.3.3
    request[projects/all]: ()
    response: []
    [intellij-stdout] 2022-08-22 10:29:17,891 [   8388]   WARN - ctionSystem.impl.ActionUpdater - 215 ms to grab EDT for LargeIconWithTextWrapper#update (com.intellij.openapi.wm.impl.welcomeScreen.WelcomeScreenActionsUtil$LargeIconWithTextWrapper) 
    [intellij-stdout] Info  | RdCoroutineScope          | 11:JobScheduler FJ pool 0/3 | RdCoroutineHost overridden 
    request[project/open]: "/tmp/machete-tests-410347444574336497/machete-sandbox-worktree"
    [intellij-stdout] opened com.intellij.openapi.wm.impl.IdeFrameImpl
    [intellij-stdout] closed Welcome to IntelliJ IDEA (ROOT)
    [intellij-stdout] 2022-08-22 10:29:23,094 [  13591]   WARN - ctionSystem.impl.ActionUpdater - 848 ms to grab EDT for CWMTelephonyGroup#update (com.jetbrains.rd.platform.codeWithMe.telephony.CWMTelephonyGroup) 
    response: ()
    Waiting for 3 background task(s)... [com.intellij.openapi.progress.impl.NonCancelableIndicator$1@0, ProgressIndicatorEx 1289239017: running=true; canceled=false; reusable=true, ProgressIndicatorEx 989899735: running=true; canceled=false; reusable=true]
    [intellij-stdout] 2022-08-22 10:29:23,810 [  14307]   WARN - ctionSystem.impl.ActionUpdater - 104 ms to call on EDT SyncProjectAction#update (com.android.tools.idea.gradle.actions.SyncProjectAction). Use `UpdateInBackground`. 
[...]

PR #1332, opened 2022-11-21
Add a more specialized test for squash (UI)

mkondratek:
not a solution. only better testing for https://github.com/VirtusLab/git-machete-intellij-plugin/issues/1079

### DIFF
--- a/src/uiTest/resources/project.rhino.js
+++ b/src/uiTest/resources/project.rhino.js
@@ -105,7 +105,7 @@ function Project(underlyingProject) {
         const branchesAndCommits = new LinkedList();
         if (b.isNonRoot()) {
           const nonRoot = b.asNonRoot();
-          branchesAndCommits.addAll(nonRoot.getCommitsUntilParent().map(c => c.getShortMessage()).toJavaList());
+          branchesAndCommits.addAll(nonRoot.getUniqueCommits().map(c => c.getShortMessage()).toJavaList());
         }
         branchesAndCommits.add(b.getName());
         return branchesAndCommits;
--- a/src/uiTest/resources/project.rhino.js
+++ b/src/uiTest/resources/project.rhino.js
@@ -98,6 +98,23 @@ function Project(underlyingProject) {
     }
   };
 
+  this.getManagedBranchesAndCommits = function () {
+    const snapshot = getGraphTable().getGitMacheteRepositorySnapshot();
+    if (snapshot != null) {
+      return snapshot.getManagedBranches().flatMap(b => {
+        const branchesAndCommits = new LinkedList();
+        if (b.isNonRoot()) {
+          const nonRoot = b.asNonRoot();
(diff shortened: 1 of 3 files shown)


src/uiTest/scala/com/virtuslab/gitmachete/uitest/UITestSuite.scala
```diff
@@ -232,21 +232,65 @@ class UITestSuite extends BaseGitRepositoryBackedIntegrationTestSuite(SETUP_WITH
...
+        "HOTFIX Add the trigger - fixes",
+        "Use new icons",
+        "Use new icons",
+        "allow-ownership-link",
+        "build-chain",
+        "call-ws",
+        "develop",
+        "hotfix/add-trigger",
+        "master",
+        "update-icons"
```
PawelLipski, line 260: Hmmm this ordering is alphabetical atm, right? 🤔 would be good to fix #1334 first

PawelLipski: So that the ordering tells more which commit belongs to which branch

PawelLipski:
> java.lang.AssertionError: expected:

Hmm okay so you've managed to repro the #1079 bug locally right?

Anyway, lemme refrain from diagnosis until is #1334 is addressed (so that there's more clarity as to what's compared to what) 🤔

mkondratek:
> > java.lang.AssertionError: expected:
> 
> Hmm okay so you've managed to repro the #1079 bug locally right?
> 
> Anyway, lemme refrain from diagnosis until is #1334 is addressed (so that there's more clarity as to what's compared to what) 🤔

no no no, my question is about `getCommitsUntilParent` method - why does is return `Develop commit` (and probably a more important question, why `isNonRoot` got true on the develop branch)?

src/uiTest/resources/project.rhino.js
```diff
@@ -98,6 +98,23 @@ function Project(underlyingProject) {
...
   };
 
+  this.getManagedBranchesAndCommits = function () {
+    const snapshot = getGraphTable().getGitMacheteRepositorySnapshot();
+    if (snapshot != null) {
+      return snapshot.getManagedBranches().flatMap(b => {
+        const branchesAndCommits = new LinkedList();
+        if (b.isNonRoot()) {
+          const nonRoot = b.asNonRoot();
+          branchesAndCommits.addAll(nonRoot.getUniqueCommits().map(c => c.getShortMessage()).toJavaList());
```
mkondratek, line 108: it's here exactly - using `getUniqueCommits` works as expected (at least to me), but with `getCommitsUntilParent` the `Develop commit` is included in the result. 

Probably it's not from the develop branch (`isNonRoot` for develop works fine) but some other branch decided that the parent is root (not develop as it should) 🤔

PawelLipski: Wow okay.... weird 😯

PawelLipski (approved):

Issue #1299, opened 2022-10-24
Reset Info dialog is too small


PawelLipski:
![image](https://user-images.githubusercontent.com/3383210/197568903-290921db-e577-48ed-9f57-28dff4d66807.png)

Scrolling is needed, and that obfuscates the somewhat non-trivial message. The dialog should be wider and higher by default 🤔

PR #1336, opened 2022-11-21
Revamp ResetInfoDialog into a more general InfoDialog

mkondratek:
Why? 

1. because we will need a similar dialog for the traverse feature (and probably somewhere else in the future) 
2. yesNoDialog (etc) have this problem: https://github.com/VirtusLab/git-machete-intellij-plugin/issues/1299


PawelLipski (approved):
:benissimo:

PR #1337, opened 2022-11-21
Ensure `GitMacheteRepositorySnapshot#getManagedBranches` returns branches in the order as in .git/machete

PawelLipski:
1️⃣ 3️⃣ 3️⃣ 7️⃣

### DIFF
--- a/backend/impl/src/main/java/com/virtuslab/gitmachete/backend/impl/GitMacheteRepository.java
+++ b/backend/impl/src/main/java/com/virtuslab/gitmachete/backend/impl/GitMacheteRepository.java
@@ -17,9 +17,9 @@ import com.jcabi.aspects.Loggable;
 import io.vavr.Tuple;
 import io.vavr.Tuple2;
 import io.vavr.collection.HashMap;
+import io.vavr.collection.LinkedHashMap;
 import io.vavr.collection.List;
 import io.vavr.collection.Map;
-import io.vavr.collection.Queue;
 import io.vavr.collection.Seq;
 import io.vavr.collection.Set;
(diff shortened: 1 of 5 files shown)


backend/impl/src/main/java/com/virtuslab/gitmachete/backend/impl/GitMacheteRepository.java
```diff
@@ -374,15 +374,15 @@ IGitMacheteRepositorySnapshot createSnapshot(BranchLayout branchLayout) throws G
...
-      Queue<IManagedBranchSnapshot> queue = Queue.ofAll(rootBranches);
-      // BFS over all branches
-      while (queue.nonEmpty()) {
-        val headAndTail = queue.dequeue();
-        val branch = headAndTail._1;
+    private LinkedHashMap<String, IManagedBranchSnapshot> createManagedBranchByNameMap(
+        List<RootManagedBranchSnapshot> rootBranches) {
+      LinkedHashMap<String, IManagedBranchSnapshot> branchByName = LinkedHashMap.empty();
+      List<IManagedBranchSnapshot> stack = List.ofAll(rootBranches);
+      // Non-recursive DFS over all branches
```
mkondratek, line 381: nit:
1. **A** non-recursive DFS over all branches
2. maybe add a comment about why we use `LinkedHashMap` (saying that it is ordered collection and will guarantee us the order of branches in the end)

mkondratek (approved):
cool, the tests are much more readable and intuitive now 🚀

mkondratek (approved):
awesome

PR #1340, opened 2022-11-21
Ensure that `EnhancedGraphTable#trackCurrentBranchChange` is actually always executed on EDT

PawelLipski:


frontend/ui/impl/src/main/java/com/virtuslab/gitmachete/frontend/ui/impl/table/EnhancedGraphTable.java
```diff
@@ -297,6 +300,7 @@ private void trackCurrentBranchChange(GitRepository repository) {
         mostRecentlyCheckedOutBranch = repositoryCurrentBranchName;
       }
     }
+    queueRepositoryUpdateAndModelRefresh();
```
mkondratek, line 303: why this change?

PawelLipski: So that `queueRepositoryUpdateAndModelRefresh()` happens actually **after** `mostRecentlyCheckedOutBranch` is assigned. The alternative would be sth like:

```java
@AlwaysSafe GitRepositoryChangeListener listener = repository -> {
ModalityUiUtil.invokeLaterIfNeeded(NON_MODAL,
        () -> trackCurrentBranchChange(repository)); // no queueRepositoryUpdateAndModelRefresh() call inside trackCurrentBranchChange
  queueRepositoryUpdateAndModelRefresh();
}
```

...which will also probably be okay (since `queueRepositoryUpdateAndModelRefresh()` will queue the action to execute on UI thread **after** `trackCurrentBranchChange(repository)`)... but the current approach is clearer if it comes to what happens after what :async-parrot:

mkondratek: oh, okay. I see now it is moved from the other place. ok then!

PR #1350, opened 2022-11-23
Prevent NPE in 'SlideOutBackgroundable'

MaciejG604:


frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/backgroundables/SlideOutBackgroundable.java
```diff
@@ -134,6 +134,7 @@ public void run(ProgressIndicator indicator) {
 
   @UIThreadUnsafe
   private void slideOutBranch(String branchName) {
+    val selectedGitRepository = this.selectedGitRepository;
```
PawelLipski, line 137: I would use a different name for un-ambiguity, like `theSelectedGitRepository`

PawelLipski (approved):

PR #1352, opened 2022-11-24
Add the traverse action (closes #246)

### LINKED ISSUE #246: Add traverse feature

mkondratek:


CHANGE-NOTES.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+## v3.3.0
+- Added the interactive traverse feature.
```
PawelLipski, line 4: I'd call it
```suggestion
- Added the semi-automatic traversal feature.
```
It's not the _interactivity_ where the traversal stands out, rather the automaticity 🤖

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/traverse/TraverseAction.java
```diff
@@ -0,0 +1,126 @@
...
+
+  private void checkoutAndTraverse(GitRepository gitRepository, BaseEnhancedGraphTable graphTable, String branchName) {
+    log().debug(() -> "Queuing '${branchName}' branch checkout background task");
+
+    @UI Runnable traverseRunnable = () -> traverse(gitRepository, graphTable, branchName);
+    Runnable repositoryRefreshRunnable = () -> graphTable.queueRepositoryUpdateAndModelRefresh(traverseRunnable);
+    Runnable callInAwtLater = () -> ModalityUiUtil.invokeLaterIfNeeded(NON_MODAL, repositoryRefreshRunnable);
+    val gitBrancher = GitBrancher.getInstance(gitRepository.getProject());
+    val repositories = Collections.singletonList(gitRepository);
+    gitBrancher.checkout(/* reference */ branchName, /* detach */ false, repositories, callInAwtLater);
```
PawelLipski, line 102: Similarly to how `git machete traverse` operates, I'd skip checking out the branches for which there's nothing to be done anyway

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/traverse/TraverseAction.java
```diff
@@ -0,0 +1,126 @@
...
+  }
```
PawelLipski, line 103: Also... probably unrealistic, but it'd be great if indexing could **not** trigger for the freshly checked-out branches 😅

CHANGE-NOTES.md
mkondratek: I do not like the name change (traverse -> traversal) - why don't we stick to CLI nomenclature?

PawelLipski: Right, _traverse_ is a noun as well, let's NOT use _traversal_ then

image

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/traverse/TraverseAction.java
mkondratek: yup, unrealistic 😢

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/traverse/TraverseAction.java
mkondratek: done! however there was one advantage of checking out the branch - the branch under operation was always indicated in the graph table this way. but I agree - it's better to avoid checkout if not needed

PawelLipski: As discussed, let's also check out the branch to be pushed/slid out (not only the ones to be rebased)

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/traverse/TraverseAction.java
PawelLipski: The problem is in auto-reloading of External Systems, like Gradle

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/traverse/TraverseSyncToRemote.java
```diff
@@ -0,0 +1,221 @@
...
+      case FORCE_PUSH :
+        new GitPushDialog(project, List.of(gitRepository), GitPushSource.create(localBranch),
+            /* isForcePushRequired */ true).show();
+        break;
+
+      case RESET_ON_REMOTE :
+        new ResetCurrentToRemoteBackgroundable(
+            getString("action.GitMachete.BaseResetToRemoteAction.task-title"),
+            /* canBeCancelled */ true, gitMacheteBranch.getName(), remoteTrackingBranch.getName(),
+            gitRepository);
```
zmerr, line 176: Maybe here, it might be necessary to invoke or `queue`.

Also, the override of `onSuccess` might be necessary to make sure the next branch, which might be a child of the current one, does not sync to parent before the reset of parent on its remote gets finished.

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/traverse/TraverseAction.java
PawelLipski: Huh in the option 2... you mean
> do checkout, slide out, but do not allow for delete

right? I'd go for that, that's what CLI does anyway in `git machete traverse` (it still doesn't have branch deletion on slide-out implemented :trollface:)

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/traverse/TraverseSyncToRemote.java
mkondratek: good point about missing `queue`! thanks!

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/traverse/TraverseAction.java
mkondratek: ok, that's the easiest and actually done by always checking out the branch during the traverse

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/traverse/TraverseSyncToRemote.java
zmerr: hmmm 🤔wouldn’t the run method itself be run concurrently to the existing thread?
or I might be missing something 😕

mkondratek: yes, you are right, but thanks to backgroundables we have `onSuccess` to get such a job done! we are cool 🚀

mkondratek: in other words, it can be concurrent - we only need a guaranty that `run` finishes what need to be finished and then runs `onSuccess`. that is what happens

mkondratek: but there is the other "issue" with reset... I will start this discussion in the traverse feature issue

zmerr: oops, the `onSuccess` does not seems to be overrided in `ResetCurrentToRemoteBackgroundable` or I might be missing sth🧐

frontend/base/src/main/java/com/virtuslab/gitmachete/frontend/defs/ActionIds.java
```diff
@@ -4,6 +4,7 @@ public final class ActionIds {
   private ActionIds() {}
 
   public static final String CHECK_OUT = "GitMachete.CheckoutSelectedAction";
```
PawelLipski, line 6: Maybe
```suggestion
  public static final String CHECK_OUT_SELECTED = "GitMachete.CheckoutSelectedAction";
```

frontend/resourcebundles/src/main/resources/GitMacheteBundle.properties
```diff
@@ -460,3 +470,65 @@ action.GitMachete.CheckoutParentAction.undefined.branch-name=Branch (''{0}'') do
...
 action.GitMachete.CheckoutFirstChildAction.undefined.branch-name=Branch (''{0}'') does not have any children in the branch layout
+
+# Traverse
+
+action.GitMachete.BaseTraverseAction.description=Traverse
+action.GitMachete.BaseTraverseAction.description.empty-layout=Cannot traverse an empty branch layout
+
+action.GitMachete.BaseTraverseAction.dialog.traverse-approval.title=Traverse Info
+action.GitMachete.BaseTraverseAction.dialog.traverse-approval.text.HTML=<html>\
+  This action triggers an interactive walk over the branches and suggests actions (pushes/pulls/rebases) \
```
PawelLipski, line 481: Theoretically, slide-out could also be mentioned here... but maybe it'd be too verbose, esp. for the first-time viewers

frontend/resourcebundles/src/main/resources/GitMacheteBundle.properties
```diff
@@ -460,3 +470,65 @@ action.GitMachete.CheckoutParentAction.undefined.branch-name=Branch (''{0}'') do
...
+  necessary to achieve both child-to-parent and local-to-remote synchronization.<br><br>\
+  Do you want to proceed?\
+  </html>
+
+action.GitMachete.BaseTraverseAction.dialog.diverged-from-remote.text.HTML=\
+  <html>\
+    <p>The branch <b>{0}</b> is diverged from remote <b>{1}</b>, and it is <b>{2}</b> remote.</p> \
+    <p>Please choose the preferred action for syncing it:</p>\
```
PawelLipski, line 489: Okay... so the change with opinionated action on diverged-from-remote is still pending/TODO right?

mkondratek: nope, I simply missed this one - the message should be different in the case of diverged and newer ofc

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/traverse/BaseTraverseAction.java
```diff
@@ -0,0 +1,129 @@
...
+  @Override
+  public LambdaLogger log() {
+    return LOG;
+  }
+
+  public static final String SHOW_TRAVERSE_INFO = "git-machete.traverse.approval.show";
+
+  @Override
+  @UIEffect
+  protected void onUpdate(AnActionEvent anActionEvent) {
```
PawelLipski, line 41: A general sanity check: is it ever possible to run another traverse when a traverse is ongoing? 🤔

PawelLipski: (ofc it shouldn't, that would be a race-condition-ridden nightmare)

frontend/resourcebundles/src/main/resources/GitMacheteBundle.properties
```diff
@@ -460,3 +470,65 @@ action.GitMachete.CheckoutParentAction.undefined.branch-name=Branch (''{0}'') do
...
+  </html>
+action.GitMachete.BaseTraverseAction.dialog.diverged-from-remote.title=Resolve Divergence from Remote
+action.GitMachete.BaseTraverseAction.dialog.diverged-from-remote.newer-than-remote=older than
+action.GitMachete.BaseTraverseAction.dialog.diverged-from-remote.older-than-remote=newer than
+
+action.GitMachete.BaseTraverseAction.dialog.diverged-from-remote.do-not-resolve-option=Do not sync
+action.GitMachete.BaseTraverseAction.dialog.diverged-from-remote.do-not-resolve-option.comment=Note: the remote and local changes will remain intact; traverse will continue
+action.GitMachete.BaseTraverseAction.dialog.diverged-from-remote.force-push-option=Force push
+action.GitMachete.BaseTraverseAction.dialog.diverged-from-remote.force-push.comment=Warning: you will lose the remote changes
```
PawelLipski, line 498: `lose` wording to be reconsidered, as discussed

frontend/resourcebundles/src/main/resources/GitMacheteBundle.properties
```diff
@@ -460,3 +470,65 @@ action.GitMachete.CheckoutParentAction.undefined.branch-name=Branch (''{0}'') do
...
+  <html>\
+    <p>The child branch is merged to the parent branch.</p> \
+    <p>Slide out <b>{0}</b>?</p>\
+  </html>
+
+action.GitMachete.BaseTraverseAction.dialog.pull-verification.title=Git Machete Traverse - Pull
+action.GitMachete.BaseTraverseAction.dialog.pull-verification.text.HTML=\
+  <html>\
+    <p>The branch <b>{0}</b> is behind its remote tracking branch <b>{1}</b>.</p> \
+    <p>Do you want to <b>pull</b> the remote changes?</p>\
```
PawelLipski, line 521: The grammar is inconsistent: sometimes it's `X?` and sometimes `Do you want to X?` 🤔

frontend/resourcebundles/src/main/resources/GitMacheteBundle.properties
```diff
@@ -460,3 +470,65 @@ action.GitMachete.CheckoutParentAction.undefined.branch-name=Branch (''{0}'') do
...
+action.GitMachete.BaseTraverseAction.dialog.push-verification.title=Git Machete Traverse - Push
+action.GitMachete.BaseTraverseAction.dialog.push-verification.ahead.text.HTML=\
+  <html>\
+    <p>The branch <b>{0}</b> is ahead of its remote tracking branch <b>{1}</b>.</p> \
+    <p>Do you want to <b>push</b> the changes to the remote?</p>\
+  </html>
+action.GitMachete.BaseTraverseAction.dialog.push-verification.untracked.text.HTML=\
+  <html>\
+    <p>The branch <b>{0}</b> is untracked.</p> \
+    <p>Do you want to <b>push</b> the changes the remote?</p>\
```
PawelLipski, line 533:
```suggestion
    <p>Do you want to <b>push</b> the changes to the remote?</p>\
```

frontend/resourcebundles/src/main/resources/GitMacheteBundle.properties
```diff
@@ -460,3 +470,65 @@ action.GitMachete.CheckoutParentAction.undefined.branch-name=Branch (''{0}'') do
...
+action.GitMachete.BaseTraverseAction.dialog.diverged-from-parent.text.HTML=\
+  <html>\
+    <p>The child branch is diverged from the parent branch.</p> \
+    <p>Rebase <b>{0}</b> branch onto <b>{1}</b>?</p>\
+  </html>
+
+action.GitMachete.BaseTraverseAction.dialog.merged-to-parent.title=Resolve Merged to Parent
+action.GitMachete.BaseTraverseAction.dialog.merged-to-parent.text.HTML=\
+  <html>\
+    <p>The child branch is merged to the parent branch.</p> \
```
PawelLipski, line 513: I'd name the child & parent branch explicitly here, so that it's clear that it's the **child** that's to be slid out

frontend/resourcebundles/src/main/resources/GitMacheteBundle.properties
```diff
@@ -460,3 +470,65 @@ action.GitMachete.CheckoutParentAction.undefined.branch-name=Branch (''{0}'') do
...
+action.GitMachete.BaseTraverseAction.dialog.diverged-from-remote.force-push-option=Force push
+action.GitMachete.BaseTraverseAction.dialog.diverged-from-remote.force-push.comment=Warning: you will lose the remote changes
+action.GitMachete.BaseTraverseAction.dialog.diverged-from-remote.reset-option=Reset to remote
+action.GitMachete.BaseTraverseAction.dialog.diverged-from-remote.reset-on-remote.comment=Warning: you will lose the local changes
+action.GitMachete.BaseTraverseAction.dialog.cancel-traverse=Quit the interactive traverse
+
+action.GitMachete.BaseTraverseAction.dialog.diverged-from-parent.title=Resolve Divergence from Parent
+action.GitMachete.BaseTraverseAction.dialog.diverged-from-parent.text.HTML=\
+  <html>\
+    <p>The child branch is diverged from the parent branch.</p> \
```
PawelLipski, line 506: As below, name the involved branches explicitly (as IIRC CLI does in a similar context)

src/test/java/com/virtuslab/archunit/ClassStructureTestSuite.java
```diff
@@ -105,6 +105,9 @@ public void all_classes_should_be_referenced() {
             // doesn't see accesses to static fields
             "com.virtuslab.gitmachete.frontend.defs")
         .and().doNotBelongToAnyOf(classesReferencedFromPluginXmlAttributes)
+        .and()
+        .doNotBelongToAnyOf(com.virtuslab.gitmachete.frontend.actions.traverse.TraverseSyncToParent.class,
```
PawelLipski, line 109: Hmmm are these classes truly unreferenced? or a false positive in the test?

frontend/resourcebundles/src/main/resources/GitMacheteBundle.properties
```diff
@@ -221,6 +221,11 @@ action.GitMachete.SyncSelectedToParentByMergeAction.warning-dialog.message.HTML=
   </html>
 
 
+action.GitMachete.TraverseFromSelectedAction.text=Git Machete: Traverse from Selected Branch\u2026
+action.GitMachete.TraverseFromSelectedAction.GitMacheteContextMenu.text=T_raverse from Selected Branch\u2026
+action.GitMachete.TraverseFromSelectedAction.description=Traverse from selected branch
```
PawelLipski, line 226: Hmm

image

It's now the sole action in context menu with `...from Selected Branch` qualifier (even though all others act on the selected branch in some way &mdash; this "acting on selected branch" is implied).

Maybe hmm... `Traverse from Here...`?

PawelLipski: Also, small inconsistency: `In` and `Out` after `Slide` are the sole capitalized prepositions... but maybe that's intentional

frontend/resourcebundles/src/main/resources/GitMacheteBundle.properties
```diff
@@ -460,3 +470,65 @@ action.GitMachete.CheckoutParentAction.undefined.branch-name=Branch (''{0}'') do
...
+  <html>\
+    <p>The branch <b>{0}</b> is diverged from remote <b>{1}</b>, and it is <b>{2}</b> remote.</p> \
+    <p>Please choose the preferred action for syncing it:</p>\
+  </html>
+action.GitMachete.BaseTraverseAction.dialog.diverged-from-remote.title=Resolve Divergence from Remote
+action.GitMachete.BaseTraverseAction.dialog.diverged-from-remote.newer-than-remote=older than
+action.GitMachete.BaseTraverseAction.dialog.diverged-from-remote.older-than-remote=newer than
+
+action.GitMachete.BaseTraverseAction.dialog.diverged-from-remote.do-not-resolve-option=Do not sync
+action.GitMachete.BaseTraverseAction.dialog.diverged-from-remote.do-not-resolve-option.comment=Note: the remote and local changes will remain intact; traverse will continue
```
PawelLipski, line 496: Maybe:
```suggestion
action.GitMachete.BaseTraverseAction.dialog.diverged-from-remote.do-not-resolve-option.comment=Note: the remote and local branches will remain intact; traverse will continue
```

PawelLipski: Also:

image

but if that wrapping is hard to fix, let's leave as is

frontend/resourcebundles/src/main/resources/GitMacheteBundle.properties
```diff
@@ -460,3 +470,65 @@ action.GitMachete.CheckoutParentAction.undefined.branch-name=Branch (''{0}'') do
...
+action.GitMachete.BaseTraverseAction.dialog.diverged-from-remote.force-push-option=Force push
+action.GitMachete.BaseTraverseAction.dialog.diverged-from-remote.force-push.comment=Warning: you will lose the remote changes
+action.GitMachete.BaseTraverseAction.dialog.diverged-from-remote.reset-option=Reset to remote
+action.GitMachete.BaseTraverseAction.dialog.diverged-from-remote.reset-on-remote.comment=Warning: you will lose the local changes
```
PawelLipski, line 500: `lose` TBD; also `s/changes/commits` to be more precise

src/main/resources/META-INF/plugin.xml
```diff
@@ -135,6 +141,12 @@
                 <override-text place="GitMacheteToolbar"/>
             </action>
 
+            <action id="GitMachete.TraverseFromCurrentAction"
+                    class="com.virtuslab.gitmachete.frontend.actions.toolbar.TraverseFromCurrentAction"
+                    icon="com.virtuslab.gitmachete.frontend.icons.MacheteIcons.TRAVERSE">
+                <override-text place="GitMacheteToolbar"/>
```
PawelLipski, line 147: The toolbar's becoming pretty crowded:

image

How about moving the relatively rarely used discover to the end, e.g. before the ❓ ?

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/traverse/BaseTraverseAction.java
```diff
@@ -0,0 +1,129 @@
...
+@ExtensionMethod({GitVfsUtils.class, GitMacheteBundle.class})
+@CustomLog
+public abstract class BaseTraverseAction extends BaseGitMacheteRepositoryReadyAction implements IBranchNameProvider {
+
+  @Override
+  public LambdaLogger log() {
+    return LOG;
+  }
+
+  public static final String SHOW_TRAVERSE_INFO = "git-machete.traverse.approval.show";
```
PawelLipski, line 37: #1382

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/traverse/BaseTraverseAction.java
```diff
@@ -0,0 +1,129 @@
...
+            getString("action.GitMachete.BaseTraverseAction.dialog.traverse-approval.title"),
+            getString("action.GitMachete.BaseTraverseAction.dialog.traverse-approval.text.HTML"),
+            SHOW_TRAVERSE_INFO,
+            /* myHeight */ 100);
+
+        yesNoResult = traverseInfoDialog.showAndGet();
+      }
+
+      if (yesNoResult) {
+        val entryName = getNameOfBranchUnderAction(anActionEvent);
```
PawelLipski, line 85: Maybe just
```suggestion
        val startBranchName = getNameOfBranchUnderAction(anActionEvent);
```

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/traverse/BaseTraverseAction.java
```diff
@@ -0,0 +1,129 @@
...
+
+  private void checkoutAndTraverse(GitRepository gitRepository, BaseEnhancedGraphTable graphTable, String branchName) {
+    log().debug(() -> "Queuing '${branchName}' branch checkout background task");
+
+    @UI Runnable traverseRunnable = () -> traverse(gitRepository, graphTable, branchName);
+    Runnable repositoryRefreshRunnable = () -> graphTable.queueRepositoryUpdateAndModelRefresh(traverseRunnable);
+    Runnable callInAwtLater = () -> ModalityUiUtil.invokeLaterIfNeeded(NON_MODAL, repositoryRefreshRunnable);
+    val gitBrancher = GitBrancher.getInstance(gitRepository.getProject());
+    val repositories = Collections.singletonList(gitRepository);
+    gitBrancher.checkout(/* reference */ branchName, /* detach */ false, repositories, callInAwtLater);
```
PawelLipski, line 103: Hmmm since this `Runnable` is supposed to be executed in EDT, I think that `ModalityUiUtil.invokeLaterIfNeeded` shouldn't be necessary. Also, `queueRepositoryUpdateAndModelRefresh` and hence `repositoryRefreshRunnable` isn't even UI-effecting.

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/traverse/BaseTraverseAction.java
```diff
@@ -0,0 +1,129 @@
...
+    if (gitMacheteBranch != null) {
+      @UI Runnable traverseNextEntry = () -> {
+        var nextBranch = branchLayout != null ? branchLayout.findNextEntry(branchName) : null;
+        if (nextBranch != null) {
+          checkoutAndTraverse(gitRepository, graphTable, nextBranch.getName());
+        }
+      };
+
+      if (gitMacheteBranch.isNonRoot()) {
+        assert repositorySnapshot != null : "repositorySnapshot is null";
```
PawelLipski, line 121: Huh Checker should get that `repositorySnapshot` can't be null here... anyway, rather than putting a assertion, do sth like that at the  beginning of the method:

```java
val repositorySnapshot = graphTable.getGitMacheteRepositorySnapshot();
if (repositorySnapshot == null) return;
```

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/traverse/BaseTraverseAction.java
```diff
@@ -0,0 +1,129 @@
...
+        new TraverseSyncToParent(gitRepository, graphTable, repositorySnapshot, gitMacheteBranch.asNonRoot(), traverseNextEntry)
+            .sync();
+      } else {
+        new TraverseSyncToRemote(gitRepository, graphTable, gitMacheteBranch, traverseNextEntry).sync();
```
PawelLipski, line 125: Huh I don't like a method called `sync()` that executes mostly **a**synchronously ;D

I mean, ofc _sync_ is unrelated here to _(a)synchronous execution_... but still, a bit misleading (even if only on subconscious level).

Maybe just `execute()`?

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/traverse/BaseTraverseAction.java
```diff
@@ -0,0 +1,129 @@
...
+        if (entryName != null) {
+          checkoutAndTraverse(gitRepository, graphTable, entryName);
+        } else {
+          LOG.warn("Skipping traverse action because entryName is undefined");
+        }
+      }
+    }
+  }
+
+  private void checkoutAndTraverse(GitRepository gitRepository, BaseEnhancedGraphTable graphTable, String branchName) {
```
PawelLipski, line 95: Nit:
```suggestion
  private void checkoutAndTraverseFrom(GitRepository gitRepository, BaseEnhancedGraphTable graphTable, String branchName) {
```

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/traverse/BaseTraverseAction.java
```diff
@@ -0,0 +1,129 @@
...
+    @UI Runnable traverseRunnable = () -> traverse(gitRepository, graphTable, branchName);
+    Runnable repositoryRefreshRunnable = () -> graphTable.queueRepositoryUpdateAndModelRefresh(traverseRunnable);
+    Runnable callInAwtLater = () -> ModalityUiUtil.invokeLaterIfNeeded(NON_MODAL, repositoryRefreshRunnable);
+    val gitBrancher = GitBrancher.getInstance(gitRepository.getProject());
+    val repositories = Collections.singletonList(gitRepository);
+    gitBrancher.checkout(/* reference */ branchName, /* detach */ false, repositories, callInAwtLater);
+  }
+
+  @UIEffect
+  private void traverse(GitRepository gitRepository, BaseEnhancedGraphTable graphTable, String branchName) {
```
PawelLipski, line 107: And here
```suggestion
  private void traverseFrom(GitRepository gitRepository, BaseEnhancedGraphTable graphTable, String branchName) {
```

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/traverse/TraverseSyncToParent.java
```diff
@@ -0,0 +1,133 @@
...
+    switch (slideOutDialog.show(project)) {
+      case MessageConstants.YES :
+        new SlideOutBackgroundable(getString("action.GitMachete.BaseSlideOutAction.task.title"),
+            gitMacheteBranch, gitRepository, currentBranchIfManaged, branchLayout, graphTable) {
+          @Override
+          public void onSuccess() {
+            graphTable.queueRepositoryUpdateAndModelRefresh(traverseNextEntry);
+          }
+        }.queue();
+        return false;
```
PawelLipski, line 91: Huh that's tricky ;D so it's now down to the newly-created backgroundable and NOT down to `sync()` to continue the traversal 🤔

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/traverse/TraverseSyncToParent.java
```diff
@@ -0,0 +1,133 @@
...
+
+      case MessageConstants.NO :
+        break;
```
PawelLipski, line 94: Nit: maybe
```suggestion
        return true;
```

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/traverse/TraverseSyncToParent.java
PawelLipski: I'd add a comment in code about this for clarity

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/traverse/BaseTraverseAction.java
```diff
@@ -0,0 +1,129 @@
...
+    val repositorySnapshot = graphTable.getGitMacheteRepositorySnapshot();
+    val branchLayout = repositorySnapshot != null ? repositorySnapshot.getBranchLayout() : null;
+    val gitMacheteBranch = repositorySnapshot != null ? repositorySnapshot.getManagedBranchByName(branchName) : null;
+
+    if (gitMacheteBranch != null) {
+      @UI Runnable traverseNextEntry = () -> {
```
PawelLipski, line 113: Hmm this it need to be `@UI`? and hence, same for `traverseNextEntry` params in `TraverseSyncToX` c'tors

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/traverse/TraverseSyncToRemote.java
```diff
@@ -0,0 +1,221 @@
...
+        break;
+
+      case AheadOfRemote :
+        if (!handleAheadOfRemote(gitMacheteBranch, localBranch)) {
+          return;
+        }
+        break;
+
+      case DivergedFromAndOlderThanRemote :
+        if (!handleDivergedFromRemote(gitMacheteBranch, syncToRemoteStatus, localBranch, /* withResetOption */ true)) {
```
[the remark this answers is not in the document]
PawelLipski: I.e. overwriting someone else's work in the remote repo etc.

frontend/actions/src/main/kotlin/com/virtuslab/gitmachete/frontend/actions/dialogs/DivergedFromRemoteDialog.kt
```diff
@@ -0,0 +1,104 @@
...
+        radioButton(
+          getString(
+            "action.GitMachete.BaseTraverseAction.dialog.diverged-from-remote.force-push-option"
+          ),
+          DivergeResolutionOption.FORCE_PUSH
+        ).commentCompat(getString("action.GitMachete.BaseTraverseAction.dialog.diverged-from-remote.force-push.comment"))
+      }
+
+      if (withResetOption) {
+        rowCompat {
```
PawelLipski, line 80: As discussed, TBD if these two options should ever be suggested at the same time

frontend/actions/src/main/kotlin/com/virtuslab/gitmachete/frontend/actions/dialogs/DivergedFromRemoteDialog.kt
```diff
@@ -0,0 +1,104 @@
...
+import com.virtuslab.gitmachete.backend.api.IManagedBranchSnapshot
+import com.virtuslab.gitmachete.backend.api.IRemoteTrackingBranchReference
+import com.virtuslab.gitmachete.backend.api.SyncToRemoteStatus
+import com.virtuslab.gitmachete.frontend.actions.compat.buttonsGroupCompat
+import com.virtuslab.gitmachete.frontend.actions.compat.commentCompat
+import com.virtuslab.gitmachete.frontend.actions.compat.rowCompat
+import com.virtuslab.gitmachete.frontend.resourcebundles.GitMacheteBundle.fmt
+import com.virtuslab.gitmachete.frontend.resourcebundles.GitMacheteBundle.getString
+
+enum class DivergeResolutionOption {
```
PawelLipski, line 16: Nit:
```suggestion
enum class DivergenceResolutionOption {
```
here and in the field

frontend/actions/src/main/kotlin/com/virtuslab/gitmachete/frontend/actions/dialogs/DivergedFromRemoteDialog.kt
PawelLipski: And this also poses a question whether radio buttons are really needed (maybe just stick to "regular" buttons at the bottom of the dialog?)

frontend/resourcebundles/src/main/resources/GitMacheteBundle.properties
```diff
@@ -460,3 +470,65 @@ action.GitMachete.CheckoutParentAction.undefined.branch-name=Branch (''{0}'') do
...
+action.GitMachete.BaseTraverseAction.dialog.cancel-traverse=Quit the interactive traverse
+
+action.GitMachete.BaseTraverseAction.dialog.diverged-from-parent.title=Resolve Divergence from Parent
```
PawelLipski, line 503: TBD what to do about the term "Divergence" here as it isn't 100%precise (unlike with Divergence from Remote, which is)

src/test/java/com/virtuslab/archunit/ClassStructureTestSuite.java
mkondratek: tested it again after the refactor from static to regular classes - still needed, and still a false positive imho

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/traverse/TraverseSyncToParent.java
```diff
@@ -0,0 +1,133 @@
...
+        getString("action.GitMachete.BaseTraverseAction.dialog.diverged-from-parent.title"),
+        getString(
+            "action.GitMachete.BaseTraverseAction.dialog.diverged-from-parent.text.HTML").fmt(
+                gitMacheteBranch.getName(),
+                gitMacheteBranch.getParent().getName()))
+        .cancelText(getString("action.GitMachete.BaseTraverseAction.dialog.cancel-traverse"));
+
+    switch (rebaseDialog.show(project)) {
+      case MessageConstants.YES :
+        new RebaseOnParentBackgroundable(
```
PawelLipski, line 114: Whooops I'm getting such a small dialog:

image

PawelLipski: Although huh, it also happens when running rebase from outside of traverse hmmm

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/traverse/BaseTraverseAction.java
PawelLipski: Huh in general AFAICS one is able to quickly click e.g. Push or Rebase actions in time windows between the modal dialogs (i.e. when background tasks are running) 😅 although it's at best a tech debt item, not really something to be addressed right now

src/test/java/com/virtuslab/archunit/ClassStructureTestSuite.java
PawelLipski: I think I know how to fix it... gimme a second...

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/traverse/BaseTraverseAction.java
mkondratek: for me:
`assert smth != null` indicates that we can be sure that the value is not null 
`if (smth == null) return` indicates that smth can be null and we simply end at this point in the method

I would rather leave the assert here - from the code reader POV assert makes me thinking "why" and then I go to the point and see why. on the other hand seeing `if smth  == null return` I do not think much - assuming thats possible and expected in some cases

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/traverse/TraverseSyncToParent.java
mkondratek: :yes:

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/traverse/TraverseSyncToParent.java
mkondratek: good point

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/traverse/TraverseSyncToRemote.java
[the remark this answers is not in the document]
mkondratek: and that's your decision (iiuc)! https://github.com/VirtusLab/git-machete-intellij-plugin/issues/246#issuecomment-1335216001

src/test/java/com/virtuslab/archunit/ClassStructureTestSuite.java
PawelLipski: Okay just apply the following:

```diff
diff --git a/src/test/java/com/virtuslab/archunit/ClassStructureTestSuite.java b/src/test/java/com/virtuslab/archunit/ClassStructureTestSuite.java
index 053f1d08..216769d4 100644
--- a/src/test/java/com/virtuslab/archunit/ClassStructureTestSuite.java
+++ b/src/test/java/com/virtuslab/archunit/ClassStructureTestSuite.java
@@ -1,5 +1,6 @@
 package com.virtuslab.archunit;

+import static com.tngtech.archunit.core.domain.JavaModifier.SYNTHETIC;
 import static com.tngtech.archunit.lang.syntax.ArchRuleDefinition.classes;
 import static com.tngtech.archunit.lang.syntax.ArchRuleDefinition.noClasses;

@@ -98,16 +99,14 @@ public class ClassStructureTestSuite extends BaseArchUnitTestSuite {
   public void all_classes_should_be_referenced() {
     val classesReferencedFromPluginXmlAttributes = extractAllClassesReferencedFromPluginXmlAttributes().toArray(Class[]::new);
     classes()
-        .that().resideOutsideOfPackages(
+        .that().doNotHaveModifier(SYNTHETIC)
+        .and().resideOutsideOfPackages(
             // Classes in *.impl.* packages may be instantiated via RuntimeBinding
             "..impl..",
             // For some reason, ArchUnit (com.tngtech.archunit.core.domain.JavaClass.getAccessesFromSelf)
             // doesn't see accesses to static fields
             "com.virtuslab.gitmachete.frontend.defs")
         .and().doNotBelongToAnyOf(classesReferencedFromPluginXmlAttributes)
-        .and()
-        .doNotBelongToAnyOf(com.virtuslab.gitmachete.frontend.actions.traverse.TraverseSyncToParent.class,
-            com.virtuslab.gitmachete.frontend.actions.traverse.TraverseSyncToRemote.class)
         // SubtypingBottom is processed by CheckerFramework based on its annotations
         .and().doNotHaveFullyQualifiedName(com.virtuslab.qual.internal.SubtypingBottom.class.getName())
         .should(new BeReferencedFromOutsideItself())
```

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/traverse/TraverseSyncToParent.java
mkondratek: happens to me sometimes with different dialogs. I think it's  some issue with IntelliJ itself

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/traverse/BaseTraverseAction.java
mkondratek: image

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/traverse/BaseTraverseAction.java
PawelLipski: Okay... but at the **beginning** of the method, it's perfectly expected that `repositorySnapshot` might be null, right? 🤔 if we return in such case, then all of:

1. `val branchLayout = repositorySnapshot != null ? repositorySnapshot.getBranchLayout() : null;`
2. `val gitMacheteBranch = repositorySnapshot != null ? repositorySnapshot.getManagedBranchByName(branchName) : null;`
3. `assert repositorySnapshot != null : "repositorySnapshot is null";`

can be simplified (1, 2) or straight removed (3)

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/traverse/TraverseSyncToRemote.java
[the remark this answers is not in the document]
PawelLipski: :fuggg: sorry, I was drunk or whatnot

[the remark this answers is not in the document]
PawelLipski: Or I didn't think about the _diverged and older -> reset and force push allowed_ part close enough ;p

mkondratek:
I can confirm the same for my macOS with intel

src/test/java/com/virtuslab/archunit/ClassStructureTestSuite.java
mkondratek: ok

src/main/resources/META-INF/plugin.xml
mkondratek: on the other hand the discover seems to be very useful thing for newbies 🤔 maybe a switch of places b/w the traverse and the discover?

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/traverse/BaseTraverseAction.java
mkondratek: oh, ok you are right!

src/main/resources/META-INF/plugin.xml
PawelLipski: Hmm okay let's start with switching places 🤔

frontend/actions/src/main/kotlin/com/virtuslab/gitmachete/frontend/actions/dialogs/DivergedFromRemoteDialog.kt
```diff
@@ -63,42 +62,43 @@ class DivergedFromRemoteDialog(
...
-            "action.GitMachete.BaseTraverseAction.dialog.diverged-from-remote.force-push-option"
-          ),
-          DivergeResolutionOption.FORCE_PUSH
-        ).commentCompat(getString("action.GitMachete.BaseTraverseAction.dialog.diverged-from-remote.force-push.comment"))
-      }
-
-      if (withResetOption) {
-        rowCompat {
+      when (relationToRemote) {
+        SyncToRemoteStatus.DivergedFromAndOlderThanRemote -> rowCompat {
```
PawelLipski, line 69: Nit: newline before `rowCompat` for consistency with the case below (unless forced this way by spotless)

frontend/resourcebundles/src/main/resources/GitMacheteBundle.properties
```diff
@@ -222,7 +222,7 @@ action.GitMachete.SyncSelectedToParentByMergeAction.warning-dialog.message.HTML=
 
 
 action.GitMachete.TraverseFromSelectedAction.text=Git Machete: Traverse from Selected Branch\u2026
-action.GitMachete.TraverseFromSelectedAction.GitMacheteContextMenu.text=T_raverse from Selected Branch\u2026
+action.GitMachete.TraverseFromSelectedAction.GitMacheteContextMenu.text=_Traverse from here\u2026
```
PawelLipski, line 225: Uppercase `Here`? dunno

mkondratek: no!

mkondratek: oh, ok! that's how it looks like in git4idea

mkondratek: capitalised `Here` 👍 but regarding to Slide In / Slide Out I would leave it as it is b/c it's like the preposition is a part of the action name in that case

frontend/resourcebundles/src/main/resources/GitMacheteBundle.properties
```diff
@@ -460,3 +470,65 @@ action.GitMachete.CheckoutParentAction.undefined.branch-name=Branch (''{0}'') do
...
+# Traverse
+
+action.GitMachete.BaseTraverseAction.description=Traverse
+action.GitMachete.BaseTraverseAction.description.empty-layout=Cannot traverse an empty branch layout
+
+action.GitMachete.BaseTraverseAction.dialog.traverse-approval.title=Traverse Info
+action.GitMachete.BaseTraverseAction.dialog.traverse-approval.text.HTML=<html>\
+  This action triggers an interactive walk over the branches and suggests actions (pushes/pulls/rebases) \
+  necessary to achieve both child-to-parent and local-to-remote synchronization.<br><br>\
+  Do you want to proceed?\
```
PawelLipski, line 483: I'd seriously consider adding a tip here about disabling Auto-reload (Settings/Preferences > Build, Execution, Deployment > Build Tools > Reload project after changes in the build scripts), or maybe even a link to that setting, if possible

Ofc it's a trade-off, coz one might want to have auto-reload enabled in general :goncern:

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/traverse/BaseTraverseAction.java
```diff
@@ -0,0 +1,128 @@
...
+            SHOW_TRAVERSE_INFO,
+            /* myHeight */ 100);
+
+        yesNoResult = traverseInfoDialog.showAndGet();
+      }
+
+      if (yesNoResult) {
+        val initialBranchName = getNameOfBranchUnderAction(anActionEvent);
+        if (initialBranchName != null) {
+          checkoutAndTraverseFrom(gitRepository, graphTable, initialBranchName);
```
PawelLipski, line 85: Btw in case of traversing from the current branch, the initial checkout is redundant... something to optimize, maybe

frontend/actions/src/main/kotlin/com/virtuslab/gitmachete/frontend/actions/dialogs/DivergedFromRemoteDialog.kt
```diff
@@ -0,0 +1,104 @@
...
+        DivergenceResolutionOption::class.java
+      )
+  }
+
+  private fun getRelationToRemoteDescription(): String =
+    when (relationToRemote) {
+      SyncToRemoteStatus.DivergedFromAndNewerThanRemote ->
+        getString("action.GitMachete.BaseTraverseAction.dialog.diverged-from-remote.older-than-remote")
+      SyncToRemoteStatus.DivergedFromAndOlderThanRemote ->
+        getString("action.GitMachete.BaseTraverseAction.dialog.diverged-from-remote.newer-than-remote")
```
PawelLipski, line 101: Whoops the messages are swapped 😅

mkondratek: actually, the message keys are swapped but the underlying values are correct for each case XD

frontend/resourcebundles/src/main/resources/GitMacheteBundle.properties
```diff
@@ -500,8 +500,8 @@ action.GitMachete.BaseTraverseAction.dialog.diverged-from-remote.reset-option=Re
 action.GitMachete.BaseTraverseAction.dialog.diverged-from-remote.reset-on-remote.comment=Warning: you will lose the local commits
 action.GitMachete.BaseTraverseAction.dialog.cancel-traverse=Quit the interactive traverse
 
-action.GitMachete.BaseTraverseAction.dialog.diverged-from-parent.title=Resolve Divergence from Parent
-action.GitMachete.BaseTraverseAction.dialog.diverged-from-parent.text.HTML=\
+action.GitMachete.BaseTraverseAction.dialog.unsynchronized-to-parent.title=Resolve Out Of Sync or In Sync but Fork-point Off to Parent
```
PawelLipski, line 503: Hmmm I'd use the specific title depending on the case (`Resolve Out of Sync to Parent`, `Resolve In Sync to Parent but Fork Point Off`)

mkondratek:
whoops, the rebase with conflicts seem to be an issue indeed 😞

mkondratek:
image

image

(the text is of the second one is fixed too)

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/traverse/TraverseSyncToRemote.java
```diff
@@ -0,0 +1,224 @@
...
+    }
+
+    val gitMacheteBranch =  repositorySnapshot.getManagedBranchByName(gitMacheteBranchOld.getName());
+    if (gitMacheteBranch == null) {
+      LOG.warn("gitMacheteBranch is null");
+      return;
+    }
+
+    val ongoingRepositoryOperationType = repositorySnapshot.getOngoingRepositoryOperation().getOperationType();
+    if (ongoingRepositoryOperationType != NO_OPERATION) {
```
mkondratek, line 69: In particular it's for the case when someone closes the dialog for conflict resolution during the rebase:
image

but actually we catch em all here! :surprised-pikachu:

mkondratek: fyi: i ve tested the happy path here (with the conflicts) - works like gold 🥇

frontend/resourcebundles/src/main/resources/GitMacheteBundle.properties
```diff
@@ -540,3 +540,9 @@ action.GitMachete.BaseTraverseAction.dialog.push-verification.untracked.text.HTM
     <p>The branch <b>{0}</b> is untracked.</p> \
     <p>Do you want to <b>push</b> the changes to the remote?</p>\
   </html>
+
+action.GitMachete.BaseTraverseAction.notification.ongoing-operation.title=Git Machete Traverse Error
+action.GitMachete.BaseTraverseAction.notification.ongoing-operation.message.HTML=<html>\
+      An ongoing repository operation <b>{0}</b> has been detected. \
+      The traverse stopped. It is suggested to resolve the ongoing operation and start over the traverse. \
```
PawelLipski, line 547: Just
```suggestion
      The traverse stopped. Resolve the ongoing operation and start over the traverse. \
```

frontend/resourcebundles/src/main/resources/GitMacheteBundle.properties
PawelLipski: @mkondratek TBD

frontend/resourcebundles/src/main/resources/GitMacheteBundle.properties
```diff
@@ -460,3 +470,79 @@ action.GitMachete.CheckoutParentAction.undefined.branch-name=Branch (''{0}'') do
...
+    <p>Please choose the preferred action:</p>\
+  </html>
+action.GitMachete.BaseTraverseAction.dialog.diverged-from-remote.title=Resolve Divergence from Remote
+action.GitMachete.BaseTraverseAction.dialog.diverged-from-remote.newer-than-remote=newer than
+action.GitMachete.BaseTraverseAction.dialog.diverged-from-remote.older-than-remote=older than
+
+action.GitMachete.BaseTraverseAction.dialog.diverged-from-remote.do-not-resolve-option=Do not sync
+action.GitMachete.BaseTraverseAction.dialog.diverged-from-remote.do-not-resolve-option.comment=Note: the remote and local branches will remain intact; traverse will continue
+action.GitMachete.BaseTraverseAction.dialog.diverged-from-remote.force-push-option=Force push
+action.GitMachete.BaseTraverseAction.dialog.diverged-from-remote.force-push.comment=Warning: you will lose the remote commits
```
PawelLipski, line 498: Maybe: _the remote commits will no longer be easily available (you might need to resort to reflogs to retrieve them)_? 🤔

mkondratek: tbh it sounds so nerdish that I would feel better simply seeing "lose" Xd

PawelLipski: Okay :trollface:

mkondratek: a person who knows that can _resort to reflogs to retrieve them_ will obviously understand that our plugin uses git under the hood (what else could it use?) and in git nothing is easily loosable

PawelLipski: Maybe... `might lose`? 🙏🏻 just **not** to make it too alarming

frontend/resourcebundles/src/main/resources/GitMacheteBundle.properties
```diff
@@ -460,3 +470,79 @@ action.GitMachete.CheckoutParentAction.undefined.branch-name=Branch (''{0}'') do
...
+
+action.GitMachete.BaseTraverseAction.dialog.fork-point-off.title=Resolve In Sync to Parent but Fork Point Off
+action.GitMachete.BaseTraverseAction.dialog.fork-point-off.text.HTML=\
+  <html>\
+    <p>The child branch <b>{0}</b> is in sync to the parent branch <b>{1}</b> \
+    but the range of commits belonging to <b>{0}</b> is uncertain (the fork point is off).</p><br>\
+    <p>Do you want to <b>rebase</b> branch <b>{0}</b> branch onto <b>{1}</b>?</p>\
+  </html>
+
+action.GitMachete.BaseTraverseAction.dialog.merged-to-parent.title=Resolve Merged to Parent
```
PawelLipski, line 518: Should there be a `Git Machete Traverse -` prefix in the title? As with `Git Machete Traverse - Pull` etc. below

frontend/resourcebundles/src/main/resources/GitMacheteBundle.properties
mkondratek: If I were a git newbie I would rather see `lose` than `might lose` b/c I will not know how to retrieve them

my thinking is like this
- who knows - knows that will not lose even though it says `lose`, and the message wont be alarming  for them
- who does not know - will lose them actually, and the message should be alarming for them

mkondratek: what's _the consistency to the cli argument_?

frontend/resourcebundles/src/main/resources/GitMacheteBundle.properties
mkondratek: good point 👍  to reflect that we are in the scope of the ongoing action

mkondratek: however I am afraid that it may be too long for some cases (e.g. `Git Machete Traverse - Resolve In Sync to Parent but Fork Point Off`)

mkondratek: image

oh, it works just fine ! cool

frontend/resourcebundles/src/main/resources/GitMacheteBundle.properties
mkondratek: image
 I see - simply name the git action and do not warn 😆  that's an option

mkondratek: no warn for the reset to...
image

mkondratek: wdyt about removing the warnings? we will meet halfway and the message will be correct

PawelLipski: Okay, let's remove them then!

PawelLipski (approved):
:benissimo:

PR #1381, opened 2022-12-04
Use `showStandardStreams` instead of modifying `events` in `testLogging`

PawelLipski:


mkondratek (approved):
So how does the logging output change?

PawelLipski:
> So how does the logging output change?

Not at all! The logic should be identical in all cases (both with `-PprintTestOutput` and without)

PR #1398, opened 2022-12-23
Apply minor fixes before the v3.3.0 release

PawelLipski:

PR #1399, opened 2022-12-24
Enforce checking `isEnabled`/`isVisible` in `onUpdate` before `setEnabled`/`setVisible`

PawelLipski:


frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/toolbar/FetchAllRemotesAction.java
```diff
@@ -32,8 +32,12 @@ public LambdaLogger log() {
   protected void onUpdate(AnActionEvent anActionEvent) {
     super.onUpdate(anActionEvent);
 
-    val project = getProject(anActionEvent);
     val presentation = anActionEvent.getPresentation();
+    if (!presentation.isEnabled()) {
```
mkondratek, line 36: I am not sure about it 🤔 maybe leave these changes to 3.3.1? 

Here I am thinking about a scenario when the action is visible, not enabled but we still want to provide some description (e.g. why the action is disabled) - we need to examine these changes carefully and see if the proper messages still appear when they should

mkondratek (changes requested):

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/toolbar/FetchAllRemotesAction.java
PawelLipski: Ok, 3.3.1 FTW

PawelLipski:
Cherry-picked the unquestionably useful parts to https://github.com/VirtusLab/git-machete-intellij-plugin/pull/1398. The rest is to be declined now, let's fix #1401 first and then maybe re-visit.

PR #1402, opened 2022-12-24
Move Discover action to the end of the toolbar

PawelLipski:
_Thots?..._ I'd really rather have more day-to-day actions visible, at the expense of Discover... and also I'd argue that if someone might have a worse UX due to weaker accessibility of Discover, then they'll have an even worse UX due to even weaker accessibility of Help (❓)

image

### DIFF
--- a/CHANGE-NOTES.md
+++ b/CHANGE-NOTES.md
@@ -1,8 +1,9 @@
 # Changelog
 
 ## v3.3.0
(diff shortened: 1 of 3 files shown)


mkondratek (approved):
> (...) then they'll have an even worse UX due to even weaker accessibility of Help (❓)

_zaorane_

mkondratek:
pls update change notes

PR #1406, opened 2022-12-27
Let ide-probe sift out more types of unnamed background tasks

PawelLipski:
See https://github.com/VirtusLab/ide-probe/pull/334

Before:

image

After:

image


PawelLipski:
Done on #1407

PR #1414, opened 2023-01-03
Ensure a fresh branch snapshot is seen by `TraverseSyncToParent#execute`

PawelLipski:


PawelLipski:
> Have you found any issues with the former solution? What's the reasoning?

Basically, the repository snapshot and branch snapshot passed to `TraverseSyncToParent` was stale, and hence traverse was making decisions based on outdated repository state (skipping sync to parent for the next branch even though it was now needed etc.) :/

mkondratek (approved):
ok then

Issue #1415, opened 2023-01-04
Race condition around Reset to Remote and Pull actions in Traverse


PawelLipski:
### Reset to Remote

```java
    switch (selectedAction) {
// ...
      case RESET_TO_REMOTE :
        new ResetCurrentToRemoteBackgroundable(
            getString("action.GitMachete.BaseResetToRemoteAction.task-title"),
            /* canBeCancelled */ true, gitMacheteBranch.getName(), remoteTrackingBranch.getName(),
            gitRepository).queue();
        return true;
// ...
```
`ResetCurrentToRemoteBackgroundable` runs simultaneously in background with the checkout of next branch :///

Instead, `onSuccess` should be implemented.

### Pull

```java
    switch (pullApprovalDialogBuilder.show(project)) {
      case MessageConstants.YES :
        val mergeProps = new MergeProps(
            /* movingBranchName */ gitMacheteBranch,
            /* stayingBranchName */ remoteTrackingBranch);

        val isUpToDate = FetchUpToDateTimeoutStatus.isUpToDate(gitRepository);
        val fetchNotificationPrefix = isUpToDate
            ? getNonHtmlString("action.GitMachete.BasePullAction.notification.prefix.no-fetch-perform")
                .fmt(FETCH_ALL_UP_TO_DATE_TIMEOUT_AS_STRING)
            : getNonHtmlString("action.GitMachete.BasePullAction.notification.prefix.fetch-perform");
        val fetchNotificationTextPrefix = fetchNotificationPrefix + (fetchNotificationPrefix.isEmpty() ? "" : " ");
        FastForwardMerge.createBackgroundable(gitRepository, mergeProps, fetchNotificationTextPrefix).queue(); /// WHOOPS
        return true;
// ...
```

`FastForwardMerge`'s backgroundable runs simultaneously in background with the checkout of next branch :///

Instead, `onSuccess` should be implemented.

PR #1416, opened 2023-01-04
Rework `FastForwardMerge#createBackgroundable` into `FastForwardMergeBackgroundable`

PawelLipski:
Needed as groundwork for #1415


mkondratek (approved):
good point. we are more consistent now in the scope of backgroundables. idk why I had not done that in the frist place

PR #1417, opened 2023-01-04
Mitigate race conditions around enqueuing backgroundables + fix race conditions in traverse

PawelLipski:

PR #1420, opened 2023-01-04
In Pull action, resort to remote's default refspec instead of an explicit `+refs/heads/*:refs/remotes/<remote>/*`

PawelLipski:


mkondratek (approved):
oh. I can imagine this is useful 😆

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/backgroundables/FetchBackgroundable.java
```diff
@@ -10,6 +10,7 @@
 import lombok.CustomLog;
```
mkondratek, line 10: pls update change notes too

mkondratek: I think at least those should be mentioned
- [ ] the new approach to pull (this one)
- [ ] race condition fix ("improved stability of plugin/traverse", #1417)
- [ ] "improved user facing messages" (#1413)

optionally 
- [ ] skip no-op checkouts (#1404) but since the traverse has not been released yet probably that one could be skipped

PawelLipski: > race condition fix ("improved stability of plugin/traverse")

The only place where a known bug has actually been fixed is in traverse; since traverse will be first released only now, it doesn't make sense to mention the bugs fixed in the process

> skip no-op checkouts

Same here, it's an integral part of the traverse feature

PR #1425, opened 2023-01-07
Properly handle the when-ready callback in `SlideOutBackgroundable`

PawelLipski:


frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/traverse/TraverseSyncToParent.java
```diff
@@ -119,15 +119,11 @@ private boolean handleMergedToParent(
...
+        // For a branch merged to its parent, we're not syncing to remote.
+        // Let's just go straight to the next branch.
+        Runnable doInUIThreadWhenReady = () -> graphTable.queueRepositoryUpdateAndModelRefresh(traverseNextEntry);
         new SlideOutBackgroundable(getString("action.GitMachete.BaseSlideOutAction.task.title"),
-            managedBranch, gitRepository, currentBranchIfManaged, branchLayout, graphTable) {
-          @Override
-          public void onSuccess() {
-            // For a branch merged to its parent, we're not syncing to remote.
-            // Let's just go straight to the next branch.
-            graphTable.queueRepositoryUpdateAndModelRefresh(traverseNextEntry);
```
mkondratek, line 128: our gain here (probably not only one) is that `queueRepositoryUpdateAndModelRefresh` is not called again right after the call from `SlideOutBackgroundable`, am I right?

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/backgroundables/SlideOutBackgroundable.java
```diff
@@ -87,10 +91,15 @@ public void run(ProgressIndicator indicator) {
...
     }
   }
 
+  /** Do not override {@code onSuccess}. Instead, pass {@code doInUIThreadWhenReady} callback to the constructor. */
+  @Override
+  public final void onSuccess() {}
+
   @ContinuesInBackground
   @UIEffect
   private void suggestBranchDeletion(String branchName) {
```
mkondratek, line 105: I am wondering... 🤔  btw you could extract a class from the `Task.Backgroundable` below similarly with a callback cstor arg, and pass the `doInUIThreadWhenReady` at the level of `SlideOutBackgroundable::run`

I just got a little bit confused seeing both `onSuccess` overrides (one saying "Do not override...") from the github changes level POV 😆 

I think that with my suggestion applied it would be more clearer that we do not call `doInUIThreadWhenReady` twice in the same if-else path (which was not so obvious to me at the first glance)

PawelLipski: Huh I thought a bit about it now... the scope of what this anonymous `new Task.Backgroundable` does is a bit unfocused (slide out branch + maybe delete branch + maybe set git config)... so I'd rather **not** extract a class for that 😅 and I can't see an elegant alternative factoring.

Still, lemme pass `doInUIThreadWhenReady` as an argument to `suggestBranchDeletion`... which is somewhat redundant, but at least gives more clarity that the callback is executed at the end of each execution path 👣

mkondratek (approved):

PR #1430, opened 2023-01-08
Rework `TraverseSyncTo...#handle...` methods to return void instead of boolean

PawelLipski:


frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/traverse/TraverseSyncToParent.java
```diff
@@ -26,8 +27,9 @@
 import com.virtuslab.gitmachete.frontend.vfsutils.GitVfsUtils;
 import com.virtuslab.qual.async.ContinuesInBackground;
 
-@ExtensionMethod({GitMacheteBundle.class, GitVfsUtils.class})
 @CustomLog
+@ExtensionMethod({GitMacheteBundle.class, GitVfsUtils.class})
+@SuppressWarnings("MissingSwitchDefault")
```
mkondratek, line 32: what happens when the value does not match any of the declared cases?

mkondratek (approved):
_widziałem to przychodzące_

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/traverse/TraverseSyncToParent.java
PawelLipski: Just nothing... traverse stops (as we've fallen into `com.intellij.openapi.ui.MessageConstants#CANCEL`)

mkondratek (approved):
ok!

PR #1434, opened 2023-01-10
Fix another race condition in slide-out action (branch layout saved concurrently with branch deletion & refresh)

PawelLipski:


mkondratek (approved):
this is fine

Issue #1436, opened 2023-01-11
NPE in `OverrideForkPointDialog` when fork point cannot be found for a branch


PawelLipski:
Git Machete plugin version: 3.3.0-61-SNAPSHOT+git.5c44d6d
IDE: IntelliJ IDEA 2022.2.4
Operating system: Mac OS X 13.1

```
java.lang.NullPointerException: null cannot be cast to non-null type com.virtuslab.gitmachete.backend.api.ICommitOfManagedBranch
	at+com.virtuslab.gitmachete.frontend.actions.dialogs.OverrideForkPointDialog$createCenterPanel$1$2$3$1.getListCellRendererComponent(OverrideForkPointDialog.kt:129)
	at+com.intellij.openapi.ui.ComboBoxWithWidePopup$AdjustingListCellRenderer.getListCellRendererComponent(ComboBoxWithWidePopup.java:85)
	at+com.intellij.ide.ui.laf.darcula.ui.DarculaComboBoxUI.paintCurrentValue(DarculaComboBoxUI.java:306)
	at+com.intellij.ide.ui.laf.darcula.ui.DarculaComboBoxUI.paint(DarculaComboBoxUI.java:270)
	at+java.desktop/javax.swing.plaf.ComponentUI.update(ComponentUI.java:161)
	at+java.desktop/javax.swing.JComponent.paintComponent(JComponent.java:845)
	at+java.desktop/javax.swing.JComponent.paint(JComponent.java:1122)
	at+com.intellij.openapi.ui.ComboBox.paint(ComboBox.java:288)
	at+java.desktop/javax.swing.JComponent.paintChildren(JComponent.java:955)
	at+java.desktop/javax.swing.JComponent.paint(JComponent.java:1131)
	at+java.desktop/javax.swing.JComponent.paintChildren(JComponent.java:955)
	at+java.desktop/javax.swing.JComponent.paint(JComponent.java:1131)
	at+java.desktop/javax.swing.JComponent.paintChildren(JComponent.java:955)
	at+java.desktop/javax.swing.JComponent.paint(JComponent.java:1131)
	at+java.desktop/javax.swing.JComponent.paintChildren(JComponent.java:955)
	at+java.desktop/javax.swing.JComponent.paint(JComponent.java:1131)
	at+java.desktop/javax.swing.JLayeredPane.paint(JLayeredPane.java:586)
[...]

PR #1438, opened 2023-01-11
Prohibit unsafe `as` cast in Kotlin in favor of `as?`

PawelLipski:


frontend/actions/src/main/kotlin/com/virtuslab/gitmachete/frontend/actions/dialogs/OverrideForkPointDialog.kt
```diff
@@ -126,8 +126,10 @@ class OverrideForkPointDialog(
               isSelected: Boolean,
               cellHasFocus: Boolean
             ): Component {
-              val commit: ICommitOfManagedBranch = value as ICommitOfManagedBranch
```
PawelLipski, line 129: Here's the actual fix for #1436

mkondratek: how does the dialog look like when the commit is null?

PawelLipski: Huh good point

image

So the entire `Inferred commit` item should be hidden then... lemme fix

PawelLipski: And `commit` param can probably only be null when `val list` is empty, lemme check... 🤔

PawelLipski: Ha! nope, `commit` will be `null` many times, as `getListCellRendererComponent` for some reasons gets invoked for `index = -1` and `value = null`

PawelLipski: Okay, all related issues fixed

frontend/actions/src/main/kotlin/com/virtuslab/gitmachete/frontend/actions/dialogs/OverrideForkPointDialog.kt
```diff
@@ -89,51 +89,54 @@ class OverrideForkPointDialog(
...
-          fmt(
+      val commits = branch.commitsUntilParent + listOf(branch.forkPoint, parent.pointedCommit)
+      val comboItems = commits.filterNotNull().distinct().reversed()
+      if (comboItems.isNotEmpty()) {
+        rowCompat {
+          val rb: Cell<JBRadioButton> = radioButton(
             getString(
               "action.GitMachete.BaseOverrideForkPointAction.dialog.override-fork-point.radio-button.custom"
             ),
-            branch.forkPoint?.shortHash ?: "cannot resolve commit hash"
```
PawelLipski, line 111: `fmt` was redundant here (`action.GitMachete.BaseOverrideForkPointAction.dialog.override-fork-point.radio-button.custom=Manually picked commit:`, no `{...}`s)

mkondratek (approved):

PR #1439, opened 2023-01-11
Enforce a hard 8KB limit on the length of GitHub URL for reporting issues

PawelLipski:


frontend/ui/impl/src/main/resources/bug_report.md
```diff
@@ -3,31 +3,15 @@
...
-## IDEA - stack trace
-%stacktraces%
-
-## IDEA - additional info
-%additionalInfo%
-
 ## Bug description
 **Please include steps to reproduce (like `go to...`/`click on...` etc.) + expected and actual behaviour**.
 
-## (optional) Logs
```
PawelLipski, line 15: It's pretty unlikely that anybody will really include these... esp. that you need to collect them BEFORE the issue occurs 😅

mkondratek: I agree but - we do have this description somewhere in developlment.md, dont we? It may be useless in this scope but still handy for devs

mkondratek (approved):

frontend/ui/impl/src/main/resources/bug_report.md
PawelLipski: image

PR #1451, opened 2023-01-13
Ensure all shell scripts pass shellcheck

PawelLipski:


scripts/sandbox-setup-utils.sh
```diff
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
+# shellcheck disable=SC2154
```
PawelLipski, line 2: `$sandboxDir` coming from the calling script

PawelLipski: (seemingly undefined variable)

testCommon/src/testFixtures/resources/setup-with-single-remote.sh
```diff
@@ -52,6 +52,7 @@ cd machete-sandbox
     commit HOTFIX Add the trigger - fixes
     push
     git reset --keep HEAD~1
+    # shellcheck disable=SC2154
```
PawelLipski, line 55: `commit_day_of_month` coming from the outside

.circleci/config.yml
```diff
@@ -3,7 +3,7 @@ version: 2.1
 executors:
   docker_executor:
     docker:
-      - image: 'gitmachete/intellij-plugin-ci:9.1.0'
+      - image: 'gitmachete/intellij-plugin-ci:9.2.0'
```
PawelLipski, line 6: To be pushed soon

scripts/enforce-version-bump
```diff
@@ -10,6 +10,6 @@ last_tag=$(git describe --tags --abbrev=0)
 last_tag_version=${last_tag#v}
 current_version=$(extract_version_from_current_wd)
 
-if [[ $(printf "$current_version\n$last_tag_version" | sort --version-sort | tail -n 1) == "$last_tag_version" ]]; then
```
mkondratek, line 13: Out of curiosity, what's the problem with printf?

scripts/enforce-change-notes-updated-on-pr-to-master
```diff
@@ -5,8 +5,6 @@ set -e -o pipefail -u
 self_dir=$(cd "$(dirname "$0")" &>/dev/null; pwd -P)
 source "$self_dir"/utils.sh
 
-change_notes_path=CHANGE-NOTES.md
```
mkondratek, line 8: Unused, right?

scripts/verify-artifact-contents
```diff
@@ -35,19 +35,19 @@ done
...
 # as this might lead to a linker error (for api) or an SLF4J error (for impl) in the runtime.
 forbidden_library_prefixes=(checker ide-probe idea junit-driver kotlin lombok remote-robot scala slf4j)
-for lib in ${forbidden_library_prefixes[@]}; do
+for lib in "${forbidden_library_prefixes[@]}"; do
   if matching_jars=$(grep "^$lib" <<< "$jars_in_plugin_zip"); then
     for jar in $matching_jars; do
       jar_base_name=${jar/-[0-9]*/}
       # Maybe the library matches the forbidden prefix but is still expected (e.g. slf4j-lambda-core)?
-      if ! [[ " ${expected_libraries[@]} " == *\ $jar_base_name\ * ]]; then
+      if ! [[ " ${expected_libraries[*]} " == *\ $jar_base_name\ * ]]; then
```
mkondratek, line 50: What's the difference?

scripts/validate-icons
```diff
@@ -19,10 +19,9 @@ for referenced_file in $(git grep -Pho '(?<=loadIcon\(").*(?="\))' -- 'frontend/
     die "Dark icon '$referenced_file' referenced from MacheteIcons.java is missing from resources!"
   fi
 
-  files=($referenced_file_path $referenced_file_path_dark)
-  for file in $files; do
-    width=$(git grep -Pho '(?<=width=")[0-9]*(?=")' -- $file | head -1)
-    height=$(git grep -Pho '(?<=height=")[0-9]*(?=")' -- $file | head -1)
+  for file in "$referenced_file_path" "$referenced_file_path_dark"; do
```
mkondratek, line 22: What was wrong here?

mkondratek (approved):

scripts/enforce-version-bump
PawelLipski: The same problem is with C `printf`, or even Java `String.format`: non-constant values should **not** occur in the format string, as they might include `%...` format characters (which in this case, won't be printed at all, as there're no further params). Not a serious issue here, but shellcheck is right in general

scripts/enforce-change-notes-updated-on-pr-to-master
PawelLipski: Yeah, some leftover

scripts/verify-artifact-contents
PawelLipski: https://github.com/koalaman/shellcheck/wiki/SC2199. Here it doesn't matter, we do want an implicit concatenation anyway.

In general:
`$@`, `${my_array[@]}`: each element is a separate string
`$*`, `${my_array[*]}`: all elements concatenated into a single string

Showcase:

```shell

array=(foo bar qux)

function print_args() {
  echo "$1:"
  shift
  for arg in "$@"; do
    echo " - $arg"
  done
  echo
}

print_args '${array[@]}' "${array[@]}"
print_args '${array[*]}' "${array[*]}"

function call_print_args() {
  print_args '$@' "$@"
  print_args '$*' "$*"
}

call_print_args foo bar qux
```

Output:
```
${array[@]}:
 - foo
 - bar
 - qux

${array[*]}:
 - foo bar qux

$@:
 - foo
 - bar
 - qux

$*:
 - foo bar qux
```

scripts/validate-icons
PawelLipski: https://github.com/koalaman/shellcheck/wiki/SC2128:

image

PawelLipski: I.e. `referenced_file_path_dark` wasn't even checked 😅

PawelLipski:
In a nutshell... shell script is a very poorly designed language, but with strict mode and shellcheck it's a bit more bearable 💩 Still the best choice for when a lot of external programs need to be called/their outputs combined etc.

PR #1453, opened 2023-01-14
Restore hooks outputs in git-machete outputs pre-recorded for tests

PawelLipski:


testCommon/src/testFixtures/resources/common.sh
```diff
@@ -37,7 +37,7 @@ function create_repo() {
   cd $dir
   shift
   git init "$@"
-  if [ "$(expr substr $(uname -s) 1 10)" != "MINGW64_NT" && "$(expr substr $(uname -s) 1 10)" != "MINGW32_NT" ]; then
```
PawelLipski, line 40: This was the culprit... the expression always evaluated to false, coz `expr substr` is some weird stuff that doesn't always work, see [SC2308](https://github.com/koalaman/shellcheck/wiki/SC2308#substr)

mkondratek (approved):

PR #1466, opened 2023-01-20
Enforce that `Task.Backgroundable#run` is always `@UIThreadUnsafe` + get rid of unneeded background tasks

PawelLipski:
image

### DIFF
--- a/frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/contextmenu/ShowSelectedInGitLogAction.java
+++ b/frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/contextmenu/ShowSelectedInGitLogAction.java
@@ -1,14 +1,8 @@
 package com.virtuslab.gitmachete.frontend.actions.contextmenu;
 
 import static com.virtuslab.gitmachete.frontend.resourcebundles.GitMacheteBundle.getNonHtmlString;
-import static com.virtuslab.gitmachete.frontend.resourcebundles.GitMacheteBundle.getString;
-
-import java.util.concurrent.CancellationException;
-import java.util.concurrent.ExecutionException;
 
 import com.intellij.openapi.actionSystem.AnActionEvent;
-import com.intellij.openapi.progress.ProgressIndicator;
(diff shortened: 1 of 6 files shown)


frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/contextmenu/ShowSelectedInGitLogAction.java
```diff
@@ -89,25 +83,12 @@ public void actionPerformed(AnActionEvent anActionEvent) {
...
-          LOG.error(msg);
-        }
-      }
-    }.queue();
+    val logUi = VcsProjectLog.getInstance(project).getMainLogUi();
+    if (logUi == null) {
+      LOG.error("Main VCS Log UI is null");
+      return;
+    }
+    logUi.getVcsLog().jumpToReference(hash.asString());
```
PawelLipski, line 91: This returns a Java `Future`. Still, we're free to ignore (fire-and-forget) that Future, just as we effectively ignore most `Task.Backgroundable`s after `queue()`.

mkondratek: no exception handling needed here? 🤔

mkondratek (approved):

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/contextmenu/ShowSelectedInGitLogAction.java
PawelLipski: I don't think it's needed TBH. It's pretty hard for exception to occur here (I've checked, and passing `lolxd` results in just a tooltip; passing `null` crashes the flow before the `Future` starts). Even if it happens... it's gonna most likely be reported to the user (who hopefully reports it to us, in turn)

mkondratek: ok,thx

PR #1477, opened 2023-01-23
Permanently disable `buildSearchableOptions` Gradle task

PawelLipski:


buildSrc/src/main/kotlin/com/virtuslab/gitmachete/buildsrc/IntellijPlugin.kt
```diff
@@ -29,10 +29,8 @@ fun Project.configureIntellijPlugin() {
     plugins.set(listOf("Git4Idea")) // Needed solely for ArchUnit
   }
 
-  // TODO (#1470): restore building searchable options in CI if there is any benefit to the users
-  // The output of this task is for some reason very poorly cached,
-  // and the task takes a significant amount of time,
-  // while the index of searchable options is of little use for local development.
+  // It only affects searchability of plugin-specific settings (which we don't provide so far).
```
mkondratek, line 32: So this may be related to https://github.com/VirtusLab/git-machete-intellij-plugin/issues/270, right? I would rather leave the issue link here.

mkondratek (approved):

buildSrc/src/main/kotlin/com/virtuslab/gitmachete/buildsrc/IntellijPlugin.kt
PawelLipski: +1, added a TODO

mkondratek (approved):

PR #1482, opened 2023-01-23
Fix open machete file action in the case missing machete file

mkondratek:
Steps to reproduce:

1. have git machete tab closed 
2. remove machete file
3. call `open machete file` from the find action dialog

result:
error - machete file not found
notification - branch layout discovered

expected: 
tbd

### DIFF
--- a/frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/toolbar/OpenMacheteFileAction.java
+++ b/frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/toolbar/OpenMacheteFileAction.java
@@ -66,6 +66,10 @@ public class OpenMacheteFileAction extends BaseProjectDependentAction {
           /* message */ getString("action.GitMachete.OpenMacheteFileAction.notification.message.machete-file-not-found")
               .fmt(gitDir.getPath()),
           NotificationType.ERROR);
+      // Note there is no `.expire()` call, so the action remains available as long as the notification.
+      // It is troublesome to track the notification and cover all cases when it shall be expired.
+      // However, the discover action does not any changes directly; a dialog appears with options to accept or cancel.
+      // Hence, there is no much harm in leaving this notification without the expiration.
       errorWithDiscover.addAction(NotificationAction
           .createSimple(getString("action.GitMachete.DiscoverAction.GitMacheteToolbar.text"),
               () -> ActionManager.getInstance().getAction(DiscoverAction.class.getSimpleName())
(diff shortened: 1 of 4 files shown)


frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/toolbar/OpenMacheteFileAction.java
```diff
@@ -56,8 +56,6 @@ public void actionPerformed(AnActionEvent anActionEvent) {
             /* message */ getString("action.GitMachete.OpenMacheteFileAction.notification.title.cannot-open"))))
         .getOrNull();
 
-    getGraphTable(anActionEvent).queueRepositoryUpdateAndModelRefresh();
```
PawelLipski, line 59: Wow WTF was the purpose of this... can you dig out any clue from the original PR that added this line?

mkondratek: I searched the git history and looked at https://github.com/VirtusLab/git-machete-intellij-plugin/pull/272 but I cannot see anything that could explain that

PawelLipski (approved):

mkondratek:
I added the notification action so one can easily resolve the problem described in error:

image

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/toolbar/OpenMacheteFileAction.java
```diff
@@ -56,13 +60,18 @@ public void actionPerformed(AnActionEvent anActionEvent) {
...
-
     if (macheteFile == null) {
-      VcsNotifier.getInstance(project).notifyError(/* displayId */ null,
+      val errorWithDiscover = VcsNotifier.STANDARD_NOTIFICATION.createNotification(
           /* title */ getString("action.GitMachete.OpenMacheteFileAction.notification.title.machete-file-not-found"),
           /* message */ getString("action.GitMachete.OpenMacheteFileAction.notification.message.machete-file-not-found")
-              .fmt(gitDir.getPath()));
+              .fmt(gitDir.getPath()),
+          NotificationType.ERROR);
+      errorWithDiscover.addAction(NotificationAction
```
mkondratek, line 69: Note there is no `expire()`. It's hard to track the notification status and all the cases that could make it expired. On the other hand the action is simply a call to discover action - one see the suggested dialog first, must accept it. So even though the action will remain available as long as the notification lives there is no big deal imho. wdyt @PawelLipski ?

PawelLipski: Makes sense... maybe pls put the above musings into a code comment!

PawelLipski (approved):

PR #1304, opened 2022-10-26
Get rid of LOG.error based IDE errors for isBranchPresent calls

LukaszKontowski:
Closes #1298

Issue #1298, opened 2022-10-24
`GitCoreRepository.isBranchPresent` throws "Not a directory" exception for local branch


LukaszKontowski:
Reproduce steps:
1. Create a new local branch in repository `A` (it should not exist remotely - only locally)
2. Open `A` as a project in IntelliJ IDEA
3. Open Machete File
4. Boom! Ide error appears.

In my case - the branch's name was `docs/improve_reference_conf_comments`. This exception looks like a false positive. It should not be displayed as an error.

Stacktrace:
```
java.nio.file.FileSystemException: /Users/lkontowski/code/ide-probe/.git/refs/remotes/upstream/docs/improve_reference_conf_comments: Not a directory
	at java.base/sun.nio.fs.UnixException.translateToIOException(UnixException.java:100)
	at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:106)
	at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:111)
	at java.base/sun.nio.fs.UnixFileAttributeViews$Basic.readAttributes(UnixFileAttributeViews.java:55)
	at java.base/sun.nio.fs.UnixFileSystemProvider.readAttributes(UnixFileSystemProvider.java:148)
	at java.base/java.nio.file.Files.readAttributes(Files.java:1851)
	at org.eclipse.jgit.util.FileUtils.fileAttributes(FileUtils.java:804)
	at org.eclipse.jgit.util.FS.fileAttributes(FS.java:1268)
	at org.eclipse.jgit.internal.storage.file.FileSnapshot.<init>(FileSnapshot.java:234)
	at org.eclipse.jgit.internal.storage.file.FileSnapshot.<init>(FileSnapshot.java:213)
	at org.eclipse.jgit.internal.storage.file.FileSnapshot.save(FileSnapshot.java:104)
	at org.eclipse.jgit.internal.storage.file.RefDirectory.lambda$1(RefDirectory.java:1095)
	at org.eclipse.jgit.util.FileUtils.readWithRetries(FileUtils.java:722)
	at org.eclipse.jgit.internal.storage.file.RefDirectory.scanRef(RefDirectory.java:1094)
	at org.eclipse.jgit.internal.storage.file.RefDirectory.readRef(RefDirectory.java:1052)
	at org.eclipse.jgit.internal.storage.file.RefDirectory.readAndResolve(RefDirectory.java:263)
[...]

PR #1501, opened 2023-01-31
Add regression test for #1298 (unwanted `LOG.error(..., e)` calls from JGit)

PawelLipski:
See #1298, #1304

### DIFF
--- a/gitCore/jGit/src/test/java/com/virtuslab/gitcore/impl/jgit/integration/GitCoreRepositoryTest.java
+++ b/gitCore/jGit/src/test/java/com/virtuslab/gitcore/impl/jgit/integration/GitCoreRepositoryTest.java
@@ -1,25 +1,56 @@
 package com.virtuslab.gitcore.impl.jgit.integration;
 
 import static com.virtuslab.gitmachete.testcommon.SetupScripts.SETUP_WITH_SINGLE_REMOTE;
+import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.verify;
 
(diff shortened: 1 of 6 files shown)


gitCore/jGit/src/test/java/com/virtuslab/gitcore/impl/jgit/integration/GitCoreRepositoryTest.java
```diff
@@ -1,25 +1,56 @@
...
+  @Test
+  public void shouldNeverLeadToLogErrorCalledWithThrowable() {
+    assertTrue(gitCoreRepository.isBranchPresent("refs/heads/develop"));
+
+    Logger logger = mock(Logger.class);
+    LoggerMock.setMock(org.eclipse.jgit.internal.storage.file.FileSnapshot.class, logger);
+
+    assertFalse(gitCoreRepository.isBranchPresent("refs/heads/develop/something-else"));
+
+    verify(logger, never()).error(anyString(), any(Throwable.class));
```
mkondratek, line 54: so... when we add this mock slf4j it automagically prevents the ide errors?

mkondratek (approved):

gitCore/jGit/src/test/java/com/virtuslab/gitcore/impl/jgit/integration/GitCoreRepositoryTest.java
PawelLipski: Explained in a newly-added comment in moaaar detail:

```java
    // In test setup, a call to `LOG.error(String, Throwable)` doesn't crash the test
    // (and we aren't able to simply catch an exception to detect whether such a call took place).
    // In fact, with slf4j-simple (rather than slf4j-mock) on classpath, we'll just see stack trace printed out
    // (unless stderr is suppressed, which is the default when running tests under Gradle).
    // In IntelliJ, however, the situation is different, as IntelliJ provides an SLF4J implementation
    // which opens an error notification for each `LOG.error(String, Throwable)` (but not `LOG.error(String)`) call.
    // In this particular case, we want to avoid an `LOG.error(String, Throwable)` call in FileSnapshot c'tor
    // ending up in an user-visible, confusing error notification.
    // See the issue and PR #1304 for more details.
```

PR #1506, opened 2023-01-31
Lock repository refresh while Slide In is running

PawelLipski:


PawelLipski:
Meh not enough, the issue persists... lemme introduce a few other unrelated fixes on another PR (which might simplify handling of repository updates in general)

PR #1516, opened 2023-02-04
Accumulate failures in `scripts/run-pre-build-checks` instead of failing fast

PawelLipski:


scripts/run-pre-build-checks
```diff
@@ -1,43 +1,57 @@
...
+
+function _() {
+  script=$1
+  option=${2-}
+  if [[ $option == "--skip-in-pre-commit-hook" ]] && [[ $0 == *.git/hooks/pre-commit ]]; then
+    echo "(skipping $script as we are running in a git pre-commit hook)" >&2
+    return
+  fi
+  echo "> $script" >&2
+  if ! eval "./scripts/$script"; then
```
PawelLipski, line 20: `eval` is redundant, this would also work:
```suggestion
  if ! "./scripts/$script"; then
```
Leaving it just for clarity (if anything like clarity could ever be achieved in shell scripts :trollface:)

mkondratek (approved):
awesome ❤️

PR #1521, opened 2023-02-07
Re-enable AspectJ plugin + tidy up AspectJ handling in ArchUnit

PawelLipski:


buildSrc/build.gradle.kts
```diff
@@ -47,7 +47,6 @@ dependencies {
   // doesn't see jetbrains-annotations as an exceeded dependency and doesn't try to downgrade.
   // See https://github.com/littlerobots/version-catalog-update-plugin#exceeded-dependencies
   implementation(libs.jetbrains.annotations)
-  implementation(libs.pluginPackages.aspectj.postCompileWeaving)
```
PawelLipski, line 50: Tl;dr: it wasn't working since the migration to Kotlin DSL xD

buildSrc/src/main/kotlin/com/virtuslab/gitmachete/buildsrc/Dependencies.kt
```diff
@@ -119,9 +118,9 @@ fun Project.ideProbe() {
 }
 
 fun Project.jcabiAspects() {
```
mkondratek, line 120: can you remind me, why we need it?

mkondratek (approved):
ok

buildSrc/src/main/kotlin/com/virtuslab/gitmachete/buildsrc/Dependencies.kt
PawelLipski: Meh, `@Loggable` to include execution time in logs for methods... albeit after the 3+ years these entire logs seem mostly useless in general anyway :///

PawelLipski: We don't use them (at least I prefer more case-by-case printlns than the massive spam in logs), and users never attach them to the issues (and I don't remember a single case where having such logs would actually help)

mkondratek: that's true 😕

mkondratek (approved):

PR #1522, opened 2023-02-07
Split the version catalog into separate VCs for buildSrc and top-level project

PawelLipski:


backend/impl/src/main/java/com/virtuslab/gitmachete/backend/impl/hooks/StatusBranchHookExecutor.java
```diff
@@ -31,7 +31,7 @@
 
 @CustomLog
 public final class StatusBranchHookExecutor extends BaseHookExecutor {
-  private static final int EXECUTION_TIMEOUT_SECONDS = 2;
+  private static final int EXECUTION_TIMEOUT_SECONDS = 5;
```
PawelLipski, line 34: Local tests were still failing when CPU is under load... maybe it should be different for (local) tests and different for production (so that rendering the status doesn't hang too long) 🤔 OTOH, I have no indication that anybody outside of myself actually uses that hooks, in plugin or CLI :trollface:

buildSrc/settings.gradle.kts
```diff
@@ -1,7 +1,5 @@
 pluginManagement {
   repositories {
     gradlePluginPortal()
-    // For nl.littlerobots.version-catalog-update
```
PawelLipski, line 4: They finally published the plugin to Gradle Plugin Portal

.circleci/config.yml
```diff
@@ -66,12 +66,12 @@ jobs:
           keys:
             # First, try finding a cache entry with the same set of libraries
             # that also comes from the same branch (so as to make the best use of Gradle compilation cache).
-            - gradle-deps-v5-{{ checksum "gradle/libs.versions.toml" }}-{{ .Branch }}
+            - gradle-deps-v6-{{ checksum "gradle/libs.versions.toml" }}-{{ .Branch }}
```
PawelLipski, line 69: To start with clean slate, in order to check if all the Maven repo config is okay

settings.gradle.kts
```diff
@@ -10,11 +10,11 @@ include(
   "backend:api",
   "backend:impl",
   "frontend:base",
-  "frontend:actions",
-  "frontend:file",
   "frontend:resourcebundles",
+  "frontend:file",
```
PawelLipski, line 14: Just to keep topo-ish order

buildSrc/build.gradle.kts
```diff
@@ -1,21 +1,22 @@
...
-import se.ascp.gradle.GradleVersionsFilterPlugin
 
 plugins {
   `kotlin-dsl`
+  alias(libs.plugins.taskTree)
+  alias(libs.plugins.versionCatalogUpdate)
+  alias(libs.plugins.versionsFilter)
 }
 
-val kotlinLanguageVersion = "1.6"
```
PawelLipski, line 13: Just un-hardcode this value

buildSrc/build.gradle.kts
```diff
@@ -1,21 +1,22 @@
...
-
 // This is needed to use kotlin language version different from the default (1.4).
 // See https://handstandsam.com/2022/04/13/using-the-kotlin-dsl-gradle-plugin-forces-kotlin-1-4-compatibility/.
 afterEvaluate {
   tasks.withType<KotlinCompile>().configureEach {
+    val kotlinLanguageVersion = libs.pluginPackages.jetbrains.kotlin
```
PawelLipski, line 17: ...since it can be inferred from versions in the catalog

.circleci/config.yml
mkondratek: while we are here... in the most common scenario we need to update three lines here. can we have smth like `gradle-deps-v{{ version-number }}` instead? so only a single variable changes? 🤔

settings.gradle.kts
mkondratek: wow, actions module was a bit high :weed:

backend/impl/src/main/java/com/virtuslab/gitmachete/backend/impl/hooks/StatusBranchHookExecutor.java
mkondratek: pls share some hooks

gradle/libs.versions.toml
```diff
@@ -31,25 +30,13 @@ slf4j-mock = "org.simplify4u:slf4j-mock:2.3.0"
 #      (1.*.*, rather than a specific version like 1.7.36).
 slf4j-simple = "org.slf4j:slf4j-simple:1.7.36"
 vavr = "io.vavr:vavr:0.10.4"
-# Plugin packages
```
mkondratek, line 34: hmmm shouldn't this comment be at the top? 🤔 anyway pls provide comments here and in the other `libs.versions.toml` that describe what should be defined and where - now it's a little confusing to me 😅

mkondratek: image
 
unrel have you ever seen that? XD

mkondratek (changes requested):

.circleci/config.yml
PawelLipski: Yeah, circleci supports `<< ... >>` -style parameters... but it then looks even more confusing then now, I'd rather stick to the current looks 😅

gradle/libs.versions.toml
PawelLipski: > unrel have you ever seen that? XD

Yeah, a bug in github... it's probably coz this file used to be a symlink but has been replaced with a regular file. This is always a nasty corner case :/

backend/impl/src/main/java/com/virtuslab/gitmachete/backend/impl/hooks/StatusBranchHookExecutor.java
PawelLipski: _share some hooks_? 🤔 as in, provide hook samples in the project directory?

gradle/libs.versions.toml
PawelLipski: > hmmm shouldn't this comment be at the top? 

It's right before `pluginPackages-...` dependencies... a proper place!

PawelLipski: > anyway pls provide comments here and in the other libs.versions.toml that describe what should be defined and where

Lemme explain a few non-trivial points... which should be enough for anybody interested to decipher the whole picture! (so as to avoid explaining each and every dependency, as TBH almost each has a slightly different use case, esp. in buildSrc/ 😅)

mkondratek (approved):
lgtm

PR #1524, opened 2023-02-08
Use dependency helper functions consistently

PawelLipski:


branchLayout/impl/build.gradle.kts
```diff
@@ -8,5 +8,5 @@ dependencies {
 junit()
 lombok()
 slf4jLambdaApi()
-slf4jSimpleTest()
+slf4jSimple("test")
```
mkondratek, line 11: I see `test` here and `testFixtures` in the other place, and I assume there are some other options to... would it make sense to extract them to some enum?

build.gradle.kts
```diff
@@ -230,19 +227,21 @@ configureIntellijPlugin()
...
+archunit()
+// Checker is needed in root project runtime (not just compile-time) classpath for ArchUnit tests
+checkerQual("test")
 ideProbe()
+jgit("test")
 junit()
 junitPlatformLauncher()
-lombok()
-vavr()
-dependencies {
```
mkondratek, line 243: oh, it was a little mess indeed 😅

mkondratek (approved):

branchLayout/impl/build.gradle.kts
PawelLipski: I'd leave it as is, even Gradle Kotlin DSL uses strings to denote these scopes... if there's any typo, we'll get a build error immediately

PR #1525, opened 2023-02-09
Ensure no ancestor commit of parent branch is included into the branch's unique range of commits

PawelLipski:
Goes in tandem with https://github.com/VirtusLab/git-machete/pull/813

### DIFF
--- a/CHANGE-NOTES.md
+++ b/CHANGE-NOTES.md
@@ -1,7 +1,8 @@
 # Changelog
 
 ## v3.6.1
-- Fixed: notification about unmanaged current branch wasn't showing up
+- Fixed: notification about unmanaged current branch wasn't showing up.
+- Fixed: in case of red edge from a branch to its parent, the unique history of a branch can no longer include any commits reachable from its parent.
 
 ## v3.6.0
 - Improved the repository selection combobox so that modules are now searchable.
--- a/backend/impl/src/test/resources/setup-with-multiple-remotes.sh-discover.txt
+++ b/backend/impl/src/test/resources/setup-with-multiple-remotes.sh-discover.txt
@@ -1,21 +1,13 @@
   master  <2 files>
-  |
-  | HOTFIX Add the trigger (amended)
-  o-hotfix/add-trigger * (diverged from remote2)  <3 files>
 
   develop  <3 files>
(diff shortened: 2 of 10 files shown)


CHANGE-NOTES.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+## v3.6.1
+- Fixed: in case of red edge from branch to its parent, the unique history of a branch never includes commits reachable from its parent.
```
PawelLipski, line 4: Sometimes it happens (in case of missing/incomplete reflegs, esp. on someone else's branchs) that I toggle listing commits and see literally  **h u n d r e d s** of commits for a red-edge branch, all of which (with the exception of the few latest ones) are reachable for the parent. This doesn't make sense; let's "cut" the unique range at any commit reachable from the parent branch.

backend/impl/src/test/resources/setup-with-multiple-remotes.sh-discover.txt
```diff
@@ -1,21 +1,13 @@
...
-  |
-  | Call web service
-  | 1st round of fixes
-  | 2nd round of fixes
-  o-call-ws (ahead of remote1)  <6 files>
-    |
-    | Drop unneeded SQL constraints
-    x-drop-constraint (untracked)  <6 files>
+
+  allow-ownership-link (ahead of remote1)  <4 files>
```
PawelLipski, line 5: Side effect of removing reflogs in this script: `git machete discover` yields some dumb results. Anyway, I lost interest in improving `discover` long time ago anyway, that's a futile effort anyway :trollface:

backend/impl/src/test/resources/version.txt
```diff
@@ -1 +1 @@
-git-machete version 3.14.2
+git-machete version 3.15.1
```
PawelLipski, line 1: Not yet released (actually, I used the state from https://github.com/VirtusLab/git-machete/pull/813), but I don't think the outputs will have changed by the time of the actual release (soonish)

PawelLipski: Update: released already

mkondratek (approved):

PR #1532, opened 2023-02-10
Extract `call...MethodsThat` helpers for ArchUnit tests

PawelLipski:


src/test/java/com/virtuslab/archunit/UIThreadUnsafeMethodInvocationsTestSuite.java
```diff
@@ -205,60 +190,27 @@ public void check(JavaMethod method, ConditionEvents events) {
...
-            });
-          }
-        })
+        .should(callAnyMethodsThat("are known to be blocking IntelliJ APIs",
+            (method, calledMethod) -> calledMethod.getFullName().equals("com.intellij.dvcs.push.PushController.push(boolean)")))
         .check(productionClasses);
   }
 
   @Test
   public void only_ui_thread_unsafe_method_should_call_git4idea_or_io_methods() {
```
mkondratek, line 202: just to be clear...  the nature b/w this one and `only_ui_thread_unsafe_method_should_call_blocking_intellij_methods` is different, right? here we block all methods  and whitelist some methods while in the other place we block some methods only? hence, it would not make sense either to merge them or to split `only_ui_thread_unsafe_method_should_call_git4idea_or_io_methods`?

mkondratek (approved):
ok

src/test/java/com/virtuslab/archunit/UIThreadUnsafeMethodInvocationsTestSuite.java
PawelLipski: Yeah, exactly... while `only_ui_thread_unsafe_method_should_call_blocking_intellij_methods` is for checking calls to UI thread **unsafe methods** from otherwise **safe packages**... `only_ui_thread_unsafe_method_should_call_git4idea_or_io_methods` is for for checking calls to UI thread **safe methods** from otherwise **unsafe packages**.

PR #1533, opened 2023-02-10
Add `org.eclipse.jgit` to UI thread unsafe package prefixes

PawelLipski:


backend/impl/src/main/java/com/virtuslab/gitmachete/backend/impl/GitMacheteRepositoryCache.java
```diff
@@ -36,6 +37,7 @@ public GitMacheteRepositoryCache(IDependencyResolver dependencyResolver) {
   }
 
   @Override
+  @UIThreadUnsafe
```
PawelLipski, line 40: TBH only due to `HookExecutors`, which in turn check git config

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseSlideInBelowAction.java
```diff
@@ -185,33 +186,29 @@ public void onFinished() {
       return Tuple.of(branchName, () -> {});
     }
 
-    Runnable preSlideInRunnable = () -> {};
     val remoteBranch = getGitRemoteBranch(gitRepository, branchName);
 
-    if (options.shouldCheckout() && remoteBranch != null) {
```
PawelLipski, line 191: Just a refactor while I'm here

mkondratek (approved):
lgtm

PR #1568, opened 2023-02-18
Refine checks for asynchronous methods

PawelLipski:
:async-parrot:

No bugs detected in the process... _for now_

As usual, a prosthetics for a proper library ecosystem based on Futures ☹️ (`@ContinuesInBackground` methods would just return a `Future`). I appreciate Scala in retrospective for that reason...


mkondratek (approved):
lgtm

Issue #859, opened 2022-05-23
Drop support for 2022.1


PawelLipski:
As of May 23rd, 2022, it looks that from 2022.2 on, IntelliJs will run on JDK 17 🎉 

Dropping support for 2022.1 might then include switching to Java 17 in code, including:
* triple-quote (Scala/Groovy-style) strings
* switch expressions (no more vavr `Match`)
* records (no more lombok `@Data`)
* pattern matching for `instanceof`

and others, see https://dzone.com/articles/whats-new-between-java-11-and-java-17

KotlinIsland:
Why not rewrite the project in Kotlin 😳

PawelLipski:
Coz **b u d g e t** 😅

KotlinIsland:
Well Kotlin is designed to be very incrementally adoptable, and it already offers the three features mentioned in the OP (and many more). The project already has Kotlin set up, If you are going to write new code, why not write it in Kotlin 😃 .

PawelLipski:
Certain parts are already written in Kotlin indeed! Like certain dialogs under `frontend/actions/`

KotlinIsland:
Any thoughts on using [Kotlin Compose](https://blog.jetbrains.com/kotlin/2021/02/jetpack-compose-for-desktop-milestone-3-released/#idea-plugins) (and [here](https://github.com/JetBrains/compose-jb/tree/master/examples/intellij-plugin)) or the [Kotlin UI DSL](https://plugins.jetbrains.com/docs/intellij/kotlin-ui-dsl-version-2.html) or for the UI?

mkondratek:
I think it's an excellent idea to have more Kotlin in the codebase 🚀  Kotlin is better in many ways (mentioned and not mentioned here). 

In particular, https://github.com/VirtusLab/git-machete-intellij-plugin/issues/791 - using Kotlin DSL for that is the only option IMHO 👍 

@KotlinIsland if you want to help us with the adoption of Kotlin feel free to take some issues 🛠️  
We are open to contributions 🙂 you can start from that **slide in auto-complete thing**

PawelLipski:
One thing to check before migration to Java 17: https://github.com/VirtusLab/git-machete-intellij-plugin/issues/1584

PR #1574, opened 2023-02-19
Use Lombok's `@StandardException` to fill up c'tors in exception classes

PawelLipski:
See https://projectlombok.org/features/experimental/StandardException

### DIFF
--- a/backend/api/src/main/java/com/virtuslab/gitmachete/backend/api/GitMacheteException.java
+++ b/backend/api/src/main/java/com/virtuslab/gitmachete/backend/api/GitMacheteException.java
@@ -1,22 +1,10 @@
 package com.virtuslab.gitmachete.backend.api;
 
+import lombok.experimental.StandardException;
+
+@StandardException
+@SuppressWarnings("nullness:argument")
 public class GitMacheteException extends Exception {
-  public GitMacheteException() {
-    super();
-  }
-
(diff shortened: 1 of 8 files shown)


mkondratek:
Are we sure that we want to introduce another dependency on Lombok? esp as it is experimental? 🤔 

btw what's our status on the migration to java 17? I see https://github.com/VirtusLab/git-machete-intellij-plugin/issues/859 is a requirement.

PawelLipski:
> Are we sure that we want to introduce another dependency on Lombok? esp as it is experimental? 🤔

Not insisting. Although... I'm not sure if we're really gonna fully drop Lombok with migration to Java 17. In most classes, Lombok is used for stuff like just a single `@Getter`, or just `@RequiredArgsConstructor`, or `@SneakyThrows`. We can only ditch `@Data` really in favor of `record`s.

> btw what's our status on the migration to java 17? I see https://github.com/VirtusLab/git-machete-intellij-plugin/issues/859 is a requirement.

Yeah, dropping 2022.1 first... also, getting an intern who'll be willing to do all the migration 😅 although maybe I'll motivate myself to do it as well

mkondratek:
I am against the Lombok expansion (my gut's feeling) but I will not block it. 

Btw I could take care of the 2022.1 drop ✋

mkondratek (approved):

PR #1605, opened 2023-03-06
[to-be-considered] Allow to push branches in parallel

mkondratek:
It's easy to change it 😄 

It happens to me quite often that Push blocks my actions for a while.

The time required may be related to my internet connection (I use mobile network sometimes) and to the commit(s)/repo "size".

Eventually I cannot perform another push, rebase or whatever at once - I need to wait a few seconds. 

### Why is it required for Push to be blocking?

A quick answer I see is that even though multiple Push operations seem to be harmless one could trigger Rebase for `foo-branch` and the Push for the same branch while there is still an ongoing rebase... tbd what happens in such case


PawelLipski (approved):

mkondratek:
@PawelLipski wdyt?

PawelLipski:
Oh! +1, already approved 😅

PR #1607, opened 2023-03-07
Add `Yes + Quit` button to traverse confirmation dialogs

PawelLipski:


frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/dialogs/TraverseStepConfirmationDialog.java
```diff
@@ -21,14 +22,19 @@ public enum Result {
 
   @UIEffect
   public Result show(Project project) {
-    int result = MessageDialogBuilder.yesNoCancel(title, message)
-        .cancelText(getNonHtmlString("action.GitMachete.BaseTraverseAction.dialog.cancel-traverse"))
-        .show(project);
+    val yesText = getNonHtmlString("action.GitMachete.BaseTraverseAction.dialog.yes");
+    val yesAndQuitText = getNonHtmlString("action.GitMachete.BaseTraverseAction.dialog.yes-and-quit");
```
PawelLipski, line 26: Still no actual handling for that case (just behaves as plain `Quit` for now)

config/checker/com.intellij.astub
```diff
@@ -353,8 +353,13 @@ class MessageDialogBuilder<T extends MessageDialogBuilder> {
 }
 
 class MessageUtil {
-  int showOkCancelDialog(@Tainted String title, @Tainted String message, @Tainted String okText, @Tainted String cancelText,
-                           Icon icon, DoNotAskOption doNotAskOption, Project project);
+  int showOkCancelDialog(@Tainted String title, @Tainted String message, @Untainted String okText, @Untainted String cancelText,
```
mkondratek, line 356: while we are here...

I am wondering 🤔 It must happen that sometimes we introduce some stubs and then, after a while modify the code so the stub becomes redundant 🗑️ 

1. For now there is no mechanism to detect unused stubs, right?
2. What if we bump IntelliJ version and the stubbed method/class no longer exists? Are there any errors in such case (I do not remember any)? Or is it ignored?
3. (assuming 1. is true) Is it a problem the over time we will have more and more redundant stubs? E.g does it significantly affect the compilation/build time? If it is a problem then I might be worth to introduce some detection mechanism.

A bunch of questions 💐 I should have asked them years ago 😅

frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/dialogs/TraverseStepConfirmationDialog.java
mkondratek: fine 👍 

- [ ] **C H E C K B O X**

config/checker/com.intellij.astub
PawelLipski: >  It must happen that sometimes we introduce some stubs and then, after a while modify the code so the stub becomes redundant 🗑️

:yes:, it does definitely... still, given the effort that's sometimes needed to figure out these annotation, I'd rather keep these stubs around (even if they might be of little value for the subsequent builds)

> 1. For now there is no mechanism to detect unused stubs, right?

Nope, and I don't think there'll be any. It'd be hard to implement on our side, and I doubt Checker itself will ever support anything like that, given that stubs are kinda redundant by design (i.e. Checker has the entire JDK annotated, even though users will never actually use all of the JDK 😅)

> 2. What if we bump IntelliJ version and the stubbed method/class no longer exists? Are there any errors in such case (I do not remember any)? Or is it ignored?

See https://checkerframework.org/manual/#stub-troubleshooting-type-checking-results... we use `-AstubWarnIfNotFoundIgnoresClasses`, so CF _reports only missing methods/fields, but ignores missing classes, even if other classes from the same package are present_

> 3. (assuming 1. is true) Is it a problem the over time we will have more and more redundant stubs? E.g does it significantly affect the compilation/build time? If it is a problem then I might be worth to introduce some detection mechanism.

I doubt it really affects compilation times... given that Checker Framework itself ships with full annotated JDK!

mkondratek: thanks! clear!

CHANGE-NOTES.md
```diff
@@ -5,6 +5,7 @@
 - Added: extra backup keyboard shortcut (`Ctrl + Option + Shift + M`) for opening Git Machete tab on macOS,
   in case `Command + Option + Shift + M` is already taken by Kotlin plugin.
 - Improved: push action is no longer blocked by the other machete actions; it can be run in parallel.
+- Added: `Yes + Quit` button in Traverse dialogs.
```
mkondratek, line 8: while we are here... does it make sense to group added, improved, fixed... ? Which means to simply sort them. I think I saw that grouping in some change notes - it's easier to review them esp if you are looking for smth. probably makes it's more important when there is a lot of changes... 🤔

mkondratek (approved):
tested. works fine 🚀

CHANGE-NOTES.md
PawelLipski: Right... makes sense, lemme reorder

CONTRIBUTING.md
```diff
@@ -283,10 +287,10 @@ Then, update `executors.docker_executor.docker[0].image` in [.circleci/config.ym
 
 We follow [Semantic versioning](https://semver.org/) for the plugin releases:
 
-* MAJOR version must be bumped for each plugin release that stops supporting any IDEA build (typically when `sinceBuild` is increased). <br/>
```
PawelLipski, line 286: https://github.com/VirtusLab/git-machete-intellij-plugin/issues/1612 for further updates here

PR #1643, opened 2023-04-07
Test handling of failures in `machete-status-branch` hook

PawelLipski:


backend/impl/src/test/resources/setup-for-diverged-and-older-than.sh-discover.txt
```diff
@@ -1,4 +1,4 @@
-  master (behind origin)  <2 files>
```
mkondratek, line 1: was it a but? 🤔

testCommon/src/testFixtures/resources/common.sh
```diff
@@ -4,6 +4,11 @@ status_branch_hook=$(cat <<'EOF'
 #!/usr/bin/env bash
 branch=$1
 file_count=$(git ls-tree $branch | wc -l | sed 's/^ *//')
```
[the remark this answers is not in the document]
PawelLipski: `git ls-tree` just lists the top-level files for the given commit reference (like branch name).  I've just picked it for the hook to provide some low-effort output that would differ between branches.

backend/impl/src/test/resources/setup-for-diverged-and-older-than.sh-discover.txt
mkondratek: bug* 🐛  I see now it was not

PR #1699, opened 2023-08-11
Set up dependabot, downgrade some versions for test

PawelLipski:


.github/dependabot.yml
```diff
@@ -0,0 +1,19 @@
+# Please see the documentation for all configuration options:
```
PawelLipski, line 1: Let's change the default branch to `develop` for the time being (since this config file is apparently picked up from the repo's default branch)

mkondratek:
why are those downgrades needed?

PawelLipski:
> why are those downgrades needed?

Just to check whether they're correctly bumped by dependabot

PR #1708, opened 2023-08-11
Revert pre-#1699 package versions + tweak dependabot + phase out old dependency update implementation

PawelLipski:


mkondratek (commented):
Nice. And it looks like code removal

PR #1720, opened 2023-08-23
Move all action updates from EDT to BGT

PawelLipski:


frontend/actions/src/main/java/com/virtuslab/gitmachete/frontend/actions/base/BaseProjectDependentAction.java
```diff
@@ -41,7 +43,7 @@ private static void preloadClasses() {
 
   @Override
   public final ActionUpdateThread getActionUpdateThread() {
-    return ActionUpdateThread.EDT;
+    return ActionUpdateThread.BGT;
```
PawelLipski, line 46: This has **no** right to work (we're calling a looot of UI-touching methods from a non-UI thread this way)... yet it somehow does 😅 lemme test that locally for a while

PawelLipski: ...no errors so far :trollface:

PR #1822, opened 2024-02-24
Use zip distribution of IntelliJ in UI tests on macOS

PawelLipski:
Likely won't work OOTB on future IntelliJ versions, alas. Proper solution involves reading `product_info.json`, as in https://github.com/VirtusLab/ide-probe/issues/332. Let's close the related issues and track the problem there from now on.

### DIFF
--- a/src/uiTest/resources/ideprobe-macos.conf
+++ b/src/uiTest/resources/ideprobe-macos.conf
@@ -1,12 +1,13 @@
-# FIXME (#1383): revert to a single config file for macOS and Linux
+# FIXME (VirtusLab/ideprobe#332): revert to a single config file for macOS and Linux
 include "ideprobe-common.conf"
 
 probe {
-  intellij {
-    version {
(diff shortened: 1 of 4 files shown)


src/uiTest/resources/ideprobe-macos.conf
```diff
@@ -1,12 +1,11 @@
...
 
 probe {
-  intellij {
-    version {
-      # Note that DMGs are apparently published for EAPs and stable releases,
-      # but NOT for candidate EAPs (which our autoupdater in CI doesn't pick up anyway).
-      ext = ".dmg"
-    }
-  }
+  driver.vmOptions = ${probe.driver.vmOptions} [
```
PawelLipski, line 5: Values taken from `ps -f` for the IntelliJ executed by `./gradlew runIde`, then narrowed down to what's necessary.

PawelLipski: Works under 2023.3 and 2024.1

PR #1962, opened 2024-12-14
Bump dependencies (leave just Vavr unbumped to allow for testing Dependabot)

PawelLipski:


gradle/libs.versions.toml
```diff
@@ -1,5 +1,5 @@
 [versions]
-checker = "3.45.0"
+checker = "3.47.0"
```
PawelLipski, line 2: 3.48.3 is the most recent; but since 3.48.0 it's failing on 

```
> Task :backend:impl:compileJava
warning: The following options were not recognized by any processor: '[SubtypingChecker_qualDirs]'
error: warnings found and -Werror specified
```

PR #2198, opened 2025-12-08
Replace certain Rhino JS-powered operations in UI tests with Kotlin DSL

PawelLipski:


PawelLipski:
Not as easy :/ tried a smaller change in https://github.com/VirtusLab/git-machete-intellij-plugin/pull/2200, still failed. Let's leave as is

PR #2200, opened 2025-12-11
Test locating buttons with `xquery`

PawelLipski:


PawelLipski:
Huh, hard to do, let's stick to JS then

PR #2226, opened 2026-01-31
Handle cross-platform shell execution and adapt test utilities accordingly

mkondratek:


backend/impl/src/test/java/com/virtuslab/gitmachete/backend/integration/StatusAndDiscoverIntegrationTestSuite.java
mkondratek, line 1: Tests in this class were failing before on Windows. Now they pass. 🟢

testCommon/src/testFixtures/java/com/virtuslab/gitmachete/testcommon/TestProcessUtils.java
```diff
@@ -20,16 +20,16 @@ public static String runProcessAndReturnStdout(Path workingDirectory, int timeou
     Process process = new ProcessBuilder()
         .command(command)
         .directory(workingDirectory.toFile())
+        .redirectErrorStream(true)
```
mkondratek, line 23: On Windows, the OS pipe buffer for process output is limited. In the original implementation, the Java process called `process.waitFor()` before reading from the output streams. When the script's output exceeded the buffer size, the script would block waiting for the buffer to be cleared, while the Java process was blocked waiting for the script to finish, resulting in a deadlock.

mkondratek:
@PawelLipski, let me know whether we want to get all these tests running and green on Windows.

I think it should be possible (including the UI tests), but we’d also need Windows CI. Without it, we’ll likely end up with Windows-only failures again sooner or later.

If setting up a Windows pipeline is too costly, maybe we should reconsider fixing these tests on Windows for now. Even this PR—which is mostly a preamble for follow-up changes—already adds quite a bit of Windows-specific boilerplate. If that code is likely to stay unused, maybe we shouldn’t introduce it yet.

WDYT?

PawelLipski:
> @PawelLipski, let me know whether we want to get all these tests running and green on Windows.
> 
> I think it should be possible (including the UI tests), but we’d also need Windows CI. Without it, we’ll likely end up with Windows-only failures again sooner or later.
> 
> If setting up a Windows pipeline is too costly, maybe we should reconsider fixing these tests on Windows for now. Even this PR—which is mostly a preamble for follow-up changes—already adds quite a bit of Windows-specific boilerplate. If that code is likely to stay unused, maybe we shouldn’t introduce it yet.
> 
> WDYT?

Definitely, let's set up Windows CI if it isn't too difficult. In git-machete CLI, there's a dedicated [Windows executor/job](https://github.com/VirtusLab/git-machete/blob/d0990ddc2c61bdfceb751cbc2a69136c5834c408/.circleci/config.yml#L19) and even a macOS executor/job. Both are free for public repositories & generally smooth to use.

backend/impl/src/test/java/com/virtuslab/gitmachete/backend/integration/StatusAndDiscoverIntegrationTestSuite.java
```diff
@@ -1,12 +1,7 @@
 package com.virtuslab.gitmachete.backend.integration;
 
-import static com.virtuslab.gitmachete.backend.api.SyncToRemoteStatus.AheadOfRemote;
```
PawelLipski, line 3: Wow weird that these imports weren't removed, despite [`removeUnusedImports()`](https://github.com/VirtusLab/git-machete-intellij-plugin/blob/21487a17e3d320f8d16eb2d5bfa4662372e31d44/buildSrc/src/main/kotlin/com/virtuslab/gitmachete/buildsrc/Spotless.kt#L19) settings.

Lemme open an issue for fixing that - https://github.com/VirtusLab/git-machete-intellij-plugin/issues/2228

src/uiTest/kotlin/com/virtuslab/gitmachete/uitest/BaseUITestSuite.kt
```diff
@@ -256,6 +256,9 @@ abstract class BaseUITestSuite : TestGitRepository(SetupScripts.SETUP_WITH_SINGL
     rootDirectoryPath.resolve("machete-post-slide-out-hook-executed")
 
   fun Path.makeExecutable() {
+    if (System.getProperty("os.name").lowercase(getDefault()).contains("windows")) {
```
PawelLipski, line 259: Let's extract this to some shared place in testCommon

testCommon/src/testFixtures/java/com/virtuslab/gitmachete/testcommon/TestFileUtils.java
```diff
@@ -23,9 +22,33 @@ public static void copyScriptFromResources(String scriptName, Path targetDir) {
...
+  public static String runProcessAndReturnStdout(int timeoutSeconds, String... command) {
+    Path currentDir = Paths.get(".").toAbsolutePath().normalize();
+    return TestProcessUtils.runProcessAndReturnStdout(currentDir, timeoutSeconds, command);
+  }
+
+  public static String getShellExecutable() {
+    String shell = "bash";
+    String osName = System.getProperty("os.name").toLowerCase();
+    if (osName.contains("windows")) {
+      String gitPath = runProcessAndReturnStdout(5, "where", "git").trim().split(System.lineSeparator())[0];
```
PawelLipski, line 35: Can it just be `where sh` instead of `where git`? 🤔

testCommon/src/testFixtures/java/com/virtuslab/gitmachete/testcommon/TestProcessUtils.java
```diff
@@ -20,16 +20,16 @@ public static String runProcessAndReturnStdout(Path workingDirectory, int timeou
...
         .start();
-    boolean completed = process.waitFor(timeoutSeconds, TimeUnit.SECONDS);
 
     String stdout = IOUtils.toString(process.getInputStream(), StandardCharsets.UTF_8);
-    String stderr = IOUtils.toString(process.getErrorStream(), StandardCharsets.UTF_8);
+    boolean completed = process.waitFor(timeoutSeconds, TimeUnit.SECONDS);
```
PawelLipski, line 27: Hmmm now that `waitFor` is after `getInputStream()`... can `String stderr = IOUtils.toString(process.getErrorStream(), ...)` be restored? or is `redirectErrorStream` still required for some reason?

testCommon/src/testFixtures/resources/common.sh
```diff
@@ -43,12 +43,10 @@ function create_repo() {
   cd $dir || exit 1
   shift
   git init "$@"
-  if [[ "$(uname -s)" != *MINGW*_NT* ]]; then
```
PawelLipski, line 46: 🤔 not needed anymore?

PR #2261, opened 2026-04-09
Fix Squash action incorrectly remaining visible (fixes #2260)

### LINKED ISSUE #2260: Switching roots breaks the offered actions
## Bug description

I have a repository with a different repository inside (this is a multiple roots case). 
Only `main` in both, checked out.
For the outer repo I can see a different set of actions than for the inner. See the screenshots.

## (optional) Screenshots

Image

Image

mkondratek:
In this PR:
1. Fix squash action visibility — previously it was not being hidden in some cases
2. Fix branch layout issues on switching git modules — previously the branch layout was persisted in UI and that led to inconsistency and wrong actions being available
3. Auto-discover branch layout in case of empty — I cannot see much point in saying to the user "the branch layout is empty, try to discover". We can do that for them. I think it is a better UX.

### DIFF
--- a/CHANGE-NOTES.md
+++ b/CHANGE-NOTES.md
@@ -1,6 +1,8 @@
 # Changelog
 
 ## v7.0.1
+- Added: auto-discovery of the branch layout in case of multiple repositories
+- Fixed: branch layout and actions availability in case of multiple repositories
 
 ## v7.0.0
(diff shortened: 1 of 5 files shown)


PawelLipski (approved):
Wzbaniale

Pls fill up CHANGE NOTES btw

PR #2278, opened 2026-05-07
fix: "Show in Git Log" actions jumping to wrong repository in multi-root projects

mkondratek:


PawelLipski (commented):
Thx, lemme apply a few fix directly onto this PR...

PawelLipski (approved):
Simplified a bit. Tested locally, seems to work ;p

